PR #38, opened 2023-11-03
fix ubuntu  environments compile errors

lec-bit:
1、In Ubnutu the default shell (sh) for executing scripts is not bash, enforce it to be bash. 
2、In Ubuntu ，files for the x86_64 architecture such as asm are located in the /usr/include/x86_64-linux-gnu directory, add it to the default include directory for C.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -21,10 +21,11 @@ function set_enhanced_kernel_env() {
 }
 
 function prepare() {
-    sh kmesh_macros_env.sh
-    sh kmesh_bpf_env.sh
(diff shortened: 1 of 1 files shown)


lec-bit:
一些编译环境中默认的sh执行脚本不是bash，强制指定为bash
一些编译环境中x86_64架构的asm等文件在/usr/include/x86_64-linux-gnu目录下，把它加入到C的默认include目录中

lec-bit:
image

lec-bit:
image

PR #39, opened 2023-11-03
kmesh marcos optimization

lec-bit:
1、Merge duplicate semantic macro definitions.
2、Optimize the logic of the macro build script. Determine the feature capabilities supported by the current environment during compilation, and then set the values of the macros accordingly.


lec-bit:
1、合并重复语义宏定义

2、优化编译宏构建脚本的逻辑，在编译时判断当前环境所支持的功能特性，然后设置宏的值

PR #64, opened 2023-12-25
support compile online image

lec-bit:
1、Obtain the bpf_mem_ptr structure from the ENHANCED_KERNEL macro and determine whether the structure exists in the environment.
2、The logic for obtaining compilation macros is modified so that /usr/include/linux/bpf.h of the host machine can be read in the online compilation environment.
3、kmesh.yaml and kmesh.dockerfile adapted

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,7 +13,20 @@ function bpf_compile_range_adjust() {
 }
 
 function set_enhanced_kernel_env() {
-    if grep -q "FN(parse_header_msg)" /usr/include/linux/bpf.h; then
+    # we use /usr/include/linux/bpf.h to determine the runtime environment’s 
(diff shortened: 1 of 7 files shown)


build.sh
```diff
@@ -13,7 +13,12 @@ function bpf_compile_range_adjust() {
 }
 
 function set_enhanced_kernel_env() {
-    if grep -q "FN(parse_header_msg)" /usr/include/linux/bpf.h; then
+    export KERNEL_HEADER_LINUX_BPF=config/linux-bpf.h
```
nlgwcy, line 16: Add comment to describe the path semantics.

nlgwcy (approved):

Issue #66, opened 2023-12-28
Improve project readme


kevin-wangzefeng:
1. The project logo is too large, which wastes page space and results in poor reading experience.
2. Project-related tags and badges to be added, e.g. CI, SLSA level, releases, openssf best practices etc.

PR #67, opened 2023-12-28
kmesh readme optimization

nlgwcy:
kmesh readme optimization
1. Change the kmesh log in the readme to landscape orientation as the original icon is too large. fix https://github.com/kmesh-net/kmesh/issues/66
2. Change the kmesh roadmap indicator from an image to a check mark (√).


nlgwcy (commented):
ok

PR #73, opened 2024-01-03
remove kmesh container prepare

hzxuzhonghu:
As we have transited to ghcr.io, we do not need to set insecure registry any more.


hzxuzhonghu:
ping @nlgwcy

PR #99, opened 2024-01-23
Donot intercept traffic from sidecar

hzxuzhonghu:
**What type of PR is this?**

/kind enhancement

With this patch, kmesh now donot take charge of traffic from sidecar. Even in case, kmesh and sidecar both simultaneously enabled, we now prioritize sidecar. 

Fixes #

```release-note

```


hzxuzhonghu:
ping @nlgwcy

hzxuzhonghu:
We need to update the docs now

PR #100, opened 2024-01-24
Fix CI workflow and fix golint issues

### LINKED ISSUE #54: Add CI Checks
This is an umbrella issue for CI tasks.
There are many checks we should make in CI

- [ ] Boilerplate header check
- [ ] Lint check
- [ ] Build
- [ ] Unit test
- [ ] Integration test

VanderChen:
**What type of PR is this?**
/kind bug

1. Adapt CI workflows to the current version.
2. Fix some golint issues, including import order and unused function codes.

Fixes #54 

There are unused function codes in the existing code, which will cause the golint check to fail. Due to possible future use, comment them out rather than delete them for now.

```release-note

```


pkg/bpf/pin.go
```diff
@@ -59,24 +59,26 @@ func unpinPrograms(value *reflect.Value) error {
...
-		if err := tp.Pin(path + info.Name); err != nil {
-			return fmt.Errorf("pin map failed, %s", err)
-		}
-	}
-
-	return nil
-}
+// Due to the golint issuse, comment unused functions temporaryly
+
+// func pinMaps(value *reflect.Value, path string) error {
```
hzxuzhonghu, line 64: @nlgwcy Do we plan to use them

hzxuzhonghu (approved):
/lgtm

PR #106, opened 2024-01-30
Intree proto apis

hzxuzhonghu:
**What type of PR is this?**

/kind enhancement

This pr does two things:

1. intree all the generated api files 
2. added gen check and go mod check

Fixes #

```release-note

```


api/Makefile
```diff
@@ -27,10 +27,11 @@ GO_BUILD_DIR := kmesh.net/kmesh/api
...
 
-.PHONY: all install clean
+.PHONY: gen-proto install clean
 
-all:
+gen-proto:
 
+	$(QUIET) bash ../hack/install-proto.sh
 	$(call printlog, PROTO, api/$(C_OUTPUT_DIR))
 	$(QUIET) protoc-c --proto_path=$(PROTO_PATH) --c_out=.  $(PROTO_SRC)
```
nlgwcy, line 36: Do we still need to execute proto-c？

hzxuzhonghu: yes, this is to gen proto. When we update the proto file, we need to run make gen-proto

PR #115, opened 2024-02-04
modify kmesh_compile.sh to suit different arch

### LINKED ISSUE #76: Support build kmesh from container
When I first run build.sh to build kmesh, it installs a lot of dependencies on my VM. While istio has an image named `build-tools`, a developer can run `make build`  `make docker`, `make lint` `make gen`, and many other cmd within the container. It can provide good contributing experience for new contributors.

So let's provide a build tools image, and support build kmesh binary/image from the container.

weli-l:
**What type of PR is this?**

/kind enhancement

modify kmesh_compile.sh to suit different arch

Fixes #76 

NONE
```release-note

```

### DIFF
--- a/kmesh_compile.sh
+++ b/kmesh_compile.sh
@@ -1,11 +1,16 @@
 #!/bin/bash
 
 function prepare() {
-    docker pull ghcr.io/kmesh-net/kmesh-build:v0.2.0
+    local arch
+    arch=$(get_arch)
+    docker pull "ghcr.io/kmesh-net/kmesh-build-${arch}:v0.2.0"
 }
 
 function run_docker_container() {
-    local container_id=$(docker run -itd --privileged=true \
+    local arch
+    arch=$(get_arch)
+    local container_id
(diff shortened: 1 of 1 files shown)


kmesh_compile.sh
```diff
@@ -1,22 +1,35 @@
 #!/bin/bash
 
 function prepare() {
-    docker pull ghcr.io/kmesh-net/kmesh-build:v0.2.0
+    local arch
+    arch=$(get_arch)
+    docker pull "ghcr.io/kmesh-net/kmesh-build-${arch}:v0.2.0"
```
nlgwcy, line 6: The compilation image should be relatively stable, and the image tag should be latest.

weli-l: > The compilation image should be relatively stable, and the image tag should be latest.

The compilation image will change with the changes in the software packages that kmesh depends on, and different versions of the compilation image can be used to compile different versions of kmesh.

PR #118, opened 2024-02-05
modify kmesh.dockerfile to adapt to throw kmesh.json

### LINKED ISSUE #76: Support build kmesh from container

weli-l:
**What type of PR is this?**

/kind enhancement

modify kmesh.dockerfile to adapt to throw kmesh.json

Fixes #76 

NONE


build/docker/kmesh_compatible.yaml
```diff
@@ -0,0 +1,101 @@
...
+           securityContext:
+             privileged: true
+             capabilities:
+               add: ["all"]
+           ports:
+           - containerPort: 6789
+             hostPort: 6789
+           env:
+           - name: MESH_CONTROLLER
+             value: istio-system:istiod
```
hzxuzhonghu, line 63: not right now

hzxuzhonghu (commented):
I think we should add comments  how does this compatible one diff with previous one

PR #129, opened 2024-02-18
cleanup tmp modified files

hzxuzhonghu:
**What type of PR is this?**

/kind enhancement

after build, we can now make clean to restore all the modified files

Fixes #

```release-note

```


Makefile
```diff
@@ -44,6 +44,14 @@ APPS2 := kmesh-cmd
 APPS3 := mdacore
 APPS4 := kmesh-cni
 
+TMP_FILES := bpf/kmesh/bpf2go/bpf2go.go \
+	bpf/kmesh/bpf2go/kmeshcgroupsockworkload_bpfeb.go \
```
nlgwcy, line 48: The two .go files are intermediate compilation files and should not be archived. Suggest adding to .gitignore

hzxuzhonghu: ? can not ignore, they are committed

hzxuzhonghu: Do you mean we should remove them?

kwb0523: yes, previously mistakenly added.

Makefile
```diff
@@ -44,6 +44,14 @@ APPS2 := kmesh-cmd
...
+	bpf/kmesh/bpf2go/kmeshcgroupsockworkload_bpfel.go \
```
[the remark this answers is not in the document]
hzxuzhonghu: IC similar as other files. But i am thinking maybe we need add all of them back once we add unit test, otherwise `go test ` would fail.

But first let we remove them.

hzxuzhonghu:
@nlgwcy @kwb0523 ptal

Issue #104, opened 2024-01-29
Improve unit test coverage


hzxuzhonghu:
**What would you like to be added**:

Currently there are no unit tests, we should add test coverage for each module. Basically i suggest doing this per package.

Eveyone can take part of this, and can list your interested package below

hzxuzhonghu:
some critical components that should be tested including:

- [ ] cache under pkg/cache/v2
- [ ] cni plugin under pkg/cni
- [ ] ads client  under pkg/controller
- [ ] wds client under pkg/controller_workload
- [ ] Ebpf map and enpf redirect @nlgwcy to confirm if this is able to be tested

PR #139, opened 2024-02-29
Demonstrate UT coverage in CI

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Demonstrate UT coverage in CI. Allow developers to create pull request with a clear understanding of the UT coverage of their code.
a part of #104

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,5 +39,8 @@ jobs:
         skip-pkg-cache: true
 
     - name: Go Test
-      run: |
-        go test -v -vet=off ./...
+      run: go test -race -coverprofile=coverage.txt -covermode=atomic $(go list ./pkg/...)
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -39,5 +39,8 @@ jobs:
         skip-pkg-cache: true
 
     - name: Go Test
-      run: |
-        go test -v -vet=off ./...
+      run: go test -race -coverprofile=coverage.txt -covermode=atomic $(go list ./pkg/...)
```
hzxuzhonghu, line 42: We can only test ./pkg right now, but we may need to test other pkg as well

LiZhenCheng9527: We can test other directories at that time.

hzxuzhonghu (approved):
/lgtm

PR #143, opened 2024-03-04
A ut of cni plugin

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

a part of #104

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,8 +39,4 @@ jobs:
         skip-pkg-cache: true
 
     - name: Go Test
-      run: go test -race -coverprofile=coverage.txt -covermode=atomic $(go list ./pkg/...)
-    - name: Upload coverage to Codecov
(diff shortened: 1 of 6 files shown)


LiZhenCheng9527:
**Note**: This PR is just an example of Kmesh UT for now.
I will improve the UT of the functions and methods in this Go file later.
/hold

hzxuzhonghu:
>time="[20](https://github.com/kmesh-net/kmesh/actions/runs/8135459431/job/22230036362?pr=143#step:8:21)24-03-04T03:43:12Z" level=fatal msg="failed to create log directory"


```
	err := os.MkdirAll(path, 0o700)
	if err != nil {
		logger.Fatal("failed to create log directory")
	}
```

It seems the permission is not allowed. We should update the logger

hzxuzhonghu (approved):
/lgtm

Approve first, but we shouldn't require run ut with sudo priv

PR #147, opened 2024-03-06
use big endian to restore IP

### LINKED ISSUE #146: ip bytes converts to uint32 incorrect
**What happened**:
As IP address is stored with big endian, but here we use little endian to restore.  https://github.com/kmesh-net/kmesh/blob/main/pkg/nets/nets.go#L56

**Environment**:
- Kmesh version:
- Others:

LiZhenCheng9527:
**What type of PR is this?**
/kind bug

Fixes #146

### DIFF
--- a/pkg/nets/nets.go
+++ b/pkg/nets/nets.go
@@ -54,5 +54,5 @@ func ConvertPortToBigEndian(little uint32) uint32 {
 
 // ConvertIpByteToUint32 converts ip to little-endian uint32 format
 func ConvertIpByteToUint32(ip []byte) uint32 {
-	return binary.LittleEndian.Uint32(ip)
+	return binary.BigEndian.Uint32(ip)
 }


hzxuzhonghu:
cc @kwb0523

pkg/nets/nets.go
```diff
@@ -54,5 +54,5 @@ func ConvertPortToBigEndian(little uint32) uint32 {
 
 // ConvertIpByteToUint32 converts ip to little-endian uint32 format
 func ConvertIpByteToUint32(ip []byte) uint32 {
-	return binary.LittleEndian.Uint32(ip)
+	return binary.BigEndian.Uint32(ip)
```
hzxuzhonghu, line 57: please update the function comments as well

LiZhenCheng9527: https://github.com/kmesh-net/kmesh/blob/67e834659309957b31e63c60cfc507d61c4a4a9b/pkg/nets/nets.go#L29
Should this function also be modified?

hzxuzhonghu: It depends on the caller, i think we get the IP slices from istiod.

pkg/nets/nets.go
```diff
@@ -54,5 +54,5 @@ func ConvertPortToBigEndian(little uint32) uint32 {
 
 // ConvertIpByteToUint32 converts ip to little-endian uint32 format
```
hzxuzhonghu, line 55:
```suggestion
// ConvertIpByteToUint32 converts big-endian ip slices to uint32 format
```

kwb0523:
there are any specific bug descriptions?

kwb0523:
> cc @kwb0523

fixed in https://github.com/kmesh-net/kmesh/pull/138

hzxuzhonghu:
@kwb0523 Not sure, let's make a pr as smaller as possible. Not to couple with that bigger refactor one

hzxuzhonghu (approved):
/lgtm

PR #152, opened 2024-03-07
Prevent writing empty cni conflist

### LINKED ISSUE #150: Kmesh CNI overwrite cni conflist with empty value
**What happened**:

After running kmesh for some time, we cannot restart kmesh daemon and also not able to start a new application pod.

By looking into kmesh logs, we do see:

```
time="2024-03-06T10:06:41Z" level=info msg="Copied /usr/bin/kmesh-cni to /opt/cni/bin." subsys="cni installer"
time="2024-03-06T10:06:41Z" level=info msg="failed to read conflist: /etc/cni/net.d/10-calico.conflist, error parsing configuration list: unexpected end of JSON input" subsys="cni installer"
time="2024-03-06T10:06:41Z" level=error msg="can not found the valid cni config!\n" subsys="cni installer"
time="2024-03-06T10:06:41Z" level=error msg="can not found the valid cni config!\n" subsys=manager
```

Kmesh should not overwrite invalid conflist

not sure how to reproduce

**Environment**:
- Kmesh version:
- Others:

hzxuzhonghu:
**What type of PR is this?**

/kind bug

Fixes #150

```release-note
Fixed kmesh cni installer write empty cni conflist when kmesh daemon is restarted.
```

### DIFF
--- a/pkg/cni/chained.go
+++ b/pkg/cni/chained.go
@@ -64,22 +64,23 @@ func getCniConfigPath() (string, error) {
 		sort.Strings(files)
 		var confList *libcni.NetworkConfigList = nil
 
-		for _, confFile = range files {
-			confList, err = libcni.ConfListFromFile(confFile)
+		for _, file := range files {
+			confList, err = libcni.ConfListFromFile(file)
 			if err != nil {
-				err = fmt.Errorf("failed to read conflist: %v, %v", confFile, err)
+				err = fmt.Errorf("failed to read conflist: %v, %v", file, err)
 				log.Info(err)
 				continue
 			}
 			if len(confList.Plugins) == 0 {
-				log.Infof("file %s plugins is empty\n", confList.Name)
+				log.Infof("file %s plugins is empty", confList.Name)
 				continue
 			} else {
+				confFile = file
 				break
 			}
 		}
 		if confList == nil || len(confList.Plugins) == 0 {
-			err = fmt.Errorf("can not found the valid cni config!\n")
+			err = fmt.Errorf("can not found the valid cni config")
 			log.Error(err)
 			return "", err
 		}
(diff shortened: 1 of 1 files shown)


pkg/cni/chained.go
```diff
@@ -64,22 +64,23 @@ func getCniConfigPath() (string, error) {
...
 		var confList *libcni.NetworkConfigList = nil
 
-		for _, confFile = range files {
-			confList, err = libcni.ConfListFromFile(confFile)
+		for _, file := range files {
+			confList, err = libcni.ConfListFromFile(file)
 			if err != nil {
-				err = fmt.Errorf("failed to read conflist: %v, %v", confFile, err)
+				err = fmt.Errorf("failed to read conflist: %v, %v", file, err)
 				log.Info(err)
```
LiZhenCheng9527, line 71:
```suggestion
				log.Info("failed to read conflist: %v, %v", file, err)
```
Why not use one line of code to output the error log?

pkg/cni/chained.go
```diff
@@ -112,7 +113,7 @@ func insertCNIConfig(oldconfig []byte) ([]byte, error) {
 			return nil, err
 		}
 		if plugin["type"] == kmeshCniPluginName {
-			return nil, nil
+			log.Infof("%s was installed but not cleaned up, but we would overwrite it", kmeshCniPluginName)
```
nlgwcy, line 116: Will append repeated？

pkg/cni/chained.go
hzxuzhonghu: Tep, this is not a good way to process err.

Let's handle errors in a separate pr

pkg/cni/chained.go
hzxuzhonghu: You are right. THis is not intended.

PR #155, opened 2024-03-07
fix HashName mem leak

### LINKED ISSUE #153: HashName mem leak
**What happened**:

In workload model, we introduced many bpf maps keyyed by str's hash value. And we used an inner object  `HashName` to provide unique hash val, it stores a map innerly. But we can see the keys is never deleted even when a workload/service removed.

FYI: https://github.com/kmesh-net/kmesh/blob/main/pkg/controller_workload/workload/workload_hash.go

**Environment**:
- Kmesh version:
- Others:

LiZhenCheng9527:
**What type of PR is this?**
/kind bug

Fixed the bug that when deleting workload and service, the related data in hashName was not deleted.
Fixes #153

### DIFF
--- a/pkg/controller_workload/workload/workload_event.go
+++ b/pkg/controller_workload/workload/workload_event.go
@@ -162,6 +162,7 @@ func RemoveWorkloadResource(removed_resources []string) error {
 			log.Errorf("BackendDelete failed: %s", err)
 			goto failed
 		}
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (approved):
/lgtm

Thinking a bit more, the HashName is a little obscures, i donot have an idea to help developer not forget to delete the hash map now.

PR #158, opened 2024-03-07
Modify error handling in ads controller

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement
Modify the unnecessary return error handling in the ads controller.


hzxuzhonghu:
Not sure I can get the motivation of this patch.

Issue #154, opened 2024-03-07
Removed resources in wds may contains both Workload and Service types


hzxuzhonghu:
**What happened**:

Now we only consider the removed resources are only pod or only service. This is not right, more details we can see the controlplane istiod. 

https://github.com/kmesh-net/kmesh/blob/e5db146f9083daf4483debca6a732543757af323/pkg/controller_workload/workload/workload_event.go#L466-L483

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kmesh version:
- Others:

hzxuzhonghu:
Also we can have `WorkloadEntry` and ServiceEntry

hzxuzhonghu:
Fixed

PR #160, opened 2024-03-08
Fix workload remove

hzxuzhonghu:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_event.go
+++ b/pkg/controller/workload/workload_event.go
@@ -108,19 +108,6 @@ func (svc *ServiceEvent) processWorkloadResponse(rsp *service_discovery_v3.Delta
 			err = fmt.Errorf("unsupport type url %s", rsp.GetTypeUrl())
 		}
 	}
-
-	if rsp.RemovedResources != nil {
-		log.Debugf("RemovedResources %s\n", rsp.RemovedResources)
-		switch rsp.GetTypeUrl() {
-		case AddressType:
-			err = handleDeleteResponse(rsp)
-		case AuthorizationType:
-			err = svc.handleAuthorizationTypeResponse(rsp)
-		default:
-			err = fmt.Errorf("unsupport type url %s", rsp.GetTypeUrl())
-		}
-	}
-
 	if err != nil {
 		log.Error(err)
 	}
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
fix https://github.com/kmesh-net/kmesh/issues/154

pkg/controller/workload/workload_event.go
```diff
@@ -108,19 +108,6 @@ func (svc *ServiceEvent) processWorkloadResponse(rsp *service_discovery_v3.Delta
 			err = fmt.Errorf("unsupport type url %s", rsp.GetTypeUrl())
 		}
 	}
-
```
supercharge-xsy, line 111: is there a situation we need to consider？  rsp.GetResources() == nil & rsp.RemovedResources != nil

pkg/controller/workload/workload_event.go
```diff
@@ -108,19 +108,6 @@ func (svc *ServiceEvent) processWorkloadResponse(rsp *service_discovery_v3.Delta
...
-	if rsp.RemovedResources != nil {
```
supercharge-xsy, line 112: is there a situation we need to consider？ rsp.GetResources() == nil & rsp.RemovedResources != nil

hzxuzhonghu: I have checked, in the function we use for range, if this is nil, it does not matter. 

Just we need to guarantee all the functions removeXXX or handleXX is safe with nil input

supercharge-xsy: we need also change this condition to enter handlexx function， when rsp.GetResources() == nil  but rsp.RemovedResources != nil, should we ? line 101 :
`	if rsp.GetResources() != nil {
		switch rsp.GetTypeUrl() {
		case AddressType:
			err = handleAddressTypeResponse(rsp)
		case AuthorizationType:
			err = svc.handleAuthorizationTypeResponse(rsp)
		default:
			err = fmt.Errorf("unsupport type url %s", rsp.GetTypeUrl())
		}
	}`

hzxuzhonghu: Not sure i can get what you mean

supercharge-xsy: > Not sure i can get what you mean

in processWorkloadResponse function ，The current code implementation misses a situation ：when rsp has RemovedResources field without Resources field

hzxuzhonghu:
@nlgwcy can you approve

hzxuzhonghu:
@nlgwcy can you approve

PR #163, opened 2024-03-09
add ut of ads_stream.go

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

add ut of ads_stream.go

### DIFF
--- a/pkg/controller/envoy/ads_stream_test.go
+++ b/pkg/controller/envoy/ads_stream_test.go
@@ -1,0 +1,234 @@
+/*
+ * Copyright 2024 The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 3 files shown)


pkg/controller/envoy/ads_stream_test.go
```diff
@@ -0,0 +1,64 @@
...
+					func(ctx context.Context, opts ...grpc.CallOption) (service_discovery_v3.AggregatedDiscoveryService_StreamAggregatedResourcesClient, error) {
+						return nil, errors.New("fail to create adsstream")
+					})
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.beforeFunc()
+			err := adsStream.AdsStreamCreateAndSend(client, context.Background())
```
hzxuzhonghu, line 56: This is not right. There is no server at all. Make no sense to me.

LiZhenCheng9527:
cancel /hold

LiZhenCheng9527:
cc @nlgwcy

PR #164, opened 2024-03-09
add ./out/amd64 in LD_LIBRARY_PATH

LiZhenCheng9527:
add ./out/amd64 in LD_LIBRARY_PATH


.github/workflows/main.yml
```diff
@@ -29,7 +29,8 @@ jobs:
 
     - name: Setup Enviroments
       run: |
-        echo "PKG_CONFIG_PATH=$GITHUB_WORKSPACE/mk" >> $GITHUB_ENV
+        echo "PKG_CONFIG_PATH=$GITHUB_WORKSPACE/mk" >> $GITHUB_ENV |
+        echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$GITHUB_WORKSPACE/out/amd64" >> $GITHUB_ENV
```
hzxuzhonghu, line 33: Is this for go test?

LiZhenCheng9527: `/tmp/go-build3582583922/b947/envoy.test: error while loading shared libraries: libkmesh_api_v2_c.so: cannot open shared object file: No such file or directory`
Because of the above error in #163 , add LD_LIBRARY_PATH to solve this error.

hzxuzhonghu: make sense

hzxuzhonghu (approved):
/lgtm

PR #177, opened 2024-03-19
set go test env

LiZhenCheng9527:
set go test env in order to run go test smoothly.


LiZhenCheng9527:
cc @nlgwcy

PR #179, opened 2024-03-20
Kmesh: Introduce a new method to query whether a pod is managed by Kmesh

bitcoffeeiux:
**What type of PR is this?**
Currently, the classid of the cgroup is used to determine whether the pod is managed by Kmesh. However, the classid is inconvenient to use. You need to change the Kubernetes hosting mode from systemd to cgroup. And not all programs can get classid in eBPF. Therefore, a new method is introduced to determine the IP address. When the CNI initializes the pod, the IP address of the pod is written into an eBPF map. In the eBPF, the map is directly read and compared with the obtained IP address to determine whether the IP address is managed by the Kmesh.

Introduce a new method to query whether a pod is managed by Kmesh

Fixes # NA

NONE
```release-note

```

### DIFF
--- a/bpf/kmesh/workload/sockops_tuple.c
+++ b/bpf/kmesh/workload/sockops_tuple.c
@@ -14,101 +14,155 @@
  * limitations under the License.
  */
 #include <linux/bpf.h>
+#include <sys/socket.h>
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_endian.h>
 #include "bpf_log.h"
 #include "workload.h"
+#include "config.h"
 
-#define AF_INET  (2)
-#define AF_INET6 (10)
+#define FORMAT_IP_LENGTH		(16) 
 
 enum family_type {
-    IPV4,
-    IPV6,
+	IPV4,
+	IPV6,
 };
 
 struct ringbuf_msg_type {
-    __u32 type; 
-    struct bpf_sock_tuple tuple;
+	__u32 type;
+	struct bpf_sock_tuple tuple;
 };
 
-// return 1 if pod with skops.daddr is managed by Kmesh, else return 0
-static inline int dst_is_managed(struct bpf_sock_ops *skops)
+/*
+ * containerID: hash for container id
+ * ipaddress: pod ip managed by Kmesh
+ * Either containerID or ipaddress must be set to 0.
+ * index: if ipaddress is set, index must be 0
+ *        if containerID is set, index == 0, map value is number of ips in containerID
+ *        if containerID is set, index > 0, map value is an ip for the containerID
+ * eg:
+ * there have a containerID and its hash is 64334212, it have 3 ip address in pod
+ * there have 7 record in thie map.
+ *      |containerID        |ip     |index      |||value        |
+ * 1.   |64334214           |0      |0          |||3            |
+ * 2.   |64334214           |0      |1          |||ip1          |
+ * 3.   |64334214           |0      |2          |||ip2          |
+ * 4.   |64334214           |0      |3          |||ip3          |
+ * 5.   |0                  |ip1    |0          |||0            |
+ * 6.   |0                  |ip2    |0          |||0            |
+ * 7.   |0                  |ip3    |0          |||0            |
+ *
+ * Why design it that way?
+ * We need a way to mark in the cni whether the current ip is managed by Kmesh.
+ * The cni inserts the ip address into the map when the pod is created and removes the ip
+ * address from the map when the pod is destroyed.
+ * However, according to the cni guide, when deleting the data, only the CONTAINER and IFNAME
+ * (https://github.com/containernetworking/cni.dev/blob/main/content/docs/spec.md#del-remove-container-from-network-or-un-apply-modifications)
+ * must be transferred. The IP address is not transferred in the cni. Therefore, the
+ * containerID and IP address must be bound and stored in the map for subsequent deletion.
+ */
+struct kmesh_manager_key {
+	__u64 containerID;
+	__u32 ipaddress;
+	__u32 index;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, struct kmesh_manager_key);
+	__type(value, __u32);
+	__uint(max_entries, MAP_SIZE_OF_MANAGER);
+	__uint(map_flags, 0);
+} map_of_kmesh_manager SEC(".maps");
+
+static inline int is_managed_by_kmesh(__u32 ip)
 {
-    return 0;
+	struct kmesh_manager_key key = {0};
+	key.ipaddress = ip;
+	__u8 *value = bpf_map_lookup_elem(&map_of_kmesh_manager, &key);
+	if (value)
+		return 1;
+	return 0;
 }
 
-static inline void build_auth_key(struct bpf_sock_ops *skops, struct bpf_sock_tuple* key)
+static inline void extract_skops_to_tuple_reverse(struct bpf_sock_ops *skops,
+	struct bpf_sock_tuple *tuple_key)
 {
-    if(key) {
-        key->ipv4.saddr = skops->remote_ip4;
-        key->ipv4.sport = skops->remote_port >> 16;
-        key->ipv4.daddr = skops->local_ip4;
-        key->ipv4.dport = bpf_htonl(skops->local_port) >> 16;
-    }
+	tuple_key->ipv4.saddr = skops->remote_ip4;
+	tuple_key->ipv4.daddr = skops->local_ip4;
+	// remote_port is network byteorder
+	// openEuler 2303 convert remote port different than other linux vendor
+#if !OE_23_03
+	tuple_key->ipv4.sport = skops->remote_port >> FORMAT_IP_LENGTH;
+#else	
+	tuple_key->ipv4.sport = skops->remote_port;
+#endif
+	// local_port is host byteorder
+	tuple_key->ipv4.dport = bpf_htonl(skops->local_port) >> FORMAT_IP_LENGTH;
 }
 
 // clean map_of_auth
 static inline void clean_auth_map(struct bpf_sock_ops *skops)
 {
-    struct bpf_sock_tuple tuple_key = {0};
-    build_auth_key(skops, &tuple_key);
-    long ret = bpf_map_delete_elem(&map_of_auth, &tuple_key);
-    if(ret && ret != -ENOENT) {
-        BPF_LOG(INFO, SOCKOPS, "map_of_auth bpf_map_delete_elem failed, ret: %d\n", ret);
-    }
+	struct bpf_sock_tuple tuple_key = {0};
+	// auth run PASSIVE ESTABLISHED CB now. In thie state cb
+	// tuple info src is server info, dst is client info
+	// During the auth, src must set the client info and dst set
+	// the server info when we transmitted to the kmesh auth info.
+	// In this way, auth can be performed normally.
+	extract_skops_to_tuple_reverse(skops, &tuple_key);
+	long ret = bpf_map_delete_elem(&map_of_auth, &tuple_key);
+	if(ret && ret != -ENOENT)
+		BPF_LOG(INFO, SOCKOPS, "map_of_auth bpf_map_delete_elem failed, ret: %d\n", ret);
 }
 
 // insert an IPv4 tuple into the ringbuf
-static inline void insert_ipv4_tuple(struct bpf_sock_ops *skops)
+static inline void auth_ip_tuple(struct bpf_sock_ops *skops)
 {
-    if (skops->family == AF_INET) {
-        struct ringbuf_msg_type *msg = bpf_ringbuf_reserve(&map_of_tuple, sizeof(*msg), 0);
-        if (!msg) {
-            BPF_LOG(WARN, SOCKOPS, "can not alloc new ringbuf in map_of_tuple");
-            return;
-        }
-        (*msg).tuple.ipv4.daddr = skops->local_ip4;
-        (*msg).tuple.ipv4.saddr = skops->remote_ip4;
-        // local_port is host byteorder
-        (*msg).tuple.ipv4.dport = bpf_htonl(skops->local_port) >> 16;
-        // remote_port is network byteorder
-        (*msg).tuple.ipv4.sport = skops->remote_port >> 16;
-        (*msg).type = (__u32)IPV4;
-        bpf_ringbuf_submit(msg, 0);
-    }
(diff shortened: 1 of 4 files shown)


pkg/cni/plugin/plugin.go
```diff
@@ -155,6 +162,106 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
...
+		value = 0
+		keyContainer.Index = totalNum + 1
+
+		if err = targetmap.Update(&keyContainer, &keyIP.IpAddress, ebpf.UpdateAny); err != nil {
+			log.Errorf("failed to record container :%v: %v", keyContainer.ContainerID, err)
+			// Try to insert the rest of the ip
+			continue
+		}
+
+		if err = targetmap.Update(&keyIP< &value, ebpf.UpdateAny); err != nil {
```
LiZhenCheng9527, line 206:
```suggestion
		if err = targetmap.Update(&keyIP, &value, ebpf.UpdateAny); err != nil {
```

LiZhenCheng9527:
FYI: Perhaps the functions `recordPod` and `cleanPodIp` could be rewritten as methods of *ebpf.Map

pkg/cni/plugin/plugin.go
```diff
@@ -155,6 +165,109 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
...
+	var keyIP kmeshManagerKey
+	var keyContainer kmeshManagerKey
+	var value uint32
+	var totalNum uint32
+	var err error
+
+	keyContainer.ContainerID = hash.Sum64String(containerID)
+
+	for _, allocIP := range preResult.IPs {
+		keyIP.IpAddress = binary.LittleEndian.Uint32(allocIP.Address.IP.To4())
```
hzxuzhonghu, line 200: can we reuse that function in nets

pkg/cni/plugin/plugin.go
```diff
@@ -39,11 +41,13 @@ import (
...
 	"kmesh.net/kmesh/pkg/utils"
+	"kmesh.net/kmesh/pkg/utils/hash"
 )
 
 var (
-	log               = logger.NewLoggerFieldWithoutStdout("plugin/cniplugin")
-	ENABLE_KMESH_MARK = "0x1000"
+	log                        = logger.NewLoggerFieldWithoutStdout("plugin/cniplugin")
+	ENABLE_KMESH_MARK          = "0x1000"
+	ENABLED_KMESH_MAP_PIN_PATH = "/sys/fs/bpf/bpf_kmesh_workload/map/map_of_kmesh_manager"
```
hzxuzhonghu, line 50: can this be more explicit?

pkg/cni/plugin/plugin.go
```diff
@@ -155,6 +165,109 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
...
+ * there have a containerID and its hash is 64334212, it have 3 ip address in pod
+ * there have 7 record in thie map.
+ *      |containerID        |ip     |index      |||value        |
+ * 1.   |64334214           |0      |0          |||3            |
+ * 2.   |64334214           |0      |1          |||ip1          |
+ * 3.   |64334214           |0      |2          |||ip2          |
+ * 4.   |64334214           |0      |3          |||ip3          |
+ * 5.   |0                  |ip1    |0          |||0            |
+ * 6.   |0                  |ip2    |0          |||0            |
+ * 7.   |0                  |ip3    |0          |||0            |
```
hzxuzhonghu, line 178: I can not fully get this point.

hzxuzhonghu (commented):
What do you want to do with the pod IP recorded

hzxuzhonghu:
I think we need a proposal before hand to help understand, before writing down this implementation

pkg/cni/plugin/plugin.go
bitcoffeeiux: This path is the pinned address of an ebpf map. This map is used to record the IP addresses that are managed by kmesh. After obtaining the FD of the map from the pinned path, we can directly perform operations on the map.

pkg/cni/plugin/plugin.go
```diff
@@ -155,6 +165,109 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
 	return nil
 }
 
+/*
+ * there have a containerID and its hash is 64334212, it have 3 ip address in pod
+ * there have 7 record in thie map.
```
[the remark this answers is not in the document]
bitcoffeeiux: similar to having multiple network devices, or multiple addresses are possible

pkg/cni/plugin/plugin.go
bitcoffeeiux: The containerID is mainly used for deletion. When the pod is destroyed, the cni is responsible for clearing the map that records the IP address (used to determine whether it is managed by kmesh). However, I cannot obtain the IP addresses of all network addresses in the cni, and can only obtain the containerID information. So I need the mapping between containerID and ip, which is the content of rows 1-4 in this table

pkg/cni/plugin/plugin.go
bitcoffeeiux: Yes, but what is the use case?

pkg/cni/plugin/plugin.go
```diff
@@ -155,6 +165,109 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
...
+	var keyContainer kmeshManagerKey
+	var value uint32
+	var totalNum uint32
+	var err error
+
+	keyContainer.ContainerID = hash.Sum64String(containerID)
+
+	for _, allocIP := range preResult.IPs {
+		keyIP.IpAddress = binary.LittleEndian.Uint32(allocIP.Address.IP.To4())
+		value = 0
```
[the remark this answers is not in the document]
bitcoffeeiux: yes, i will fix it

pkg/cni/plugin/plugin.go
```diff
@@ -155,6 +165,109 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
...
+		targetmap.Delete(&keyIP)
+		targetmap.Delete(&keyContainer)
+	}
+	keyContainer.Index = 0
+	targetmap.Delete(&keyContainer)
+}
+
+func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod, preResult *cniv1.Result, containerID string) error {
+	if err := markPodCgroupClassID(client, pod); err != nil {
+		return err
```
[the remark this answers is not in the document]
bitcoffeeiux: That's it for now. We'll need to think about it later.

bitcoffeeiux:
> FYI: Perhaps the functions `recordPod` and `cleanPodIp` could be rewritten as methods of *ebpf.Map

ebpf.Map is a kind of cilium api. It's a little difficult.

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,104 +14,178 @@
...
 
 // clean map_of_auth
-static inline void clean_auth_map(struct bpf_sock_ops *skops)
+static inline void clean_auth_map(struct bpf_sock_tuple *tuple_info)
+{
+	struct bpf_sock_tuple tuple_key = {0};
+	tuple_key.ipv4.saddr = tuple_info->ipv4.daddr;
+	tuple_key.ipv4.daddr = tuple_info->ipv4.saddr;
+	tuple_key.ipv4.sport = tuple_info->ipv4.dport;
+	tuple_key.ipv4.dport = tuple_info->ipv4.sport;
```
hzxuzhonghu, line 118: not sure i understand this conversion

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,104 +14,178 @@
...
+	struct ringbuf_msg_type *msg = bpf_ringbuf_reserve(&map_of_tuple, sizeof(*msg), 0);
+	if (!msg) {
+		BPF_LOG(WARN, SOCKOPS, "can not alloc new ringbuf in map_of_tuple");
+		return 1;
+	}
+	// auth run PASSIVE ESTABLISHED CB, src is server, dst is client
+	(*msg).tuple.ipv4.daddr = tuple_info->ipv4.saddr;
+	(*msg).tuple.ipv4.saddr = tuple_info->ipv4.daddr;
+	(*msg).tuple.ipv4.dport = tuple_info->ipv4.sport;
+	(*msg).tuple.ipv4.sport = tuple_info->ipv4.daddr;
```
hzxuzhonghu, line 137: same here

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,104 +14,178 @@
...
+	// reciver check src(server ip) is managed by the kmesh.
+	// If yes, auth need to be check.
+	// NOTICE! in PASSIVE ESTABLISHED CB state, server ip is src ip!
+	if (!kmesh_manager_by_ip(tuple_info.ipv4.saddr))
+		return 0;
+	switch (skops->op) {
+		case BPF_SOCK_OPS_ACTIVE_ESTABLISHED_CB:
+			if(bpf_sock_ops_cb_flags_set(skops, BPF_SOCK_OPS_STATE_CB_FLAG) != 0)
+				BPF_LOG(ERR, SOCKOPS, "set sockops cb failed!\n");
+			enable_encoding_metadata(skops, &tuple_info);
```
hzxuzhonghu, line 172: This seems not right, only when we have waypoint, we need to encode. Here we cannot know if we have any waypoint, right?

pkg/cni/plugin/plugin.go
```diff
@@ -126,7 +135,7 @@ func checkKmesh(client kubernetes.Interface, pod *v1.Pod) (bool, error) {
 	return false, nil
 }
 
-func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
+func kmeshCtlByClassid(client kubernetes.Interface, pod *v1.Pod) error {
```
hzxuzhonghu, line 138: do we still classID now?

pkg/cni/plugin/plugin.go
```diff
@@ -160,6 +169,108 @@ func enableKmeshControl(client kubernetes.Interface, pod *v1.Pod) error {
...
+	var keyContainer kmeshManagerKey
+	var value uint32 = 0
+	var totalNum uint32
+	var err error
+
+	keyContainer.ContainerID = hash.Sum64String(containerID)
+
+	for _, allocIP := range preResult.IPs {
+		keyIP.IpAddress = binary.BigEndian.Uint32(allocIP.Address.IP.To4())
+		keyContainer.Index = totalNum + 1
```
hzxuzhonghu, line 205: why +1

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,102 +14,174 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_SOCKHASH);
+	__type(key, struct bpf_sock_tuple);
+	__type(value, __u32);
+	__uint(max_entries, MAP_SIZE_OF_MANAGER);
+	__uint(map_flags, 0);
+} map_of_kmesh_hashmap SEC(".maps");
+
+static inline int kmesh_manager_by_ip(__u32 ip)
```
hzxuzhonghu, line 86:
```suggestion
static inline int is_managed_by_kmesh(__u32 ip)
```

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,102 +14,156 @@
...
+	// During the auth, src must set the client info and dst set
+	// the server info when we transmitted to the kmesh auth info.
+	// In this way, auth can be performed normally.
+	tuple_key.ipv4.saddr = skops->remote_ip4;
+	tuple_key.ipv4.daddr = skops->local_ip4;
+	tuple_key.ipv4.sport = skops->remote_port >> 16;
+	tuple_key.ipv4.dport = bpf_htonl(skops->local_port) >> 16;
+	long ret = bpf_map_delete_elem(&map_of_auth, &tuple_key);
+	if(ret && ret != -ENOENT)
+		BPF_LOG(INFO, SOCKOPS, "map_of_auth bpf_map_delete_elem failed, ret: %d\n", ret);
```
hzxuzhonghu, line 111: log the key to, is there a helper in c to convert struct to string

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,102 +14,156 @@
...
+	// tuple info src is server info, dst is client info
+	// During the auth, src must set the client info and dst set
+	// the server info when we transmitted to the kmesh auth info.
+	// In this way, auth can be performed normally.
+	(*msg).tuple.ipv4.saddr = skops->remote_ip4;
+	(*msg).tuple.ipv4.daddr = skops->local_ip4;
+	// remote_port is network byteorder
+	(*msg).tuple.ipv4.sport = skops->remote_port >> 16;
+	// local_port is host byteorder
+	(*msg).tuple.ipv4.dport = bpf_htonl(skops->local_port) >> 16;
```
hzxuzhonghu, line 132: Isn't this same with that in clean_auth_map

hzxuzhonghu (approved):
defer @supercharge-xsy  to have a last look

bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -14,102 +14,156 @@
...
+} map_of_kmesh_manager SEC(".maps");
 
-static inline void build_auth_key(struct bpf_sock_ops *skops, struct bpf_sock_tuple* key)
+struct {
+	__uint(type, BPF_MAP_TYPE_SOCKHASH);
+	__type(key, struct bpf_sock_tuple);
+	__type(value, __u32);
+	__uint(max_entries, MAP_SIZE_OF_MANAGER);
+	__uint(map_flags, 0);
+} map_of_kmesh_hashmap SEC(".maps");
```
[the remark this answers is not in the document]
bitcoffeeiux: This map is used to trigger the sendmsg program used by the TLV encode. Currently, it is not needed in this PR. I will move it to the codec PR.

bpf/kmesh/workload/sockops_tuple.c
bitcoffeeiux: C does not provide the function of converting struct to string by default. We need to implement it. The recorded error codes can be used to locate problems such as out of memory. However, it seems that recording keys is unnecessary because residual keys can be easily found by viewing maps in the system.

bpf/kmesh/workload/sockops_tuple.c
bitcoffeeiux: Yes, the logic is the same. Let's make a simple macro wrapper.

PR #156, opened 2024-03-07
set resource name when response for rds and eds type

supercharge-xsy:
…s and cds are WildcardType

**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #https://github.com/kmesh-net/kmesh/issues/144

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #181, opened 2024-03-21
performance optimisation 

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Optimised CreateApiListenerByLds function to reduce the system resource when ApiStatus is Unchange.


hzxuzhonghu (commented):
We cannot simply doing this change. because with https://github.com/kmesh-net/kmesh/pull/156, we rely on the parsing to calculate the route names need to be subscribed.

PR #187, opened 2024-03-26
Change `UpdateApiClusterStatus` in handleLdsResponse to `UpdateApiListenerStatus`

### LINKED ISSUE #185: Will the listener's address be nil in real use case?
**Please provide an in-depth description of the question you have**:
https://github.com/kmesh-net/kmesh/blob/e6f8c191b97293fa70d6a5ea00251622c21881d1/pkg/cache/v2/listener.go#L105-L129

Getaddress() method on line 112. Getaddress() method in line 112. Because it returns listener.Address, if listener.Address is nil, it will trigger panic.

```console
	t.Run("test3: listener in rsp is nil", func(t *testing.T) {
		adsLoader := NewAdsLoader()
		adsLoader.routeNames = []string{
			"ut-route-to-client",
			"ut-route-to-service",
		}
		svc := &ServiceEvent{
			DynamicLoader: adsLoader,
		}
		listener := &config_listener_v3.Listener{}
		anyListener, err := anypb.New(listener)
		assert.NoError(t, err)
		rsp := &service_discovery_v3.DiscoveryResponse{
			Resources: []*anypb.Any{
				anyListener,
			},
		}
		err = svc.handleLdsResponse(rsp)
		assert.NoError(t, err)
		apiMethod := svc.DynamicLoader.ListenerCache.GetApiListener("ut-listener").ApiStatus
		assert.Equal(t, core_v2.ApiStatus_UPDATE, apiMethod)
		wantHash := hash.Sum64String(anyListener.String())
		actualHash := svc.DynamicLoader.ListenerCache.GetLdsHash(listener.GetName())
		assert.Equal(t, wantHash, actualHash)
	})
```

The above problem was discovered when I wrote the above test case for SeviceEnevt.handleLdsResponse.

The question I want to ask is: Will there be a situation where the listenerCache in adsLoader is nil and the listener in rsp is also empty?

**What do you think about this question?**:
I think we should add nil pointer check.
**Environment**:
- Kmesh version: 0.2.0
- Others:

LiZhenCheng9527:
…erStatus

**What type of PR is this?**
/kind bug
Fixes #185

### DIFF
--- a/pkg/controller/envoy/ads_event.go
+++ b/pkg/controller/envoy/ads_event.go
@@ -229,7 +229,7 @@ func (svc *ServiceEvent) handleLdsResponse(rsp *service_discovery_v3.DiscoveryRe
 
 	removed := svc.DynamicLoader.ListenerCache.GetResourceNames().Difference(current)
 	for key := range removed {
-		svc.DynamicLoader.UpdateApiClusterStatus(key, core_v2.ApiStatus_DELETE)
+		svc.DynamicLoader.UpdateApiListenerStatus(key, core_v2.ApiStatus_DELETE)
 	}
 
 	svc.DynamicLoader.ListenerCache.Flush()


hzxuzhonghu (approved):
/lgtm

Issue #186, opened 2024-03-25
Kmesh does not handle Transitive Requests


tmodak27:
**What happened**:

(Refer to below diagram) Kmesh does not handle "transitive requests"; If the service which receives a request calls another service in the same context, Kmesh does not handle the second request.

<img width="731" alt="kmesh_routing_issue_1" src="https://github.com/kmesh-net/kmesh/assets/60729152/3a9876b0-6b3e-47a0-af92-7a443d2564d1">

**What you expected to happen**:

The second request should also have been handled by Kmesh

**How to reproduce it (as minimally and precisely as possible)**:

**1. Deploy Kmesh and Istio 1.19** 

**2. Deploy the 3 backend services**

Note: 

- These 3 backend services are written in python and they expose a GET API `/message` that returns the `message` string.
- The API takes an optional `dest` url parameter, which is the name of another service the user wants to forward the request to. If the user  does NOT pass a `dest`, the original service returns a response. Otherwise the request is forwarded to the service whose name is `dest`. 

For example

```

# user does not pass a dest
$ curl app-b:8000/hello_a
{"message":"Service app-a has received Message 'hello_a'","src":"","location":""}

# user passes a dest app-b
$ curl app-a:8000/hello_a?dest=app-b
{"message":"Service app-b has received Message 'hello_a'","src":"app-a","location":""}

```

```
apiVersion: v1
kind: Service
metadata:
  name: app-a
  labels:
    app: app-a
    service: app-a
spec:
  ports:
  - port: 8000
    name: httpa
  selector:
    app: app-a
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-a
  labels:
    app: app-a
spec:
  replicas: 1
  selector:
    matchLabels:
      app: app-a
  template:
    metadata:
      labels:
        app: app-a
    spec:
      containers:
      - name: app-a
        image: tm27/fastapi-samples:new
        env:
        - name: SERVICE_NAME
          value: "app-a"
        resources:
          requests:
            cpu: "100m"
        imagePullPolicy: IfNotPresent #Always
        ports:
        - containerPort: 8000
---

apiVersion: v1
kind: Service
metadata:
  name: app-b
  labels:
    app: app-b
    service: app-b
spec:
  ports:
  - port: 8000
    name: httpb
  selector:
    app: app-b
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-b
  labels:
    app: app-b
spec:
  replicas: 1
[...]

PR #191, opened 2024-03-27
fix kmesh http parse header bug

lec-bit:
**What type of PR is this?**
/kind bug

Fixed an issue with Kmesh failing to parse HTTP headers. When a field in the HTTP header has an empty value, Kmesh would fail to parse it, leading to routing issues. 

Fixes #https://github.com/kmesh-net/kmesh/issues/186

```release-note

```

### DIFF
--- a/kernel/ko_src/kmesh/kmesh_parse_http_1_1.c
+++ b/kernel/ko_src/kmesh/kmesh_parse_http_1_1.c
@@ -307,7 +307,10 @@ static bool parse_header(struct bpf_mem_ptr *context)
 					continue;
 				field_value_begin_position = i;
 				field_value_end_position = i;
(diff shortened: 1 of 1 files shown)


kernel/ko_src/kmesh/kmesh_parse_http_1_1.c
```diff
@@ -307,10 +307,13 @@ static bool parse_header(struct bpf_mem_ptr *context)
 					continue;
 				field_value_begin_position = i;
 				field_value_end_position = i;
-				current_state = ST_FIELD_VALUE;
+				if (ch == CR)
```
lec-bit, line 310: According to the HTTP protocol details in RFC2616, standalone CR or LF is not allowed, so CR is considered as the end of the line.

PR #192, opened 2024-03-27
add ut of cache/v2/route.go

### LINKED ISSUE #104: Improve unit test coverage

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

add ut of cache/v2/route.go
a part of #104


hzxuzhonghu (approved):
/lgtm

PR #194, opened 2024-03-28
add ut of cache/v2/listener.go

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

add ut of cache/v2/listener.go
a part of #104

### DIFF
--- a/pkg/cache/v2/listener_test.go
+++ b/pkg/cache/v2/listener_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 The Kmesh Authors.
+ * Copyright 2024 The Kmesh Authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pkg/cache/v2/listener_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	core_v2 "kmesh.net/kmesh/api/v2/core"
+	listener_v2 "kmesh.net/kmesh/api/v2/listener"
+	maps_v2 "kmesh.net/kmesh/pkg/cache/v2/maps"
+	"kmesh.net/kmesh/pkg/utils/hash"
+)
+
+func TestListenerFlush(t *testing.T) {
```
hzxuzhonghu, line 32: I am thinking you are testing `handleLdsResponse`, is that covered by those tests

LiZhenCheng9527: Yes, I'm testing if the listenercache is handled correctly when maps_v2.ListenerUpdate returns nil.
This covers the handleLdsResponse usage scenario.

pkg/cache/v2/listener_test.go
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright 2023 The Kmesh Authors.
```
hzxuzhonghu, line 2: 2024

hzxuzhonghu (approved):
/lgtm

PR #195, opened 2024-03-29
Fix POD_NAME env

hzxuzhonghu:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/deploy/helm/templates/daemonset.yaml
+++ b/deploy/helm/templates/daemonset.yaml
@@ -26,17 +26,14 @@ spec:
         - name: POD_NAME
           valueFrom:
             fieldRef:
-              apiVersion: v1
-              fieldPath: metadata.namespace
+              fieldPath: metadata.name
         - name: POD_NAMESPACE
           valueFrom:
             fieldRef:
-              apiVersion: v1
               fieldPath: metadata.namespace
         - name: INSTANCE_IP
           valueFrom:
             fieldRef:
-              apiVersion: v1
               fieldPath: status.podIP
         - name: MESH_CONTROLLER
           value: {{ quote .Values.deploy.kmesh.env.meshController }}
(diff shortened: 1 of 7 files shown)


hzxuzhonghu:
Tested with @lec-bit 

cc @YaoZengzeng @nlgwcy

deploy/helm/templates/daemonset.yaml
```diff
@@ -26,24 +26,31 @@ spec:
...
         - name: MESH_CONTROLLER
           value: {{ quote .Values.deploy.kmesh.env.meshController }}
         - name: BPF_LOG_SIZE
           value: {{ quote .Values.deploy.kmesh.env.bpfLogSize }}
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ quote .Values.kubernetesClusterDomain }}
+        - name: SERVICE_ACCOUNT
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
```
YaoZengzeng, line 47: `apiVersion` here could also be deleted?

pkg/controller/config/config.go
```diff
@@ -93,12 +108,37 @@ func (c *XdsConfig) Init() error {
 
 	log.Infof("service node %v connect to discovery address %v", c.ServiceNode, c.DiscoveryAddress)
 
+	c.Metadata.Namespace = podNamespace
+	c.Metadata.ClusterID = cluster.ID(clusterID)
+	c.Metadata.InstanceIPs = []string{ip}
+	// TODO: add labels to support localiy load balancing
+	c.Metadata.Labels = nil
+	c.Metadata.MeshID = "cluster.local"
```
YaoZengzeng, line 116: Should `MeshID` also be allowed to get from env or command line flag?

hzxuzhonghu: sgtm

deploy/helm/templates/daemonset.yaml
hzxuzhonghu: i think so, this is an optional field, and default to `v1`

hzxuzhonghu (commented):
@nlgwcy have a look when you are free

PR #197, opened 2024-03-29
add ut of cache/v2/cluster.go

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

add ut of cache/v2/cluster.go
a part of #104

### DIFF
--- a/pkg/cache/v2/cluster_test.go
+++ b/pkg/cache/v2/cluster_test.go
@@ -31,12 +31,17 @@ import (
 
 func TestClusterFlush(t *testing.T) {
 	t.Run("cluster status is UPDATE", func(t *testing.T) {
+		updateClusterName := []string{}
+		deleteClusterName := []string{}
+
 		patches1 := gomonkey.NewPatches()
 		patches2 := gomonkey.NewPatches()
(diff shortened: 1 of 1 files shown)


pkg/cache/v2/cluster_test.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		cluster2 := &cluster_v2.Cluster{
+			ApiStatus:      core_v2.ApiStatus_UPDATE,
+			Name:           "ut-cluster2",
+			ConnectTimeout: uint32(60),
+			LbPolicy:       cluster_v2.Cluster_ROUND_ROBIN,
+		}
+		cache.SetApiCluster(cluster1.Name, cluster1)
+		cache.SetApiCluster(cluster2.Name, cluster2)
+		cache.Flush()
```
hzxuzhonghu, line 62: I mean we should know that it really does write bpf map.

LiZhenCheng9527: I think it's up to the UT of `map_v2.ClusterUpdate`, to test if it actually writes the bpf map

hzxuzhonghu: I mean according to the apiStatus, we should know whether this handler work as expected.  Am i clear?

hzxuzhonghu (approved):
/lgtm

PR #200, opened 2024-04-02
Fixed incorrect function usage when deleting routes

LiZhenCheng9527:
Fixed incorrect function usage when deleting routes


hzxuzhonghu (approved):
/lgtm

PR #204, opened 2024-04-02
add ut of workload_stream.go

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

add ut of workload_stream.go
a part of #104

### DIFF
--- a/pkg/controller/workload/workload_controller_test.go
+++ b/pkg/controller/workload/workload_controller_test.go
@@ -1,0 +1,269 @@
+/*
+ * Copyright 2024 The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package workload
+
+import (
+	"context"
+	"errors"
+	"net"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/agiledragon/gomonkey/v2"
+	config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	discoveryv3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	cluster_v2 "kmesh.net/kmesh/api/v2/cluster"
+	core_v2 "kmesh.net/kmesh/api/v2/core"
+	"kmesh.net/kmesh/pkg/auth"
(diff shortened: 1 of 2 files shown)


pkg/controller/workload/workload_stream_test.go
```diff
@@ -0,0 +1,301 @@
...
+	t.Cleanup(func() {
+		grpcServer.Stop()
+		close(s.close)
+	})
+	s.Listener = listener
+	return s
+}
+
+func (f *MockDiscovery) StreamAggregatedResources(server discoveryv3.AggregatedDiscoveryService_StreamAggregatedResourcesServer) error {
+	numberOfSends := 0
```
hzxuzhonghu, line 68: what is this value for? I do not see it's checked

pkg/controller/workload/workload_stream_test.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/agiledragon/gomonkey/v2"
+	config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	discoveryv3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/test/bufconn"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+type MockDiscovery struct {
```
hzxuzhonghu, line 36: Can we import istio? I think we have done that for ads test

pkg/controller/workload/workload_stream_test.go
```diff
@@ -0,0 +1,301 @@
...
+		{
+			name:       "test2: no movk, send request successful, should return nil",
+			beforeFunc: func() {},
+			afterFunc:  func() {},
+			wantErr:    false,
+		},
+		{
+			name: "test3: fail to create workloadStream, should return error",
+			beforeFunc: func() {
+				patches1.ApplyMethod(reflect.TypeOf(client), "DeltaAggregatedResources",
```
hzxuzhonghu, line 160: I think you donot need to patch here, you can just put this resource into `f.responses`.

pkg/controller/workload/workload_stream_test.go
LiZhenCheng9527: This is because the workloadapi of kmesh is used in the package workload, and there is a conflict with istio.workload in the `TunnelProtocol`

```console
proto: file "api/workloadapi/workload.proto" has a name conflict over istio.workload.TunnelProtocol
```

pkg/controller/workload/workload_stream_test.go
LiZhenCheng9527: It is refer to https://github.com/istio/istio/blob/master/pilot/test/xdstest/mock_discovery.go

pkg/controller/workload/workload_stream_test.go
hzxuzhonghu: I've met this error during go build, and resolve it by https://github.com/kmesh-net/kmesh/pull/195/files

Not a must though

pkg/controller/workload/workload_stream_test.go
LiZhenCheng9527: This test case is to cover the case when creating a stream fails. How can I use `f.responses` to cause stream creation to fail?

hzxuzhonghu: If create stream failed, mostly because of auth/network
But this test does not include auth or network

PR #206, opened 2024-04-07
fix no responseNonce in AdsRequest

### LINKED ISSUE #201: Kmesh eds/rds new discovery request should include the `ResponseNonce` last time received
**What happened**:

When we construct a new request for eds because of a new cluster added, we donot set the `ResponseNonce`
https://github.com/kmesh-net/kmesh/blob/6165d954a9c2a9f19332a22a53dd8e947f5fa638/pkg/controller/envoy/ads_event.go#L67-L74

According to API document, we should set the response nonce in this case.

```
	// nonce corresponding to DiscoveryResponse being ACK/NACKed. See above
	// discussion on version_info and the DiscoveryResponse nonce comment. This
	// may be empty only if 1) this is a non-persistent-stream xDS such as HTTP,
	// or 2) the client has not yet accepted an update in this xDS stream (unlike
	// delta, where it is populated only for new explicit ACKs).
```

Otherwise, in istiod, this request will be ignored, and we will never be able to subscribe to new eds

FYI  https://github.com/istio/istio/blob/4ee56447daf484a90d92376703519b3b3fc6bf7e/pilot/pkg/xds/ads.go#L427-L449

**Environment**:
- Kmesh version:
- Others:

LiZhenCheng9527:
**What type of PR is this?**
/kind bug

fix no responseNonce in AdsRequest
Fixes #201

### DIFF
--- a/pkg/controller/envoy/ads_event.go
+++ b/pkg/controller/envoy/ads_event.go
@@ -41,11 +41,18 @@ const (
 	maxAdminRequests = 16
 )
 
+type lastNonce struct {
+	cdsNonce string
+	edsNonce string
+	ldsNonce string
+	rdsNonce string
+}
 type ServiceEvent struct {
 	DynamicLoader *AdsLoader
 	ack           *service_discovery_v3.DiscoveryRequest
 	rqt           *service_discovery_v3.DiscoveryRequest
 	adminChan     chan *admin_v2.ConfigResources
+	LastNonce     *lastNonce
 }
 
 func NewServiceEvent() *ServiceEvent {
@@ -54,6 +61,16 @@ func NewServiceEvent() *ServiceEvent {
 		ack:           nil,
 		rqt:           nil,
 		adminChan:     make(chan *admin_v2.ConfigResources, maxAdminRequests),
+		LastNonce:     NewLastNonce(),
+	}
+}
+
+func NewLastNonce() *lastNonce {
+	return &lastNonce{
+		cdsNonce: "",
+		edsNonce: "",
+		ldsNonce: "",
+		rdsNonce: "",
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/envoy/ads_event.go
```diff
@@ -162,20 +162,20 @@ func (svc *ServiceEvent) handleCdsResponse(rsp *service_discovery_v3.DiscoveryRe
 
 	// when the list of eds typed clusters subscribed changed, we should resubscrbe to new eds.
 	if !slices.EqualUnordered(svc.DynamicLoader.edsClusterNames, lastEdsClusterNames) {
-		svc.rqt = newAdsRequest(resource_v3.EndpointType, svc.DynamicLoader.edsClusterNames)
+		svc.rqt = newAdsRequest(resource_v3.EndpointType, svc.DynamicLoader.edsClusterNames, resp.Nonce)
```
hzxuzhonghu, line 165: This is not right. 

We should pass the eds nonce rather than the cds nonce. If this is the first req for eds, we can pass empty

pkg/controller/envoy/ads_event.go
```diff
@@ -197,20 +197,20 @@ func (svc *ServiceEvent) handleEdsResponse(rsp *service_discovery_v3.DiscoveryRe
 		svc.ack.ResourceNames = append(svc.ack.ResourceNames, loadAssignment.GetClusterName())
 	}
 
-	svc.rqt = newAdsRequest(resource_v3.ListenerType, nil)
+	svc.rqt = newAdsRequest(resource_v3.ListenerType, nil, resp.Nonce)
```
hzxuzhonghu, line 200: Same as above.

Another point from me: we donot need to send lds request if it has been sent once

pkg/controller/envoy/ads_event.go
```diff
@@ -235,19 +235,19 @@ func (svc *ServiceEvent) handleLdsResponse(rsp *service_discovery_v3.DiscoveryRe
 	svc.DynamicLoader.ListenerCache.Flush()
 
 	if !slices.EqualUnordered(svc.DynamicLoader.routeNames, lastRouteNames) {
-		svc.rqt = newAdsRequest(resource_v3.RouteType, svc.DynamicLoader.routeNames)
+		svc.rqt = newAdsRequest(resource_v3.RouteType, svc.DynamicLoader.routeNames, resp.Nonce)
```
hzxuzhonghu, line 238: ditto

pkg/controller/envoy/ads_event.go
```diff
@@ -94,35 +111,39 @@ func newAckRequest(rsp *service_discovery_v3.DiscoveryResponse) *service_discove
...
+	svc.ack = newAckRequest(resp)
+	if resp.GetResources() == nil {
 		return
 	}
 
-	switch rsp.GetTypeUrl() {
+	switch resp.GetTypeUrl() {
 	case resource_v3.ClusterType:
-		err = svc.handleCdsResponse(rsp)
+		svc.LastNonce.cdsNonce = resp.Nonce
```
hzxuzhonghu, line 126: I would move it down into each handler handleCdsResponse

hzxuzhonghu (approved):

hzxuzhonghu (commented):
/lgtm

PR #212, opened 2024-04-09
fix go test env variable invalidation

LiZhenCheng9527:
fix go test env variable invalidation


.github/workflows/main.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       shell: bash
       run: |
         sudo env "PATH=$PATH" bash ./build.sh
-
+    
     - name: Setup Enviroments
       run: |
         echo "PKG_CONFIG_PATH=$GITHUB_WORKSPACE/mk" >> $GITHUB_ENV
```
LiZhenCheng9527, line 32: save this because golangci-lint requires `api-v2-c`

.github/workflows/main.yml
```diff
@@ -40,4 +40,4 @@ jobs:
 
     - name: Go Test
       run: |
-        sudo env LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib:$GITHUB_WORKSPACE/api/v2-c:$GITHUB_WORKSPACE/bpf/deserialization_to_bpf_map go test -v -vet=off ./pkg/...
+        sudo env LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib:$GITHUB_WORKSPACE/api/v2-c:$GITHUB_WORKSPACE/bpf/deserialization_to_bpf_map PKG_CONFIG_PATH=$GITHUB_WORKSPACE/mk go test -v -vet=off ./pkg/...
```
hzxuzhonghu, line 43: isn't GITHUB_ENV a way to set global env?

.github/workflows/main.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       shell: bash
       run: |
         sudo env "PATH=$PATH" bash ./build.sh
-
+
```
hzxuzhonghu, line 29: restore

pkg/controller/workload/workload_cache_test.go
```diff
@@ -0,0 +1,138 @@
+/*
```
hzxuzhonghu, line 1: separate this test with another pr

hzxuzhonghu (approved):
/lgtm

PR #219, opened 2024-04-12
fix kmesh.ko location

lec-bit:
**What type of PR is this?**

In the startup script in the kmesh image, the location of the copied kmesh.ko is incorrect and cannot be loaded later. Fix and adjust the location of kmesh.ko.
Fixes #

```release-note

```

### DIFF
--- a/build/docker/start_kmesh.sh
+++ b/build/docker/start_kmesh.sh
@@ -4,8 +4,7 @@
 
 lsmod | grep kmesh > /dev/null
 if [ $? -ne 0 ] && [ -f kmesh.ko ]; then
-        mkdir -p /lib/modules/kmesh
-        cp kmesh.ko /lib/modules/kmesh
+	cp kmesh.ko /lib/modules/$(uname -r)
         depmod -a
         modprobe kmesh
 fi


build/docker/start_kmesh.sh
```diff
@@ -4,8 +4,7 @@
 
 lsmod | grep kmesh > /dev/null
 if [ $? -ne 0 ] && [ -f kmesh.ko ]; then
-        mkdir -p /lib/modules/kmesh
-        cp kmesh.ko /lib/modules/kmesh
+	cp kmesh.ko /lib/modules/$(uname -r)
```
LiZhenCheng9527, line 7: Why do I need to make this change, and what's wrong with copying to /lib/modules/kmesh?

lec-bit: The modprobe kmesh command can search for and load ko only in /lib/modules/$(uname -r).
In the past, a Symbolic Link of kmesh.ko is created in the directory when the docker is made locally. Therefore, this problem is not found,
Actually, ko cannot be loaded when the kmesh image is directly pulled.

Issue #203, opened 2024-04-02
It looks like we need to support asynchronous refresh bpf map operations.


supercharge-xsy:
**What would you like to be added**:

The BPF map can be updated asynchronously when xDS messages are processed.

**Why is this needed**:
Map update is time-consuming. Synchronous update may cause istiod push exceptions hen the configuration is frequently updated:

like this “Expired nonce received ”
```
info	ads	RDS: PUSH for node:kmesh-p5glc.kmesh-system resources:13 size:11.6kB cached:0/13 nonce:o5ib5q+RXAQ=13ec70c5-a30d-4711-98f2-fb999c807b1d version:2024-03-29T09:33:18Z/23
debug	ads	ADS:RDS: REQ kmesh-p5glc.kmesh-system-22 Expired nonce received mzV8zXBMXxg=faa5988c-66eb-4c78-9397-f2547e8a3f84, sent o5ib5q+RXAQ=13ec70c5-a30d-4711-98f2-fb999c807b1d
debug	ads	ADS:RDS: REQ kmesh-p5glc.kmesh-system-22 Expired nonce received o5ib5q+RXAQ=13ec70c5-a30d-4711-98f2-fb999c807b1d, sent ynDoAQuRIFw=f2c93f7c-5e85-4368-8db1-c4c841d4f5e5
```

LiZhenCheng9527:
I think the most time consuming point when refreshing bpf map is the update of ebpf maps. So we can use timed sync instead of immediate sync. Set the time interval to the approximate time it takes for bpf maps to refresh once.

PR #221, opened 2024-04-12
Support for asynchronous refresh of ads map

supercharge-xsy:
**What type of PR is this?**

-->

Fixes #https://github.com/kmesh-net/kmesh/issues/203

```release-note

```


pkg/controller/envoy/ads_event.go
```diff
@@ -182,7 +182,7 @@ func (svc *ServiceEvent) handleCdsResponse(resp *service_discovery_v3.DiscoveryR
 	if !slices.EqualUnordered(svc.DynamicLoader.edsClusterNames, lastEdsClusterNames) {
 		svc.rqt = newAdsRequest(resource_v3.EndpointType, svc.DynamicLoader.edsClusterNames, svc.LastNonce.edsNonce)
 	} else {
-		svc.DynamicLoader.ClusterCache.Flush()
+		go svc.DynamicLoader.ClusterCache.Flush()
```
LiZhenCheng9527, line 185: If updating the ads map using asynchronous refresh, you need to remove the command in ut as follow.
```console
assert.Equal(t, sets.Set[string]{"ut-cluster": sets.Empty{}}, clusterNames)
```

hzxuzhonghu (commented):
in the Flush function, it hold the lock from beginning to end, so i am concerned this won't help mitigate much

supercharge-xsy:
> in the Flush function, it hold the lock from beginning to end, so i am concerned this won't help mitigate much

you are right，asynchronous refresh does not completely solve the problem，the next step is to optimize the flush performance.

pkg/controller/envoy/ads_event.go
supercharge-xsy: got

hzxuzhonghu:
So i think we can reduce the lock holding time on order to make it async. This can be done by copying a temporary cache within the lock, then we donot need to lock with the lock. 

The only one question to confirm, does the bpf map update need to be locked?

PR #229, opened 2024-04-16
automatically mount cgroup2 at kmesh startup

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

automatically mount cgroup2 at kmesh startup

### DIFF
--- a/pkg/bpf/bpf.go
+++ b/pkg/bpf/bpf.go
@@ -158,6 +158,7 @@ func mountCgroup2(cfg *Config) error {
 	}
 
 	if err := syscall.Mount("none", cfg.Cgroup2Path, "cgroup2", 0, ""); err != nil {
+		cleanupMountPath()
 		log.Errorf("failed to mount %s: %v", cfg.Cgroup2Path, err)
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/bpf/bpf.go
```diff
@@ -89,6 +89,10 @@ func Start() error {
 		return err
 	}
 
+	if err = mountCgroup2(&config); err != nil {
+		return err
```
YaoZengzeng, line 93: AFAIK, cgroup v2 only available in linux kernel v5.8 or above. 

Is mouting cgroup v2 a necessary step use kmesh? If not we could just print a warning message and not return error directly, otherwise kmesh could not run on older versions of the kernel. 

I guess v5.8 is a pretty high version for production environments.

LiZhenCheng9527: https://github.com/kmesh-net/kmesh/blob/main/build/docker/README.md#start_kmeshsh

It is stated that to start kmesh, have to mount cgroupv2.

hzxuzhonghu (commented):
/assign @bitcoffeeiux

pkg/bpf/bpf.go
```diff
@@ -142,3 +150,36 @@ func Stop() {
...
+
+func mountCgroup2(cfg *Config) error {
+	if err := os.MkdirAll(cfg.Cgroup2Path, 0755); err != nil {
+		log.Errorf("failed to create dir %s: %v", cfg.Cgroup2Path, err)
+		return err
+	}
+
+	if err := syscall.Mount("none", cfg.Cgroup2Path, "cgroup2", 0, ""); err != nil {
+		log.Errorf("failed to mount %s: %v", cfg.Cgroup2Path, err)
+		return err
```
supercharge-xsy, line 162: need clean up resources before return err

PR #232, opened 2024-04-16
Optimised UT with test bpf map in ut

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Optimised UT with test bpf map in ut
a part of #104


hzxuzhonghu (approved):
/lgtm

We need benchmark on `flush`

PR #236, opened 2024-04-17
Enable workload mode by default

hzxuzhonghu:
This is changing the --mode from ads to workload,  since we are in very early stage, i think it is ok.

### DIFF
--- a/pkg/bpf/bpf.go
+++ b/pkg/bpf/bpf.go
@@ -89,11 +89,11 @@ func Start() error {
 		return err
 	}
 
-	if config.EnableKmesh {
+	if config.AdsEnabled() {
 		if err = StartKmesh(); err != nil {
 			return err
 		}
-	} else if config.EnableKmeshWorkload {
+	} else if config.WdsEnabled() {
 		if err = StartKmeshWorkload(); err != nil {
 			return err
 		}
@@ -123,12 +123,12 @@ func StopMda() error {
 func Stop() {
 	var err error
 
-	if config.EnableKmesh {
+	if config.AdsEnabled() {
 		if err = Obj.Kmesh.Detach(); err != nil {
 			log.Errorf("failed detach when stop kmesh, err:%s", err)
 			return
 		}
-	} else if config.EnableKmeshWorkload {
+	} else if config.WdsEnabled() {
 		if err = ObjWorkload.KmeshWorkload.Detach(); err != nil {
 			log.Errorf("failed detach when stop kmesh, err:%s", err)
 			return
--- a/pkg/bpf/config.go
+++ b/pkg/bpf/config.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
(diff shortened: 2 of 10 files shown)


pkg/bpf/config.go
```diff
@@ -40,31 +40,23 @@ func init() {
...
+	Cgroup2Path      string
+	EnableMda        bool
+	BpfVerifyLogSize int
 }
 
 func (c *Config) AttachFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&c.BpfFsPath, "bpf-fs-path", "/sys/fs/bpf", "bpf fs path")
 	cmd.PersistentFlags().StringVar(&c.Cgroup2Path, "cgroup2-path", "/mnt/kmesh_cgroup2", "cgroup2 path")
-	cmd.PersistentFlags().StringVar(&c.Mode, "mode", "ads", "controller plane mode, valid values are [ads, workload]")
+	cmd.PersistentFlags().StringVar(&c.Mode, "mode", "workload", "controller plane mode, valid values are [ads, workload]")
```
YaoZengzeng, line 53: To prevent misconfiguration, it would be better to check if the `mode` value is expected. 

Because other code depends on this value, it can only be `ads` or `workload`.

pkg/bpf/bpf.go
```diff
@@ -89,11 +89,11 @@ func Start() error {
...
-	if config.EnableKmesh {
+	if config.AdsEnabled() {
 		if err = StartKmesh(); err != nil {
 			return err
 		}
-	} else if config.EnableKmeshWorkload {
+	} else if config.WdsEnabled() {
 		if err = StartKmeshWorkload(); err != nil {
 			return err
 		}
```
YaoZengzeng, line 99: It would be better to return error, if config is neither ads or workload. It would be more robust in `start()` stage

pkg/controller/client.go
```diff
@@ -73,12 +72,12 @@ func (c *XdsClient) createStreamClient() error {
...
 	c.client = service_discovery_v3.NewAggregatedDiscoveryServiceClient(c.grpcConn)
 
-	if bpfConfig.EnableKmesh {
+	if bpfConfig.AdsEnabled() {
 		if err = c.AdsStream.AdsStreamCreateAndSend(c.client, c.ctx); err != nil {
 			_ = c.grpcConn.Close()
 			return fmt.Errorf("create ads stream failed, %s", err)
 		}
-	} else if bpfConfig.EnableKmeshWorkload {
+	} else if bpfConfig.WdsEnabled() {
```
LiZhenCheng9527, line 80: If the default is to use the mode of workload, is it more efficient to first determine `bpfconfig.WdsEnable` or not.

pkg/bpf/config.go
hzxuzhonghu: we can add a validation in the parse funcction

PR #241, opened 2024-04-17
update gitignore because bpf2go update

LiZhenCheng9527:
update gitignore because bpf2go update


hzxuzhonghu (approved):
/lgtm

PR #246, opened 2024-04-18
add benchmark of bpf write test

LiZhenCheng9527:
add benchmark of bpf write test.
Including cluster map write  test and listener map write test.

### DIFF
--- a/pkg/controller/envoy/ads_event_test.go
+++ b/pkg/controller/envoy/ads_event_test.go
@@ -21,17 +21,27 @@ import (
 	"os"
 	"syscall"
 	"testing"
+	"time"
 
+	accesslogv3 "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
 	config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (commented):
Please https://pkg.go.dev/testing#hdr-Benchmarks check for benchmark test

pkg/controller/envoy/ads_event_test.go
```diff
@@ -772,3 +884,56 @@ func TestHandleRdsResponse(t *testing.T) {
...
+		svc.LastNonce.rdsNonce = "utLdstoRds"
+		listener.Name = rand.String(6)
+		anyListener, err := anypb.New(listener)
+		assert.NoError(t, err)
+		rsp := &service_discovery_v3.DiscoveryResponse{
+			Resources: []*anypb.Any{
+				anyListener,
+			},
+		}
+		err = svc.handleLdsResponse(rsp)
```
hzxuzhonghu, line 934: we should only test the `flush`

hzxuzhonghu (commented):
Can you add more ~10 endpoints and~ 5 filterchains and paste the test results

PR #252, opened 2024-04-19
Kmesh: adapt ads ebpf program for cni

bitcoffeeiux:
**What type of PR is this?**

Fixes #

```release-note

```


hzxuzhonghu (commented):
I think they are same with workload mode, can we split these snippet into a separate file, and then both modes can include it?

PR #256, opened 2024-04-19
image update

lec-bit:
**What type of PR is this?**

To select the corresponding binary file based on the kernel version when the image is started.
Fixes #

```release-note

```


kmesh_compile.sh
```diff
@@ -46,20 +46,25 @@ function copy_to_host() {
     else
         arch="aarch64"
     fi
+    
+    local kernel="normal"
```
LiZhenCheng9527, line 50: What kernel does normal kernel refer to, linux kernel 5.10 and above?
A comment could be added to clarify this.

PR #260, opened 2024-04-22
Enable xdp auth when kmesh works in workload mode

hzxuzhonghu:
**What type of PR is this?**

/kind enhancement

By passing working mode as cni configuration. With this, the config now looks like below:

```
    {
      "kubeConfig": "/etc/cni/net.d/kmesh-cni-kubeconfig",
      "mode": "ads",
      "type": "kmesh-cni"
    }

```

Fixes #

```release-note

```

### DIFF
--- a/pkg/cni/chained.go
+++ b/pkg/cni/chained.go
@@ -26,9 +26,9 @@ import (
 	"path/filepath"
 	"sort"
 
-	"kmesh.net/kmesh/pkg/utils"
-
 	"github.com/containernetworking/cni/libcni"
+
+	"kmesh.net/kmesh/pkg/utils"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


hzxuzhonghu:
@lec-bit @bitcoffeeiux

pkg/cni/chained.go
```diff
@@ -124,6 +124,7 @@ func insertCNIConfig(oldconfig []byte) ([]byte, error) {
 	// add kmesh-cni configuration
 	kmeshConfig["type"] = kmeshCniPluginName
 	kmeshConfig["kubeConfig"] = kubeconfigFilepath
+	kmeshConfig["mode"] = mode // provide mode here, so that kmesh-cni can decide how to run
```
LiZhenCheng9527, line 127: I think it's better for the comment to explain the two modes that are now supported

PR #272, opened 2024-04-25
fix authz on server listening on both ipv4 and ipv6 port

hzxuzhonghu:
THis is a bug found by accessing hello world application from istio sample

### DIFF
--- a/bpf/kmesh/workload/sockops_tuple.c
+++ b/bpf/kmesh/workload/sockops_tuple.c
@@ -173,12 +173,15 @@ static inline bool conn_from_cni_sim_delete(struct bpf_sock_ops *skops)
 			(bpf_ntohl(skops->remote_port) == 0x3a2));
 }
 
+static inline bool ipv4_mapped_addr(__u32 ip6[4])
+{
+	return ip6[0] == 0 && ip6[1] == 0 && ip6[2] == 0xFFFF0000;
+}
 
 SEC("sockops")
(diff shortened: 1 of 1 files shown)


bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -177,9 +177,6 @@ static inline bool conn_from_cni_sim_delete(struct bpf_sock_ops *skops)
 SEC("sockops")
 int record_tuple(struct bpf_sock_ops *skops)
 {
-	// only support IPV4
-	if (skops->family != AF_INET)
```
nlgwcy, line 181: 1. kmesh supports only IPv4.
2. In all-zero port listening scenario, socket on the server can process both IPv4 and IPv6 packets. When the server receives an IPv4 packet, the address is mapped to the IPv6 format (IPv4-mapped IPv6 address). However, the family may be AF_INET6. As a result, an error occurs.
Recommended solution:
1. IPv4-mapped IPv6 addresses
image
if (skops->family != AF_INET) -> if (skops->family != AF_INET || (IPV4-MAPPED_ADDR(skops->local_ip6))

hzxuzhonghu: good to know.

Can you check only skops->local_ip4?

hzxuzhonghu: @nlgwcy ITested, the 3rd u32 is 0xfffff0000

PR #273, opened 2024-04-25
Asynchronous execution of map flush and cache flush

LiZhenCheng9527:
**What type of PR is this?**

Asynchronous execution of map flush and cache flush
#203

### DIFF
--- a/pkg/cache/v2/cluster.go
+++ b/pkg/cache/v2/cluster.go
@@ -128,12 +128,17 @@ func (cache *ClusterCache) BpfMapFlush() {
 				case core_v2.ApiStatus_UPDATE:
 					if err := maps_v2.ClusterUpdate(name, cluster); err != nil {
 						log.Errorf("cluster %s update failed: %v", name, err)
+						// Rollback on update failure
+						cache.apiClusterCache = cacheDeepCopy.apiClusterCache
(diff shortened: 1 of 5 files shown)


pkg/cache/v2/cluster.go
```diff
@@ -105,28 +105,71 @@ func (cache *ClusterCache) SetEdsHash(key string, value uint64) {
...
+func (cache *ClusterCache) BpfMapFlush() {
+	for {
+		// if cacheDeepCopy not equal cache, need flush
+		if !clusterEqualFromHash(cacheDeepCopy, cache) {
+			cacheDeepCopy = cache.DeepCopy()
+			for name, cluster := range cacheDeepCopy.apiClusterCache {
+				switch cluster.GetApiStatus() {
+				case core_v2.ApiStatus_UPDATE:
+					if err := maps_v2.ClusterUpdate(name, cluster); err != nil {
+						log.Errorf("cluster %s update failed: %v", name, err)
```
YaoZengzeng, line 130: If we failed to update or delete, how to ensure consistency?

hzxuzhonghu (commented):
Can you write a proposal, no need to to very complex. Just an introduction of current xds response handling

PR #278, opened 2024-04-28
rename service_id to upstream_id in frontend_value struct

### LINKED ISSUE #277: frontend data structure adjustment in the kmesh workload mode
1.the name of service_id variable in frontend value has limitations and needs to be renamed;
2.the port variable in the frontend key seems to be redundant and needs to be removed;
3.docs of workload mode need to be refreshed.
Plan to handle the above issues in three separate PR
**Why is this needed**:

kwb0523:
**What type of PR is this?**

Fixes #277

```release-note

```


hzxuzhonghu (approved):
This is kind of cleanup
/lgtm

PR #286, opened 2024-04-29
Fix the issue of incorrect domain matching

### LINKED ISSUE #244: Kmesh does not handle request sent from Java app
image

**What happened**:
I have two services: B and C which are written in Java. I use curl to send a http request to B and then B will send a http request to C. 

what I suspect is kmesh could handle request from curl -> B, while kmesh didn't pick up the request sent from B -> C. I tried with different java http libiaries including restTemplate and httpclient, but neither works.

If I send a request directly from curl -> C, Kmesh could handle it

**I tried with same apps that written in Python and Golang and works as expected.**

logs: for B -> C

image

I'm not sure if this caused by coreDNS. but in service B, I use `app-a.default.svc.cluster.local` as the uri

Kmesh could handle request from curl -> B and B -> C.

app-a code:
```java
package org.example;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.http.ResponseEntity;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RestController;
import org.springframework.web.client.RestTemplate;

@SpringBootApplication
public class ConsumerApplication {
    public static void main(String[] args) {
        SpringApplication.run(ConsumerApplication.class, args);
    }
}

@RestController
class ConsumerController {
    @RequestMapping("/echo")
    public String echo() {
        RestTemplate restTemplate = new RestTemplate();
        ResponseEntity<String> response = restTemplate.getForEntity("http://app-c.default.svc.cluster.local:6001/echo", String.class);
        if (response.getStatusCode().is2xxSuccessful()) {
            String body = response.getBody();
            return body;
        } else {
            return "request error";
        }
    }
}
```

app-c code
```java
package org.example;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.PathVariable;
import org.springframework.web.bind.annotation.RestController;

@SpringBootApplication
public class ProviderApplication {
    public static void main(String[] args) {
        SpringApplication.run(ProviderApplication.class, args);
    }
}

@RestController
class ProviderController {
    @GetMapping("/echo")
    public String echo() {
        return "Provider received msg V1";
    }
}
```

deploy the provider in the cluster with 2 different versions, 
```yaml
apiVersion: v1
kind: Service
metadata:
  name: app-b
  labels:
    app: app-b
    service: app-b
spec:
  ports:
  - port: 6002
    name: http
  selector:
    app: app-b
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-b
  labels:
    app: app-b
spec:
  selector:
    matchLabels:
      app: app-b
  template:
    metadata:
      labels:
        app: app-b
    spec:
      containers:
      - name: consumer
        image: swr.cn-north-4.myhuaweicloud.com/hwstaff_pub_torontodsdelab/consumer:v7
        resources:
          requests:
            cpu: "100m"
        imagePullPolicy: Always
        ports:
        - containerPort: 6002
---
apiVersion: v1
kind: Service
metadata:
  name: app-c
  labels:
    app: app-c
    service: app-c
spec:
  ports:
  - port: 6001
    name: http
  selector:
    app: app-c
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-c
  labels:
    app: app-c
spec:
  selector:
    matchLabels:
      app: app-c
  template:
    metadata:
      labels:
        app: app-c
    spec:
      containers:
      - name: app-c
        image: swr.cn-north-4.myhuaweicloud.com/hwstaff_pub_torontodsdelab/provider:v2
        resources:
          requests:
            cpu: "100m"
        imagePullPolicy: IfNotPresent #Always
        ports:
        - containerPort: 6001
```

apply the following virtualservice to redirect the request from provider to provider-v1

virtual service:
```yaml
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: route
spec:
  hosts:
  - app-c.default.svc.cluster.local
  http:
  - route:
    - destination:
        host: app-c.default.svc.cluster.local
```

send request to app-b:
```bash
$ kubectl exec -it netutils -- curl http://app-c.default.svc.cluster.local:6001/echo
```

**Environment**:
- Kmesh version: Latest
- Ubuntu 22.04 with enhanced kernal
- CCE 1.27 with coreDNS plugin
- Istio 1.19

lec-bit:
**What type of PR is this?**
/kind bug

The problem was that using URI to match with domain led to the inability to find the correct route. The solution is to use host matching instead.
Fixes #244 

```release-note

```

### DIFF
--- a/bpf/kmesh/include/route_config.h
+++ b/bpf/kmesh/include/route_config.h
@@ -43,7 +43,7 @@ static inline Route__RouteConfiguration *map_lookup_route_config(const char *rou
 }
 
 static inline int
-virtual_host_match_check(Route__VirtualHost *virt_host, address_t *addr, ctx_buff_t *ctx, struct bpf_mem_ptr *uri)
+virtual_host_match_check(Route__VirtualHost *virt_host, address_t *addr, ctx_buff_t *ctx, struct bpf_mem_ptr *host)
 {
     int i;
     void *domains = NULL;
@@ -51,14 +51,14 @@ virtual_host_match_check(Route__VirtualHost *virt_host, address_t *addr, ctx_buf
     void *ptr;
     __u32 ptr_length;
 
-    if (!uri)
+    if (!host)
         return 0;
 
-    ptr = _(uri->ptr);
+    ptr = _(host->ptr);
     if (!ptr)
         return 0;
 
-    ptr_length = _(uri->size);
+    ptr_length = _(host->size);
 
     if (!virt_host->domains)
         return 0;
@@ -105,8 +105,8 @@ virtual_host_match(Route__RouteConfiguration *route_config, address_t *addr, ctx
     void *ptrs = NULL;
     Route__VirtualHost *virt_host = NULL;
     Route__VirtualHost *virt_host_allow_any = NULL;
-    char uri_key[4] = {'U', 'R', 'I', '\0'};
-    struct bpf_mem_ptr *uri;
+    char host_key[5] = {'H', 'o', 's', 't', '\0'};
+    struct bpf_mem_ptr *host;
 
     if (route_config->n_virtual_hosts <= 0 || route_config->n_virtual_hosts > KMESH_PER_VIRT_HOST_NUM) {
         BPF_LOG(WARN, ROUTER_CONFIG, "invalid virt hosts num=%d\n", route_config->n_virtual_hosts);
@@ -119,8 +119,8 @@ virtual_host_match(Route__RouteConfiguration *route_config, address_t *addr, ctx
         return NULL;
     }
 
-    uri = bpf_get_msg_header_element(uri_key);
-    if (!uri) {
+    host = bpf_get_msg_header_element(host_key);
+    if (!host) {
         BPF_LOG(ERR, ROUTER_CONFIG, "failed to get URI in msg\n");
         return NULL;
     }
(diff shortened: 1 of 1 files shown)


bpf/kmesh/include/route_config.h
```diff
@@ -105,7 +105,7 @@ virtual_host_match(Route__RouteConfiguration *route_config, address_t *addr, ctx
     void *ptrs = NULL;
     Route__VirtualHost *virt_host = NULL;
     Route__VirtualHost *virt_host_allow_any = NULL;
-    char uri_key[4] = {'U', 'R', 'I', '\0'};
+    char host_key[5] = {'H', 'o', 's', 't', '\0'};
```
hzxuzhonghu, line 108: This can be `authority` as well

hzxuzhonghu: NVM, authority  is for http2

hzxuzhonghu, line 108: Where is `host_key` header value read

lec-bit: The kmesh.ko file parses and saves the HTTP packet header. The user-defined bpf_get_msg_header_element helper function is added to read the header.

bpf/kmesh/include/route_config.h
```diff
@@ -119,8 +119,8 @@ virtual_host_match(Route__RouteConfiguration *route_config, address_t *addr, ctx
         return NULL;
     }
 
-    uri = bpf_get_msg_header_element(uri_key);
-    if (!uri) {
+    host = bpf_get_msg_header_element(uri_key);
```
nlgwcy, line 122: uri_key -> host_key

hzxuzhonghu (approved):

lec-bit:
image

PR #287, opened 2024-04-30
FIX bpf_map_delete_elem unexpected error log

hzxuzhonghu:
**What type of PR is this?**

We areusing bpf_map_delete_elem  which returns longs type. We see very weired phenomenon though it is -2 when no key exists, but it is not equal to int -2.

Fixes #

```release-note

```

### DIFF
--- a/bpf/kmesh/workload/sockops_tuple.c
+++ b/bpf/kmesh/workload/sockops_tuple.c
@@ -89,17 +89,17 @@ static inline void clean_auth_map(struct bpf_sock_ops *skops)
     // the server info when we transmitted to the kmesh auth info.
     // In this way, auth can be performed normally.
     extract_skops_to_tuple_reverse(skops, &tuple_key);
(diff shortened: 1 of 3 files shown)


bpf/kmesh/workload/sockops_tuple.c
```diff
@@ -199,8 +198,7 @@ int record_tuple(struct bpf_sock_ops *skops)
         auth_ip_tuple(skops);
         break;
     case BPF_SOCK_OPS_STATE_CB:
-        if (skops->args[1] == BPF_TCP_CLOSE || skops->args[1] == BPF_TCP_CLOSE_WAIT
-            || skops->args[1] == BPF_TCP_FIN_WAIT1) {
+        if (skops->args[1] == BPF_TCP_CLOSE) {
```
nlgwcy, line 201: @bitcoffeeiux

PR #295, opened 2024-05-06
Separate bypass from Kmesh mode

### LINKED ISSUE #294: Bypass controller should not be coupled ads or worklaod mode
**Task description**:

Curretly bypass controller only start when kmesh works in ads or workload mode.  https://github.com/kmesh-net/kmesh/blob/ac76f79373da36fec4cbac50cb150c2cbd96cfdd/pkg/controller/controller.go#L48

But as i can see bypass can be used separately. So we may need to allow it running alone

**Solution**:

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

After a contributor has successfully completed 1-2 good first issue's,
they should be ready to move on to `help wanted` items, saving the remaining `good first issue` for other new contributors.

**How to join or take the task**:

Just reply on the issue with the message `/assign` in a separate line.

Then, the issue will be assigned to you.

**How to ask for help**:

If you need help or have questions, please feel free to ask on this issue.
The issue author or other members of the community will guide you through the contribution process.

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Separate bypass from Kmesh mode so that it can run separately
Fixes #294

### DIFF
--- a/daemon/manager/manager.go
+++ b/daemon/manager/manager.go
@@ -29,6 +29,7 @@ import (
 	"kmesh.net/kmesh/pkg/bpf"
 	"kmesh.net/kmesh/pkg/cni"
 	"kmesh.net/kmesh/pkg/controller"
+	"kmesh.net/kmesh/pkg/controller/bypass"
 	"kmesh.net/kmesh/pkg/controller/dump"
 	"kmesh.net/kmesh/pkg/logger"
 	"kmesh.net/kmesh/pkg/options"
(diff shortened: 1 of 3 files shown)


daemon/manager/manager.go
```diff
@@ -76,6 +77,10 @@ func Execute() error {
 	log.Info("controller Start successful")
 	defer controller.Stop()
 
+	if err := bypass.StartByPass(); err != nil {
```
hzxuzhonghu, line 80: I am thinking this maybe better in controller.Start

PR #309, opened 2024-05-09
Fix the Enhanced Compilation Failure Issue

lec-bit:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/bpf/bpf_kmesh.go
+++ b/pkg/bpf/bpf_kmesh.go
@@ -276,7 +276,7 @@ func (sc *BpfKmesh) ApiEnvCfg() error {
 	var err error
 	var info *ebpf.MapInfo
 	var id ebpf.MapID
-	info, err = Obj.Kmesh.SockOps.KmeshSockopsMaps.KmeshListener.Info()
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (approved):
Thanks @lec-bit 

/lgtm

bfforever:
https://github.com/kmesh-net/kmesh/blob/92ec79142f6c7d56eaafce9fa303108c822fe9df/pkg/bpf/bpf_kmesh.go#L83
https://github.com/kmesh-net/kmesh/blob/92ec79142f6c7d56eaafce9fa303108c822fe9df/pkg/bpf/bpf_kmesh.go#L86
Maybe above two place need to change  `Bpfmesh` to reference type to fix enhanced complilation failure.

hzxuzhonghu:
In theory this is better to return pointer type in case there is some copy and we mutate the original object later . @lec-bit @bfforever Did you test the features with this patch, i canot see where we copy it

bfforever:
My development machine is enhanced kernel, `go build ` would build with bpf_kmesh.go file, not bpf_kmesh_l4.go file, so would report a `inconsistent type error` at compile time.

Issue #324, opened 2024-05-11
Accelerate bpf map update in ads mode


hzxuzhonghu:
**What would you like to be added**:

As we have tested the listener, route, cluster, endpoints updates all showed bad performance. It requires much time weiting bpf map. As investigated, most of that is costed from map in map. 

Should do some re design the structure.

PR #326, opened 2024-05-13
optimization: Accelerates the performance of updating the xDS configuration

nlgwcy:
**What type of PR is this?**

/kind enhancement

Fixes https://github.com/kmesh-net/kmesh/issues/324

benchmark test:
before modify:
```sh
[root@localhost v2]# go test -bench=. --benchtime=100x
goos: linux
goarch: amd64
pkg: kmesh.net/kmesh/pkg/cache/v2
cpu: Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz
BenchmarkClusterFlush-16             100         835615271 ns/op
BenchmarkFlush-16                    100        1595920212 ns/op
PASS
ok      kmesh.net/kmesh/pkg/cache/v2    245.370s
[root@localhost v2]#
```
after modify:
```sh
[root@localhost v2]# go test -bench=. --benchtime=100x
goos: linux
goarch: amd64
pkg: kmesh.net/kmesh/pkg/cache/v2
cpu: Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz
BenchmarkClusterFlush-16             100            600145 ns/op
BenchmarkFlush-16                    100            283194 ns/op
PASS
ok      kmesh.net/kmesh/pkg/cache/v2    4.047s
```

```release-note

```


pkg/bpf/bpf.go
```diff
@@ -140,6 +149,7 @@ func (l *BpfLoader) Stop() {
 	var err error
```
LiZhenCheng9527, line 149: There is no need to define an error first

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -69,6 +71,27 @@ struct op_context {
         (context).curr_fd = (fd);                                                                                      \
     } while (0)
 
+#define TASK_SIZE (100)
```
LiZhenCheng9527, line 74: What is the meaning of Task_Size? Is it the size of the map for single step processing? Can you briefly explain it by commenting.

LiZhenCheng9527 (commented):
Can you run a benchmark that gives the results of bpf map update?

pkg/bpf/bpf.go
nlgwcy: ok have updated PR desc

hzxuzhonghu:
Why `BenchmarkFlush`(actually listener) is quicker than cluster flush

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
nlgwcy: yes, update for the map-in-map table is slow(5ms/per-record). When the Kmesh-daemon is started, a group of working threads are created to update records in the map-in-map table. Each thread is responsible for creating `task_size` records. I'll update proposal later.

nlgwcy:
> Why `BenchmarkFlush`(actually listener) is quicker than cluster flush

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -1363,3 +1246,166 @@ void deserial_free_elem(void *value)
...
+        sem_getvalue(sem, &sem_val);
+    } while (sem_val < wait_value);
+}
+
+int outter_map_update(int outter_fd)
+{
+    int i, ret = 0;
+    pthread_t tid;
+    struct task_contex *task_ctx = NULL;
+    int threads = MAX_OUTTER_MAP_ENTRIES / TASK_SIZE + 1;
```
hzxuzhonghu, line 1305: What does it mean

nlgwcy:
> Why `BenchmarkFlush`(actually listener) is quicker than cluster flush

from the prespective of the testcase design:
BenchmarkClusterFlush
Cluster records are continuously added to the map. (the keys of cluster map(cluster name) are different.)
```c
// cluster map define
struct {
    __uint(type, BPF_MAP_TYPE_HASH);
    __uint(key_size, CLUSTER_NAME_MAX_LEN);
    __uint(value_size, sizeof(Cluster__Cluster));
    __uint(map_flags, BPF_F_NO_PREALLOC);
    __uint(max_entries, MAP_SIZE_OF_CLUSTER);
} map_of_cluster SEC(".maps");
```
BenchmarkFlush:
Only one listener record is updated. (The `listener.Name` is randomly set, but the key of the listener map is `addr`.)
```c
// listener map define
struct {
    __uint(type, BPF_MAP_TYPE_HASH);
    __uint(key_size, sizeof(address_t));
    __uint(value_size, sizeof(Listener__Listener));
    __uint(max_entries, MAP_SIZE_OF_LISTENER);
    __uint(map_flags, BPF_F_NO_PREALLOC);
} map_of_listener SEC(".maps");
```

hzxuzhonghu:
Good analyze, So we should update the listener  benchmark

nlgwcy:
> Good analyze, So we should update the listener benchmark

may not be required. This case is used to test the scenario where the same record is updated. In `deserial_update_elem`, the instance tree corresponding to the record will be deleted first.

hzxuzhonghu:
So we need to test different scenaios. Add/update/delete

hzxuzhonghu (approved):
/lgtm

PR #328, opened 2024-05-15
update listener benchmark

LiZhenCheng9527:
**What type of PR is this?**

A random address is set for the listener in the test case
Fixes #

```release-note

```

### DIFF
--- a/pkg/cache/v2/listener_test.go
+++ b/pkg/cache/v2/listener_test.go
@@ -301,7 +301,5 @@ func BenchmarkListenerFlush(b *testing.B) {
 }
 
 func randomIPv4() string {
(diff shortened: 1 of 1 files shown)


pkg/cache/v2/listener_test.go
```diff
@@ -291,9 +292,16 @@ func BenchmarkFlush(b *testing.B) {
...
 
 		cache.Flush()
 		assert.Equal(t, listener.GetApiStatus(), core_v2.ApiStatus_NONE)
 	}
 }
+
+func randomIPv4() string {
+	ip := uint32(rand.Intn(256))<<24 | uint32(rand.Intn(256))<<16 | uint32(rand.Intn(256))<<8 | uint32(rand.Intn(256))
+
+	return fmt.Sprintf("%d.%d.%d.%d", ip&0xff000000>>24, ip&0x00ff0000>>16, ip&0x0000ff00>>8, ip&0x000000ff)
```
hzxuzhonghu, line 306: This can be optimized

PR #336, opened 2024-05-18
fix: should get authz policy from dst-worklod when rbac

supercharge-xsy:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/auth/rbac.go
+++ b/pkg/auth/rbac.go
@@ -157,14 +157,13 @@ func (r *Rbac) RemovePolicy(policyKey string) {
 }
 
 func (r *Rbac) doRbac(conn *rbacConnection) bool {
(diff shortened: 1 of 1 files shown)


supercharge-xsy:
@nlgwcy @hzxuzhonghu

hzxuzhonghu (commented):
I am wondering how it works when we tested

supercharge-xsy:
> I am wondering how it works when we tested

because another issue， we update by_namespace store when apply a workload-select policy, it will leader that  policy makes effect when client and server in a same namspace;
 that issue fix by: https://github.com/kmesh-net/kmesh/pull/333

hzxuzhonghu (approved):

PR #350, opened 2024-05-21
Fix status server panic

### LINKED ISSUE #348: Kmesh panic when `curl 127.0.0.1:15200/bpf/kmesh/maps`
Kmesh panicked when execute debug by request the status server. 

**How to reproduce:**

run kmesh in workload mode, and then `curl 127.0.0.1:15200/bpf/kmesh/maps` in kmesh daemon.

More worse is that when after it panicked, it showed its still healthy

```
root@kurator-linux-0001:~# k get pod -n kmesh-system
NAME           READY   STATUS    RESTARTS   AGE
kmesh2-8nrwk   1/1     Running   0          11d

```

hzxuzhonghu:
**What type of PR is this?**

Fixes #348

```release-note

```


pkg/status/status_server.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+var log = logger.NewLoggerField("status")
+
+const (
+	adminAddr = "localhost:15200"
+
+	patternHelp       = "/help"
+	patternOptions    = "/options"
+	patternBpfAdsMaps = "/debug/bpf/ads"
```
hzxuzhonghu, line 41: @nlgwcy Should we separate ads and workload here?

pkg/status/status_server.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	cache := client.AdsController.Processor.Cache
+	dynamicRes := &adminv2.ConfigResources{}
+
+	dynamicRes.ClusterConfigs = cache.ClusterCache.DumpBpf()
+	dynamicRes.ListenerConfigs = cache.ListenerCache.DumpBpf()
+	dynamicRes.RouteConfigs = cache.RouteCache.StatusLookup()
```
nlgwcy, line 108: cache.RouteCache.StatusLookup() -> cache.RouteCache.DumpBpf

pkg/status/status_server.go
nlgwcy: yes, need to be separated.

pkg/status/status_server.go
hzxuzhonghu: missed it

Issue #355, opened 2024-05-23
Unexpected Bpf tracelog 


hzxuzhonghu:
**What happened**:

Running L7 waypoint example  https://kmesh.net/en/docs/userguide/try_waypoint/.

I have seen some errors in bpf log. not sure why it failed. But I can see the waypoint access log via enabling accesslog.

```
[2024-05-23T02:20:26.624Z] "GET /reviews/0 HTTP/1.1" 200 - via_upstream - "-" 0 358 3 3 "-" "curl/8.7.1" "6a4eac59-a99e-4a3b-af12-f28589732903" "reviews:9080" "10.244.0.100:9080" inbound-vip|9080|http/v1|reviews.default.svc.cluster.local 10.244.0.109:37840 10.96.165.244:9080 envoy://internal_client_address/ - -
[2024-05-23T02:20:27.740Z] "GET /reviews/0 HTTP/1.1" 200 - via_upstream - "-" 0 358 2 2 "-" "curl/8.7.1" "dd57416d-a285-490b-b451-8a70609ba482" "reviews:9080" "10.244.0.100:9080" inbound-vip|9080|http/v1|reviews.default.svc.cluster.local 10.244.0.109:37840 10.96.165.244:9080 envoy://internal_client_address/ - -

```

```

           <...>-3051027 [000] d...1 2317308.111889: bpf_trace_printk: [KMESH] DEBUG: origin addr=[153706506:30755]

           <...>-3051027 [000] d...1 2317308.111897: bpf_trace_printk: [KMESH] DEBUG: bpf find frontend addr=[153706506:30755]

           <...>-3051034 [000] d...1 2317308.115448: bpf_trace_printk: [KMESH] DEBUG: origin addr=[1310744586:30755]

           <...>-3051034 [000] d...1 2317308.115455: bpf_trace_printk: [KMESH] DEBUG: bpf find frontend addr=[1310744586:30755]

          python-3051034 [000] d...1 2317308.119152: bpf_trace_printk: [KMESH] DEBUG: origin addr=[4104478730:30755]

          python-3051034 [000] d...1 2317308.119163: bpf_trace_printk: [KMESH] DEBUG: bpf find frontend addr=[4104478730:30755]

          python-3051034 [000] d...1 2317308.119165: bpf_trace_printk: [KMESH] DEBUG: origin addr=[1856135178:41018]

[...]

PR #359, opened 2024-05-23
backend_manager can return 1 from tail call, so check negative value

hzxuzhonghu:
**What type of PR is this?**

Fix https://github.com/kmesh-net/kmesh/issues/355

```release-note

```


hzxuzhonghu:
Wa are not sure how the `1` was returned, and could not reproduce with latest code

hzxuzhonghu:
didnot see again, this was likely to be caused by invalid mem access

PR #368, opened 2024-05-27
Improve isManagedWorkload

hzxuzhonghu:
This is adding check node name

### DIFF
--- a/pkg/auth/rbac.go
+++ b/pkg/auth/rbac.go
@@ -49,8 +49,9 @@ var (
 )
 
 type Rbac struct {
-	policyStore *policyStore
-	bpfWorkload *bpf.BpfKmeshWorkload
+	policyStore   *policyStore
+	workloadCache cache.WorkloadCache
+	bpfWorkload   *bpf.BpfKmeshWorkload
 }
 
 type Identity struct {
@@ -67,10 +68,11 @@ type rbacConnection struct {
 	dstPort     uint32
 }
 
-func NewRbac(workloadObj *bpf.BpfKmeshWorkload) *Rbac {
+func NewRbac(workloadObj *bpf.BpfKmeshWorkload, workloadCache cache.WorkloadCache) *Rbac {
 	return &Rbac{
-		policyStore: newPolicystore(),
-		bpfWorkload: workloadObj,
+		policyStore:   newPolicystore(),
+		workloadCache: workloadCache,
+		bpfWorkload:   workloadObj,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/auth/rbac.go
```diff
@@ -67,10 +68,11 @@ type rbacConnection struct {
...
 }
 
-func NewRbac(workloadObj *bpf.BpfKmeshWorkload) *Rbac {
+func NewRbac(workloadObj *bpf.BpfKmeshWorkload, workloadCache cache.WorkloadCache) *Rbac {
 	return &Rbac{
-		policyStore: newPolicystore(),
-		bpfWorkload: workloadObj,
+		policyStore:   newPolicystore(),
+		workloadCache: workloadCache,
+		bpfWorkload:   workloadObj,
```
supercharge-xsy, line 75: Why add workload info to the rbac cache？

hzxuzhonghu: because this patch is also removing the global cache var

pkg/controller/workload/cache/workload_cache.go
```diff
@@ -25,35 +25,38 @@ import (
...
-	return &workloadStore{
+func NewWorkloadCache() *cache {
+	return &cache{
 		byUid:  make(map[string]*workloadapi.Workload),
 		byAddr: make(map[NetworkAddress]*workloadapi.Workload),
 	}
 }
 
-func (w *workloadStore) GetWorkloadByUid(uid string) *workloadapi.Workload {
+func (w *cache) GetWorkloadByUid(uid string) *workloadapi.Workload {
```
LiZhenCheng9527, line 53: Wouldn't it be better to modify the `w` as well.

hzxuzhonghu: right, should do

pkg/controller/workload/workload_processor.go
```diff
@@ -467,16 +470,26 @@ func (p *Processor) handleDataWithoutService(workload *workloadapi.Workload) err
...
+		oldIdentity := p.getIdentityByUid(workload.Uid)
+		newIdentity := spiffe.Identity{
+			TrustDomain:    workload.TrustDomain,
+			Namespace:      workload.Namespace,
+			ServiceAccount: workload.ServiceAccount,
+		}.String()
+		if oldIdentity == "" {
+			// This is the case workload added first time
+			p.Sm.SendCertRequest(newIdentity, kmeshsecurity.ADD)
+		} else if oldIdentity != newIdentity {
```
hzxuzhonghu, line 484: maybe no need to check this, in real case workload identity cannot change unless recreate a pod

hzxuzhonghu:
@supercharge-xsy any other comments

PR #390, opened 2024-05-28
Make use of FIFQ instead of priority queue

hzxuzhonghu:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/security/manager.go
+++ b/pkg/controller/security/manager.go
@@ -17,11 +17,11 @@
 package security
 
 import (
-	"container/heap"
 	"sync"
 	"time"
 
 	istiosecurity "istio.io/istio/pkg/security"
+	"k8s.io/client-go/util/workqueue"
 
 	"kmesh.net/kmesh/pkg/constants"
 	"kmesh.net/kmesh/pkg/logger"
@@ -59,14 +59,9 @@ type SecretManager struct {
 	certsCache *certsCache
 
 	// certs rotation priority queue based on exp
-	certsRotateQueue *rotateQueue
+	certsRotateQueue workqueue.Interface
 
 	certRequestChan chan certRequest
-}
-
-type rotateQueue struct {
-	certs []*certExp
-	mu    sync.Mutex
 }
 
 func (s *SecretManager) SendCertRequest(identity string, op int) {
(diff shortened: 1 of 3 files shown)


pkg/controller/security/manager.go
```diff
@@ -268,7 +217,6 @@ func (s *SecretManager) deleteCert(identity string) {
 	log.Debugf("remove identity: %v refCnt : %v", identity, certificate.refCnt)
 	if certificate.refCnt == 0 {
 		delete(s.certsCache.certs, identity)
-		s.certsRotateQueue.delete(identity)
```
lec-bit, line 271: If no delete operation here, does it mean that the certificate will only be automatically deleted when it expires?
So do we no longer need to consider the delayed deletion logic in the case of pod restarts?

pkg/controller/security/manager.go
```diff
@@ -212,31 +156,37 @@ func NewSecretManager() (*SecretManager, error) {
...
 	for {
-		if s.certsRotateQueue.Len() != 0 {
-			top := s.certsRotateQueue.pop()
-			time.Sleep(time.Until(top.exp.Add(-1 * time.Hour)))
-			s.SendCertRequest(top.identity, Rotate)
-		} else {
-			time.Sleep(5 * time.Second)
+		element, quit := s.certsRotateQueue.Get()
+		if quit {
+			return
```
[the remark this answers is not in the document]
LiZhenCheng9527: If don't sleep for 5 seconds, what are the consequences?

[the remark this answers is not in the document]
lec-bit: When the queue is empty, it will loop until the queue is filled with content. Without sleeping, it will consume a significant amount of resources.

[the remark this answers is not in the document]
hzxuzhonghu: Nope, the `get` is stuck until an element added. It uses condition variable instead of polling

pkg/controller/security/manager.go
hzxuzhonghu: It is intended to let it exist, we do check whether we  should sign for a identity when we do sign.

I would double check the case at a time the identity refcnt is 0. And at a future time it is added. 

I will report back

hzxuzhonghu: not a problem now

pkg/controller/security/manager.go
```diff
@@ -283,5 +220,10 @@ func (s *SecretManager) rotateCert(identity string) {
 	}
 	s.certsCache.mu.Unlock()
 
+	if time.Until(certificate.cert.ExpireTime) >= 1*time.Hour {
+		// This can happen when delete a certificate following adding the same one later.
+		log.Debugf("cert %s expire at %T, skip rotate now", identity, certificate.cert.ExpireTime)
+	}
```
lec-bit, line 226: How is `workqueue.DelayingInterface` used? Can it accept duplicate elements? I don't fully understand the case how this scenario occured.

hzxuzhonghu: WorkQueue doesnot store duplicate elements, but as we are using delay push, there can be duplicate elmenets delay pushes. And as long as one is consumed, the other can be consumed later.

lec-bit: WorkQueue is indeed very convenient. 
I have no other questions.

hzxuzhonghu: What is the progress of your unit test, could this one be merged first?

Issue #388, opened 2024-05-28
Kmesh eBPF log user space dumping


nlgwcy:
**What would you like to be added**:
Kmesh logs governance process information via `BPF_LOG`. Logs are expected to be dumped to the user space:
1. Log information should be dumped to user space via ringbuf. 
2. In user space, the log information should be obtained through `epoll` and formatted for printing.

reference: https://github.com/anakryiko/bpf-ringbuf-examples/blob/main/src/ringbuf-output.bpf.c

**Why is this needed**:
The current API has the following issues:
1. Logs are recorded in memory through trace_pipe and can be viewed using `bpftool prog tracelog`, limited space.
3. Format parameters such as "%pI4h" are [supported latter than v5.13 kernel version](https://github.com/torvalds/linux/commit/d9c9e4db186ab),  not yet supported in 5.10 (such as openEuler 22.03 SP3), 
4. Only three parameters are supported.

hzxuzhonghu:
>Logs are recorded in memory through trace_pipe and can be viewed using bpftool prog tracelog, limited space.

Not an expert on this, does it occupy kernel memory?

bfforever:
Maybe we need to firstly define some simple fixed format，because ebpf prog seems not support var args. 
For example:   
- no args `BPF_LOG(ERR, SOCKOPS, "set sockops cb failed!\n")`
- 1 int args or 2 int args `BPF_LOG(ERR, KMESH, "set bypass value failed!, err is %d\n", err)`  and `BPF_LOG(DEBUG, KMESH, "bpf find frontend addr=[%pI4h:%u]\n", &ip, bpf_ntohs(ctx->user_port));`, or more int args.
- Also can add a flag to represent which kind of log is occured. Just as this implement https://github.com/cilium/cilium/blob/d18d41e6995679624980671c46b25e5a80d40136/bpf/lib/dbg.h#L8

I am not sure whether the format should mix with string args `%s`, the situation may become complex. Dose anything need to supplement?

nlgwcy:
> > Logs are recorded in memory through trace_pipe and can be viewed using bpftool prog tracelog, limited space.
> 
> Not an expert on this, does it occupy kernel memory?

This is a debugfs file system, and it will be cleared after the system reboot.

nlgwcy:
> Maybe we need to firstly define some simple fixed format，because ebpf prog seems not support var args. For example:
> 
[...]

PR #402, opened 2024-05-30
Kmesh eBPF log user space dumping

bfforever:
**What type of PR is this?**
/kind feature
/kind enhancement

1. Kmesh eBPF log user space dumping.
version > 5.13 dump bpf log to userspace, otherwise print log in trace pipe.

2. Support dynamic bpf log level.
 `k exec ds/kmesh -n kmesh-system -- curl -s localhost:15200/debug/bpfLogLevel/3`

Fixes [388](https://github.com/kmesh-net/kmesh/issues/388)

bpf_log_3

When print bpf log to kmesh daemon process or user space,  just use BPF_LOG_U instead of BPF_LOG, the args which input is same as BPF_LOG.
```release-note

```

### DIFF
--- a/bpf/include/bpf_log.h
+++ b/bpf/include/bpf_log.h
@@ -39,23 +39,32 @@ struct {
     __uint(max_entries, 1);
     __type(key, int);
     __type(value, struct log_event);
-} heap SEC(".maps");
+} tmp_log_buf SEC(".maps");
 
-#define BPF_LOG(l, t, f, ...)                                                                                          \
-    do {                                                                                                               \
-        int loglevel = BPF_MIN((int)BPF_LOG_LEVEL, ((int)BPF_LOG_DEBUG + (int)(BPF_LOGTYPE_##t)));                     \
-        if ((int)(BPF_LOG_##l) <= loglevel) {                                                                          \
-            char fmt[] = "[" #t "] " #l ": " f "";                                                                     \
-            bpf_trace_printk(fmt, sizeof(fmt), ##__VA_ARGS__);                                                         \
-        }                                                                                                              \
-    } while (0)
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
+    __uint(max_entries, 1);
+    __uint(key_size, sizeof(__u32));
+    __uint(value_size, sizeof(__u32));
+} bpf_log_level SEC(".maps");
 
 /* Add this macro to get ip addr from ctx variable, include bpf_sock_addr or bpf_sock_ops, weird
-reason is direct access would not be print ipaddr when pass `&ctx->remote_ipv4` to bpf_trace_printk */
+reason is that would not be print ipaddr, when directly pass `&ctx->remote_ipv4` to bpf_trace_printk, maybe ctx pass in
+to printk would be changed*/
 #define DECLARE_VAR_IPV4(ctx_ip, name)                                                                                 \
     __u32 name = 0;                                                                                                    \
     name = ctx_ip;
 
+#define DECLARE_VAR_IPV6(ctx_ip, name)                                                                                 \
+    __u32 name[4] = {0};                                                                                               \
+    name[0] = ctx_ip[0];                                                                                               \
+    name[1] = ctx_ip[1];                                                                                               \
+    name[2] = ctx_ip[2];                                                                                               \
+    name[3] = ctx_ip[3];
+
+/* Add KERNEL_VERSION_HIGHER_5_13_0 to resolve problem, which kernel version lower than 5.13, or linux distribution
+lower than 22.09, compile would report an error of bpf_snprintf dont exist */
+#if KERNEL_VERSION_HIGHER_5_13_0
 #define Kmesh_BPF_SNPRINTF(out, out_size, fmt, args...)                                                                \
     ({                                                                                                                 \
         unsigned long long ___param[___bpf_narg(args)];                                                                \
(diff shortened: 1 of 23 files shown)


daemon/options/bpf.go
```diff
@@ -32,13 +32,15 @@ type BpfConfig struct {
...
 	BpfVerifyLogSize int
+	EnableBpfLog     bool
 }
 
 func (c *BpfConfig) AttachFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&c.BpfFsPath, "bpf-fs-path", "/sys/fs/bpf", "bpf fs path")
 	cmd.PersistentFlags().StringVar(&c.Cgroup2Path, "cgroup2-path", "/mnt/kmesh_cgroup2", "cgroup2 path")
 	cmd.PersistentFlags().StringVar(&c.Mode, "mode", "workload", "controller plane mode, valid values are [ads, workload]")
 	cmd.PersistentFlags().BoolVar(&c.EnableMda, "enable-mda", false, "enable mda")
+	cmd.PersistentFlags().BoolVar(&c.EnableBpfLog, "enable-bpf-log", false, "enable ebpf log in daemon process")
```
LiZhenCheng9527, line 43: Is the bpf log function turned off by default?

pkg/controller/controller.go
```diff
@@ -71,6 +76,12 @@ func (c *Controller) Start() error {
 		return fmt.Errorf("secretManager create failed: %v", err)
 	}
 
+	if c.enableBpfLog {
+		if err = logger.StartRingBufReader(ctx, c.mode, c.bpfFsPath); err != nil {
+			return err
```
LiZhenCheng9527, line 81:
```suggestion
			return fmt.Errorf("fail to star bufreader: %v", err)
```

daemon/options/bpf.go
bfforever: I am also not sure whether need to enable ebpf log in userspace , I wrote this code just as yesterday said, it is like a feature gateway.

bfforever: > I am also not sure whether need to enable ebpf log default  in userspace , I wrote this code just as yesterday said, it is like a feature gateway.

bpf/include/bpf_log.h
```diff
@@ -32,13 +32,31 @@
...
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 256 * 1024 /* 256 KB */);
+} kmesh_events SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(max_entries, 1);
+    __type(key, int);
+    __type(value, struct log_event);
+} heap SEC(".maps");
```
supercharge-xsy, line 58: What's this map for? It's better to name it based on usage scenario

daemon/options/bpf.go
bfforever: > Good job~ The following two points need to be confirmed to determine whether tunned off by default:
> 
> 1. Can the 5.10 kernel version run properly?
> 2. Test the interface performance.

for 2, possiblely it won't turn on in product env, I think it only need to run this feature when user want to debug or development Kmesh. BTW, how to test this kind of interface?

bpf/include/bpf_log.h
bfforever: > What's this map for? It's better to name it based on usage scenario

This map is to avoid exceeding ebpf stack which is 512B, so use this map to hold the log_event object.

daemon/options/bpf.go
nlgwcy: Log needs to be opened in the production environment, otherwise there is no way to locate the problem.
We can write a test example, generate a batch of logs and count the time spent.

nlgwcy: In the production environment, we can use the log level to control the printing of only critical error messages to avoid massive log output.

bfforever: For 1, I had searched 5.10 kernel code, the `bpf_snprintf` or `BPF_SNPRINTF` not exist.

hzxuzhonghu (commented):
can we use `bpf_snprintf_btf`, which is supported since 5.10

bfforever:
> can we use `bpf_snprintf_btf`, which is supported since 5.10

Possiblely can't, I see this function use to render type info with type value of a certain struct, to output a string, not provide format  capability.

bpf/include/bpf_log.h
```diff
@@ -54,4 +83,70 @@ reason is direct access would not be print ipaddr when pass `&ctx->remote_ipv4`
     __u32 name = 0;                                                                                                    \
     name = ctx_ip;
 
+/* Add KERNEL_VERSION_HIGHER_5_13_0 to resolve problem, when kernel version lower than 5.13, or linux distribution lower
+than 22.09, compile would report an error which bpf_snprintf dont exist */
+#if KERNEL_VERSION_HIGHER_5_13_0
+#define Kmesh_BPF_SNPRINTF(out, out_size, fmt, args...)                                                                \
```
nlgwcy, line 89: The current implementation may be complex.
IMO, we could implement `ip2str` function and  print ip address based on strings. for example:
```c
char * ip2str(u32 *ip, bool is_v4) {
// fmt ip to string
}

BPF_LOG("ip:%s, port:%d\n", ip2str(&ip, true), port);
```

bfforever: Using ip2str maybe cause whatever log level will execute ip convert logic. Do we
need to control it by log level?

pkg/status/status_server.go
```diff
@@ -192,6 +200,27 @@ func (s *Server) readyProbe(w http.ResponseWriter, r *http.Request) {
 	_, _ = w.Write([]byte("OK"))
 }
 
+func (s *Server) bpfLogLevel(w http.ResponseWriter, r *http.Request) {
+	pattern := regexp.MustCompile(`^/debug/bpfLogLevel/(\d+)$`)
+	matches := pattern.FindStringSubmatch(r.URL.Path)
+	if len(matches) > 1 {
+		level, err := strconv.Atoi(matches[1])
+		if err != nil || (level < 0 || level > 3) {
```
nlgwcy, line 208:
```suggestion
		if err != nil || (level < BPF_LOG_ERR || level > BPF_LOG_DEBUG) {
```

bpf/include/common.h
```diff
@@ -87,4 +87,150 @@ static inline int kmesh_map_update_elem(void *map, const void *key, const void *
 
 #define GET_SKOPS_LOCAL_PORT(sk_ops) (__u16)((sk_ops)->local_port)
 
-#endif  // _COMMON_H_
+#define MAX_BUF_LEN 100
+#define MAX_IP4_LEN 16
+#define MAX_IP6_LEN 40
+
+#ifndef u8
+#define u8 __u8
```
nlgwcy, line 95: Why not directly include the corresponding header file?

bfforever: It seems couldnot find u8, u16 type def in ebpf enviroment except include vmlinux.h. So back to use __u8, __u16 instead of u8 u16.

nlgwcy:
Good Job~ The following two points can be optimized in the future:
1. byte order conversion in `ip2str` can be extracted.
2. execute `ip2str` according to the log level. (e.g: IP2STR_ON_ERR/IP2STR_ON_DBG/...)

bfforever:
> Good Job~ The following two points can be optimized in the future:
> 
> 1. byte order conversion in `ip2str` can be extracted.
> 2. execute `ip2str` according to the log level. (e.g: IP2STR_ON_ERR/IP2STR_ON_DBG/...)

Okay.

hzxuzhonghu:
can you add a document on this feature to website https://kmesh.net/en/docs/

bfforever:
> can you add a document on this feature to website https://kmesh.net/en/docs/

Of course.

PR #413, opened 2024-06-04
Support service dump

hzxuzhonghu:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/api/v2/workloadapi/workload_helper.go
+++ b/api/v2/workloadapi/workload_helper.go
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2024 The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


api/v2/workloadapi/workload_helper.go
```diff
@@ -0,0 +1,27 @@
...
+
+package workloadapi
+
+// ResourceName returns the unique key of Workload.
+func (x *Workload) ResourceName() string {
+	return x.Uid
+}
+
+// ResourceName returns the unique key of Service.
+func (x *Service) ResourceName() string {
```
LiZhenCheng9527, line 25: Would changing the name to GetResourceName be better?

hzxuzhonghu: nit：I donot think so, it is actually generating resourceName.  `ResourceName` looks much better/simpler to me.

PR #414, opened 2024-06-04
support kmesh manage in pod level

weli-l:
**What type of PR is this?**

support Kmesh manage in pod level
Fixes #

```release-note

```


hzxuzhonghu (changes requested):
Please add test coverage

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,235 @@
...
+	"kmesh.net/kmesh/pkg/logger"
+)
+
+var (
+	log                = logger.NewLoggerField("kmesh_manage")
+	annotationDelPatch = []byte(fmt.Sprintf(
+		`{"metadata":{"annotations":{"%s":null}}}`,
+		KmeshAnnotation,
+	))
+	annotationAddPatch = []byte(fmt.Sprintf(
```
LiZhenCheng9527, line 46: If this variable is a fixed value, you can use the constant

pkg/controller/netns/netns.go
```diff
@@ -0,0 +1,139 @@
...
+
+	nd "istio.io/istio/cni/pkg/nodeagent"
+	"istio.io/istio/pkg/util/sets"
+	"istio.io/pkg/log"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var (
+	FS embed.FS
```
LiZhenCheng9527, line 37: Does this variable have to be a global variable?

pkg/controller/kmesh/kmesh_manage.go
hzxuzhonghu, line 1: move to controller if this is not shared with other components

pkg/controller/controller.go
```diff
@@ -75,7 +76,17 @@ func (c *Controller) Start() error {
 	if c.client.workloadController != nil {
 		c.client.workloadController.Processor.Sm = secertManager
 	}
+	clientset, err := utils.GetK8sclient()
```
kwb0523, line 79: Can the clientset be reused with bypass? and does the kmesh manager logic need to be moved before the bypass logic.

pkg/controller/controller.go
```diff
@@ -22,6 +22,7 @@ import (
 	"kmesh.net/kmesh/pkg/bpf"
 	"kmesh.net/kmesh/pkg/constants"
 	"kmesh.net/kmesh/pkg/controller/bypass"
+	kmeshmanage "kmesh.net/kmesh/pkg/controller/kmesh"
```
YaoZengzeng, line 25: It's weired to name a sub package as `kmesh`

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+			namespace, err := client.CoreV1().Namespaces().Get(context.TODO(), pod.GetNamespace(), metav1.GetOptions{})
+			if err != nil {
+				log.Errorf("failed to get namespace %s, err: %v", pod.GetNamespace(), err)
+				return
+			}
+
+			if _, ok := namespace.Labels[LabelSelectorKmesh]; ok {
+				log.Infof("namespace %s has label 'istio.io/dataplane-mode=Kmesh', skipping further processing", pod.GetNamespace())
+				return
```
YaoZengzeng, line 89: Use constant defined above to replace 'istio.io/dataplane-mode=Kmesh'

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+
+			log.Infof("%s/%s: kmesh manage pod:", pod.GetNamespace(), pod.GetName())
+
+			nspath, _ := ns.GetNSpath(pod)
+
+			if err := handleKmeshManage(nspath, true); err != nil {
+				log.Errorf("failed to enable kmesh manage")
```
YaoZengzeng, line 97: Use `Kmesh` instead of `kmesh`

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+
+			namespace, err := client.CoreV1().Namespaces().Get(context.TODO(), pod.GetNamespace(), metav1.GetOptions{})
+			if err != nil {
+				log.Errorf("failed to get namespace %s, err: %v", pod.GetNamespace(), err)
+				return
+			}
+
+			if _, ok := namespace.Labels[LabelSelectorKmesh]; ok {
+				log.Infof("namespace %s has label 'istio.io/dataplane-mode=Kmesh', skipping further processing", pod.GetNamespace())
+				return
```
YaoZengzeng, line 123: ditto

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+			log.Infof("%s/%s: disable kmesh manage", pod.GetNamespace(), pod.GetName())
+
+			nspath, _ := ns.GetNSpath(pod)
+			if err := handleKmeshManage(nspath, false); err != nil {
+				log.Errorf("failed to disable kmesh manage")
+				return
+			}
+
+			if err := patchKmeshAnnotation(client, pod, false); err != nil {
+				log.Errorf("failed to add kmesh annotation, err is %v", err)
```
YaoZengzeng, line 140: ditto

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+				return
+			}
+		},
+	}); err != nil {
+		return fmt.Errorf("error adding event handler to podInformer: %v", err)
```
YaoZengzeng, line 145:
```suggestion
		return fmt.Errorf("failed to add event handler to podInformer: %v", err)
```

pkg/controller/kmesh/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+	go podInformer.Run(stopChan)
+
+	return nil
+}
+
+func isPodBeingDeleted(pod *corev1.Pod) bool {
+	return pod.ObjectMeta.DeletionTimestamp != nil
+}
+
+func patchKmeshAnnotation(client kubernetes.Interface, pod *corev1.Pod, oper bool) error {
```
YaoZengzeng, line 157:
```suggestion
func patchKmeshAnnotation(client kubernetes.Interface, pod *corev1.Pod, op bool) error {
```

`oper` is weired

pkg/controller/netns/netns.go
```diff
@@ -0,0 +1,139 @@
...
+		return false
+	}
+
+	if strings.IndexFunc(entry.Name(), isNotNumber) != -1 {
+		return false
+	}
+	return true
+}
+
+// copied from istio/cni/pkg/nodeagent/podcgroupns.go
```
YaoZengzeng, line 94: It's better to add url

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+	podInformer := informerFactory.Core().V1().Pods().Informer()
+
+	if _, err := podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			pod, ok := obj.(*corev1.Pod)
+			if !ok {
+				log.Errorf("expected *corev1.Pod but got %T", obj)
+				return
+			}
+			namespace, err := client.CoreV1().Namespaces().Get(context.TODO(), pod.GetNamespace(), metav1.GetOptions{})
```
supercharge-xsy, line 81: This logic of checking if ns has a label can be abstracted into a common function； used by AddFunc and  DeleteFunc

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+		 * connection triggers the cgroup/connect4 ebpf
+		 * program and records the netns cookie information
+		 * of the current connection. The cookie can be used
+		 * to determine whether the pod is been bypass.
+		 * 0.0.0.1:<port> is "cipher key" for cgroup/connect4
+		 * ebpf program. 929/930 is the specific port handled by
+		 * daemon to enable/disable kmesh manage
+		 */
+		simip := net.ParseIP("0.0.0.1")
+		port := 929
```
supercharge-xsy, line 187: magic numbers

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		err = syscall.Connect(sockfd, &syscall.SockaddrInet4{
+			Port: port,
+			Addr: [4]byte(simip.To4()),
+		})
+		if err == nil {
+			return err
+		}
+		errno, ok := err.(syscall.Errno)
+		if ok && errno == 115 { // -EINPROGRESS, Operation now in progress
```
supercharge-xsy, line 206: magic numbers

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+	))
+)
+
+const (
+	DefaultInformerSyncPeriod = 30 * time.Second
+	LabelSelectorKmesh        = "istio.io/dataplane-mode=Kmesh"
+	KmeshAnnotation           = "kmesh.net/redirection"
+)
+
+func StartKmeshManageController(client kubernetes.Interface) error {
```
hzxuzhonghu, line 59:
```suggestion
func NewKmeshManageController(client kubernetes.Interface) error {
```

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,217 @@
...
+	stopChan := make(chan struct{})
+	nodeName := os.Getenv("NODE_NAME")
+
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(client, DefaultInformerSyncPeriod,
```
hzxuzhonghu, line 63: nit: we can share one factory with bypass

weli-l: > nit: we can share one factory with bypass

I think it is not necessary to share the factory, because labels selector in factory is different

pkg/controller/manage/kmesh_manage.go
weli-l: > 

done

pkg/controller/manage/kmesh_manage.go
weli-l: > magic numbers

I have defined the constant

pkg/controller/manage/kmesh_manage.go
weli-l: > magic numbers

I have defined constant

pkg/controller/manage/kmesh_manage.go
weli-l: > This logic of checking if ns has a label can be abstracted into a common function； used by AddFunc and DeleteFunc

I have deleted this logic, because it may affect the data in bpf map kmesh_manage when kmesh restart

pkg/controller/netns/netns.go
weli-l: > It's better to add url

done

pkg/controller/kmesh/kmesh_manage.go
weli-l: > `oper` is weired

I have modified

pkg/controller/kmesh/kmesh_manage.go
weli-l: > 

done

pkg/controller/kmesh/kmesh_manage.go
weli-l: > ditto

done

pkg/controller/kmesh/kmesh_manage.go
weli-l: > Use `Kmesh` instead of `kmesh`

Yes, I have modified

pkg/controller/kmesh/kmesh_manage.go
weli-l: > Use constant defined above to replace 'istio.io/dataplane-mode=Kmesh'

defined

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,199 @@
...
+		"enabled",
+	))
+)
+
+const (
+	DefaultInformerSyncPeriod = 30 * time.Second
+	SpecialIpForKmesh         = "0.0.0.1"
+	EnableKmeshPort           = 929
+	DisableKmeshPort          = 930
+	LabelSelectorKmesh        = "istio.io/dataplane-mode=Kmesh"
```
hzxuzhonghu, line 58:
```suggestion
	KmeshManagedLabelSelector       = constants.DataPlaneModeLabel + DataPlaneModeKmesh
```

hzxuzhonghu, line 58: BTW, i think we not only support `Kmesh`, but also `kmesh`. We are case insensitive, so i donot think this label selector suits here

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,199 @@
...
+			options.FieldSelector = fmt.Sprintf("spec.nodeName=%s", nodeName)
+			options.LabelSelector = LabelSelectorKmesh
+		}))
+
+	informerFactory.Start(wait.NeverStop)
+	informerFactory.WaitForCacheSync(wait.NeverStop)
+
+	podInformer := informerFactory.Core().V1().Pods().Informer()
+
+	if _, err := podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
hzxuzhonghu, line 77: This is not right, we should add the handler before starting informer

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,199 @@
...
+	nodeName := os.Getenv("NODE_NAME")
+
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(client, DefaultInformerSyncPeriod,
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.FieldSelector = fmt.Sprintf("spec.nodeName=%s", nodeName)
+			options.LabelSelector = LabelSelectorKmesh
+		}))
+
+	informerFactory.Start(wait.NeverStop)
+	informerFactory.WaitForCacheSync(wait.NeverStop)
```
hzxuzhonghu, line 73: This is blocking, so let;s move this to a Run method, so we can run all the controller's Run async without waiting each other.

hzxuzhonghu (changes requested):
Please add some tests

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,202 @@
...
+			return
+		}
+	}()
+}
+
+func isPodBeingDeleted(pod *corev1.Pod) bool {
+	return pod.ObjectMeta.DeletionTimestamp != nil
+}
+
+func patchKmeshAnnotation(client kubernetes.Interface, pod *corev1.Pod, op bool) error {
```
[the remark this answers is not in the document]
weli-l: > Nit: donot use `op`, i would rather split into two functions

splited

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,209 @@
...
+	annotationAddPatch = []byte(fmt.Sprintf(
+		`{"metadata":{"annotations":{"%s":"%s"}}}`,
+		KmeshRedirectionAnnotation,
+		"enabled",
+	))
+)
+
+const (
+	DefaultInformerSyncPeriod  = 30 * time.Second
+	SpecialIpForKmesh          = "0.0.0.1"
```
hzxuzhonghu, line 55: Please check with that from CNI, we use 0.0.0.2 now for ipv4 and for v6  `::2`

We can reuse `TriggerControlCommand` function

pkg/controller/manage/kmesh_manage_test.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+	}
+
+	wg.Add(1)
+	_, err = client.CoreV1().Pods("default").Create(context.TODO(), pod, metav1.CreateOptions{})
+	assert.NoError(t, err)
+
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
```
hzxuzhonghu, line 97: I think you should not wait in a go routine

pkg/controller/manage/kmesh_manage_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	select {
+	case <-done:
+		mu.Lock()
+		assert.True(t, enabled, "expected handleKmeshManage to be called for enabling Kmesh manage")
+		assert.False(t, disabled, "expected handleKmeshManage not to be called for disabling Kmesh manage")
+		mu.Unlock()
+	case <-time.After(1 * time.Second):
+		t.Fatalf("timed out waiting for handleKmeshManage to be called")
+	}
```
hzxuzhonghu, line 109: wrap this with a checker function, so you canreuse it below

pkg/controller/manage/kmesh_manage_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	podLister := controller.informerFactory.Core().V1().Pods().Lister()
+	pods, err := podLister.Pods(pod.Namespace).List(labels.Everything())
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(pods), "expected One Pod in the lister after addition")
```
hzxuzhonghu, line 114: This can be not tested in out case, this is the implement detail of k8s

pkg/controller/manage/kmesh_manage_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Reset variables
+	mu.Lock()
+	enabled = false
+	disabled = false
+	mu.Unlock()
```
hzxuzhonghu, line 120: Not sure i understand why you need to lock them. Any race?

pkg/controller/manage/kmesh_manage_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	wg.Add(1)
+	err = client.CoreV1().Pods("default").Delete(context.TODO(), "test-pod", metav1.DeleteOptions{})
```
hzxuzhonghu, line 123: can you add a case of removing the label first

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,209 @@
...
+	EnableKmeshPort            = 929
+	DisableKmeshPort           = 930
+	LabelSelectorKmesh         = constants.DataPlaneModeLabel + "=" + constants.DataPlaneModeKmesh
```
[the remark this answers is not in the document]
weli-l: > As i said, we must support both `kmesh` and `Kmesh`, should be case insensitive to be consistent with previous behavior

supported

pkg/controller/manage/kmesh_manage.go
weli-l: > Please check with that from CNI, we use 0.0.0.2 now for ipv4 and for v6 `::2`
> 
> We can reuse `TriggerControlCommand` function

of course

pkg/controller/manage/kmesh_manage_test.go
weli-l: > I think you should not wait in a go routine

I have delete relevant logic

pkg/controller/manage/kmesh_manage_test.go
weli-l: > wrap this with a checker function, so you canreuse it below

done

pkg/controller/manage/kmesh_manage_test.go
weli-l: > Not sure i understand why you need to lock them. Any race?

There is no race ,deleted

pkg/controller/manage/kmesh_manage_test.go
weli-l: > can you add a case of removing the label first

yes, added

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,194 @@
...
+	"k8s.io/client-go/tools/cache"
+
+	"kmesh.net/kmesh/pkg/constants"
+	ns "kmesh.net/kmesh/pkg/controller/netns"
+	"kmesh.net/kmesh/pkg/logger"
+	"kmesh.net/kmesh/pkg/nets"
+)
+
+var (
+	log                = logger.NewLoggerField("kmesh_manage")
```
hzxuzhonghu, line 41:
```suggestion
	log                = logger.NewLoggerField("manage_controller")
```

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,194 @@
...
+			return
+		}
+	}()
+}
+
+func isPodBeingDeleted(pod *corev1.Pod) bool {
+	return pod.ObjectMeta.DeletionTimestamp != nil
+}
+
+func checkValid(pod *corev1.Pod) bool {
```
hzxuzhonghu, line 153:
```suggestion
func shouldEnroll(pod *corev1.Pod) bool {
```

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,194 @@
...
+
+			log.Infof("%s/%s: enable Kmesh manage", pod.GetNamespace(), pod.GetName())
+
+			nspath, _ := ns.GetPodNSpath(pod)
+
+			if err := handleKmeshManage(nspath, true); err != nil {
+				log.Errorf("failed to enable Kmesh manage")
+				return
+			}
+			if err := addKmeshAnnotation(client, pod); err != nil {
```
hzxuzhonghu, line 89: should we only patch if the annotation not exist?

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,194 @@
...
+				log.Errorf("failed to add Kmesh annotation, err is %v", err)
```
hzxuzhonghu, line 90: For this failure, it could not recover, we can add a workqueue for better resilience

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldPod, okOld := oldObj.(*corev1.Pod)
+			newPod, okNew := newObj.(*corev1.Pod)
+			if !okOld || !okNew {
+				log.Errorf("expected *corev1.Pod but got %T and %T", oldObj, newObj)
+				return
+			}
+
+			if checkValid(oldPod) && !checkValid(newPod) {
```
hzxuzhonghu, line 102: This is only handing label deleted, and there is another case, new label added. Does your proposal include this case?

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,194 @@
...
+				if isPodBeingDeleted(newPod) {
```
hzxuzhonghu, line 103: move before L102

hzxuzhonghu (commented):
Some nits, otherwise lgtm

pkg/controller/manage/kmesh_manage.go
weli-l: > should we only patch if the annotation not exist?

I have added relevant logic in addKmeshAnnotation and delKmeshAnnotation

pkg/controller/manage/kmesh_manage.go
weli-l: > For this failure, it could not recover, we can add a workqueue for better resilience

yes, added

pkg/controller/manage/kmesh_manage.go
weli-l: > This is only handing label deleted, and there is another case, new label added. Does your proposal include this case?

Yes, AddFunc would be triggered in the previous design, but now the informer will obtain all pods on the current node. The logic of whether to manage now is judged in the callback function. The logic you mentioned has been added.

pkg/controller/manage/kmesh_manage.go
weli-l: > move before L102

done

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (c *KmeshManageController) Run() {
+	go func() {
+		c.informerFactory.Start(c.stopChan)
+		if !cache.WaitForCacheSync(c.stopChan, c.podInformer.HasSynced) {
+			log.Error("Timed out waiting for caches to sync")
+			return
+		}
+		c.processItems()
```
hzxuzhonghu, line 169: we need a loop here

pkg/controller/manage/kmesh_manage.go
```diff
@@ -0,0 +1,275 @@
...
+func (c *KmeshManageController) handleAddAnnotation(pod *corev1.Pod) error {
+	return retry.OnError(retry.DefaultRetry, func(err error) bool {
+		return true
+	}, func() error {
+		return addKmeshAnnotation(c.client, pod)
+	})
+}
+
+func (c *KmeshManageController) handleDelAnnotation(pod *corev1.Pod) error {
+	return retry.OnError(retry.DefaultRetry, func(err error) bool {
```
hzxuzhonghu, line 216: donot need to retry here as you have done it in the above caller

pkg/controller/manage/kmesh_manage_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	enabled = false
+	disabled = false
+
+	delete(pod.Labels, constants.DataPlaneModeLabel)
+	wg.Add(1)
+	_, err = client.CoreV1().Pods("default").Update(context.TODO(), pod, metav1.UpdateOptions{})
+	assert.NoError(t, err)
+
+	done = make(chan struct{})
```
hzxuzhonghu, line 120: remove the channel

pkg/controller/manage/kmesh_manage.go
weli-l: > we need a loop here

done

pkg/controller/manage/kmesh_manage.go
weli-l: > donot need to retry here as you have done it in the above caller

yes, this logic is redundant

pkg/controller/manage/kmesh_manage_test.go
weli-l: > remove the channel

removed，channel seems redundant

hzxuzhonghu (approved):
/lgtm

PR #419, opened 2024-06-06
Added dataplane mode constants

hzxuzhonghu:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -20,6 +20,11 @@ const (
 	AdsMode      = "ads"
 	WorkloadMode = "workload"
 
+	// DataPlaneModeLabel is the label used to indicate the data plane mode
+	DataPlaneModeLabel = "istio.io/dataplane-mode"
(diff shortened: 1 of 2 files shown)


pkg/constants/constants.go
```diff
@@ -20,6 +20,11 @@ const (
 	AdsMode      = "ads"
 	WorkloadMode = "workload"
 
+	// DataPlaneModeLabel is the label used to indicate the data plane mode
+	DataPlaneModeLabel = "istio.io/dataplane-mode"
+	// DataPlaneModeKmesh is the value of the label to indicate the data plane mode is kmesh
+	DataPlaneModeKmesh = "kmesh"
```
supercharge-xsy, line 26: It seems that the letter case is not same with before；Kmesh->kmesh

hzxuzhonghu: good catch, we did treat both `Kmesh` and `kmesh` valid with case-insensitive comparing

PR #424, opened 2024-06-07
drop package when auth processing

supercharge-xsy:
**What type of PR is this?**
/kind enhancement

Currently, authentication is performed in user mode. Some packets may be missed during authentication. This pr attempts to drop the package until authentication is complete .  will test the behavior of the client.
Fixes #

```release-note

```

### DIFF
--- a/bpf/kmesh/workload/sockops.c
+++ b/bpf/kmesh/workload/sockops.c
@@ -125,6 +125,16 @@ static inline void auth_ip_tuple(struct bpf_sock_ops *skops)
     bpf_ringbuf_submit(msg, 0);
 }
 
+// insert an init state to auth_map, indicates that auth is being performed 
+static inline void auth_init(struct bpf_sock_ops *skops)
+{
+    struct bpf_sock_tuple tuple_info = {0};
+    extract_skops_to_tuple(skops, &tuple_info);
+    int err = bpf_map_update_elem(&map_of_auth, &tuple_info, 2, BPF_ANY);
+    if (err)
+        BPF_LOG(ERR, SOCKOPS, "insert auth init record failed!, err is %d", err);
+}
+
 // update sockmap to trigger sk_msg prog to encode metadata before sending to waypoint
 static inline void enable_encoding_metadata(struct bpf_sock_ops *skops)
 {
@@ -268,6 +278,7 @@ int sockops_prog(struct bpf_sock_ops *skops)
         if (bpf_sock_ops_cb_flags_set(skops, BPF_SOCK_OPS_STATE_CB_FLAG) != 0)
             BPF_LOG(ERR, SOCKOPS, "set sockops cb failed!\n");
         auth_ip_tuple(skops);
+        auth_init();
         break;
     case BPF_SOCK_OPS_STATE_CB:
         if (skops->args[1] == BPF_TCP_CLOSE) {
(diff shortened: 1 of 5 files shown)


bpf/kmesh/workload/sockops.c
```diff
@@ -106,6 +106,16 @@ static inline void auth_ip_tuple(struct bpf_sock_ops *skops)
     bpf_ringbuf_submit(msg, 0);
 }
 
+// insert an init state to auth_map, indicates that auth is being performed 
+static inline void auth_init(struct bpf_sock_ops *skops)
+{
+    struct bpf_sock_tuple tuple_info = {0};
+    extract_skops_to_tuple(skops, &tuple_info);
+    int err = bpf_map_update_elem(&map_of_auth, &tuple_info, 2, BPF_ANY);
```
hzxuzhonghu, line 114: since we have multi values now, can we define macros for them to prevent hardcoding

bpf/kmesh/workload/sockops.c
```diff
@@ -235,6 +245,7 @@ int sockops_prog(struct bpf_sock_ops *skops)
         if (bpf_sock_ops_cb_flags_set(skops, BPF_SOCK_OPS_STATE_CB_FLAG) != 0)
             BPF_LOG(ERR, SOCKOPS, "set sockops cb failed!\n");
         auth_ip_tuple(skops);
+        auth_init(skops);
```
hzxuzhonghu, line 248: look should mve infront of `auth_ip_tuple`

bpf/kmesh/workload/sockops.c
```diff
@@ -106,6 +106,16 @@ static inline void auth_ip_tuple(struct bpf_sock_ops *skops)
     bpf_ringbuf_submit(msg, 0);
 }
 
+// insert an init state to auth_map, indicates that auth is being performed 
+static inline void auth_init(struct bpf_sock_ops *skops)
```
hzxuzhonghu, line 110: prefer `record_auth_running`

bpf/kmesh/workload/sockops.c
hzxuzhonghu: see macro `AUTH_INIT`, why not use it

hzxuzhonghu (commented):
Assume this is not complete, what is the goal of recording the auth status?

bpf/kmesh/workload/sockops.c
supercharge-xsy: you are right， will  define macros in public header file

supercharge-xsy:
> Assume this is not complete, what is the goal of recording the auth status?

behavior in this pr， drop the message in xdp prog when the tuple is doing auth

hzxuzhonghu:
> drop the message in xdp prog when the tuple is doing auth

@supercharge-xsy If we drop the message, then client side would resend the packet. Have you tested that? 

I would suggest manually increase the auth latency to see how it behaves

supercharge-xsy:
> If we drop the message, then client side would resend the packet. Have you tested that?
> 
> I would suggest manually increase the auth latency to see how it behaves

sure， will try it

supercharge-xsy:
drop packet deteriorates short connection performance. will analyze the feasibility of the xdp acl solution.
/close

PR #426, opened 2024-06-07
add make test for run ut easily

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Fixes #

```release-note

```

### DIFF
--- a/hack/run-ut.sh
+++ b/hack/run-ut.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+ROOT_DIR=$PWD
+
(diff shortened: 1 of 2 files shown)


LiZhenCheng9527:
I'll add the documentation for running ut as soon as I can!

hack/run-ut.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+ROOT_DIR=$PWD
```
hzxuzhonghu, line 3: pwd?

LiZhenCheng9527: fixed

hzxuzhonghu (approved):

hzxuzhonghu:
/lgtm

We should improve the scripts under `./hack`, not sure it they can run sucessfully because we use relative patch `./pkg`

PR #429, opened 2024-06-11
Subscribe cached resources when delta xds reconnect

### LINKED ISSUE #415: Delta xds reconnect may lead to resource leak
**What happened**:

Currently we are uing delta xds for workloads subscribe.  But if a reconnect happens, we may miss knowing which resources are removed during the reconnect.

As delta xds documented  https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol
```
On reconnect the Incremental xDS client may tell the server of its known resources 
to avoid resending them over the network by sending them in initial_resource_versions. 
Because no state is assumed to be preserved from the previous stream, the reconnecting client 
must provide the server with all resource names it is interested in.
```

**Environment**:
- Kmesh version:
- Others:

Okabe-Rintarou-0:
**What type of PR is this?**
bug

Fixes #415

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_controller.go
+++ b/pkg/controller/workload/workload_controller.go
@@ -56,14 +56,16 @@ func (ws *Controller) WorkloadStreamCreateAndSend(client discoveryv3.AggregatedD
 		return fmt.Errorf("DeltaAggregatedResources failed, %s", err)
 	}
 
-	for _, service := range ws.Processor.ServiceCache.List() {
-		// The xds primary key is "namespace/hostname"
-		addressResourceNames = append(addressResourceNames, fmt.Sprintf("%s/%s", service.Namespace, service.Hostname))
-	}
+	if ws.Processor != nil {
+		for _, service := range ws.Processor.ServiceCache.List() {
+			// The xds primary key is "namespace/hostname"
+			addressResourceNames = append(addressResourceNames, fmt.Sprintf("%s/%s", service.Namespace, service.Hostname))
+		}
 
-	for _, workload := range ws.Processor.WorkloadCache.List() {
-		// The xds primary key is "uid"
-		addressResourceNames = append(addressResourceNames, workload.Uid)
+		for _, workload := range ws.Processor.WorkloadCache.List() {
+			// The xds primary key is "uid"
+			addressResourceNames = append(addressResourceNames, workload.Uid)
+		}
 	}
 
 	log.Infof("Address resource names: %v", addressResourceNames)
--- a/pkg/controller/workload/workload_controller.go
+++ b/pkg/controller/workload/workload_controller.go
@@ -46,14 +46,29 @@ func NewController(workloadMap bpf2go.KmeshCgroupSockWorkloadMaps) *Controller {
 }
 
 func (ws *Controller) WorkloadStreamCreateAndSend(client discoveryv3.AggregatedDiscoveryServiceClient, ctx context.Context) error {
-	var err error
+	var (
+		err                  error
+		addressResourceNames []string
+	)
 
 	ws.Stream, err = client.DeltaAggregatedResources(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/workload/workload_controller.go
```diff
@@ -46,14 +46,31 @@ func NewController(workloadMap bpf2go.KmeshCgroupSockWorkloadMaps) *Controller {
...
 	ws.Stream, err = client.DeltaAggregatedResources(ctx)
 	if err != nil {
 		return fmt.Errorf("DeltaAggregatedResources failed, %s", err)
 	}
 
-	if err := ws.Stream.Send(newWorkloadRequest(AddressType, nil)); err != nil {
+	if ws.Processor != nil {
+		for _, service := range ws.Processor.ServiceCache.List() {
+			// The xds primary key is "namespace/hostname"
+			addressResourceNames = append(addressResourceNames, fmt.Sprintf("%s/%s", service.Namespace, service.Hostname))
```
hzxuzhonghu, line 62: nit: use service.ResourceName()

pkg/controller/workload/workload_controller.go
```diff
@@ -46,14 +46,31 @@ func NewController(workloadMap bpf2go.KmeshCgroupSockWorkloadMaps) *Controller {
...
+		}
+
+		for _, workload := range ws.Processor.WorkloadCache.List() {
+			// The xds primary key is "uid"
+			addressResourceNames = append(addressResourceNames, workload.Uid)
+		}
+	}
+
+	log.Infof("Address resource names: %v", addressResourceNames)
```
hzxuzhonghu, line 71:
```suggestion
	log.Debugf("Initial address resource names: %v", addressResourceNames)
```

hzxuzhonghu (commented):
nit: we can preallocate capacity for the slice via make([]string, 0, len(xx))

pkg/controller/workload/workload_controller.go
```diff
@@ -46,14 +46,34 @@ func NewController(workloadMap bpf2go.KmeshCgroupSockWorkloadMaps) *Controller {
...
+		}
+
+		for _, workload := range cachedWorkloads {
+			addressResourceNames = append(addressResourceNames, workload.ResourceName())
+		}
+	}
+
+	log.Debugf("Initial address resource names: %v", addressResourceNames)
+
+	if err := ws.Stream.Send(newWorkloadRequest(AddressType, addressResourceNames)); err != nil {
```
hzxuzhonghu, line 76: This is not right, you should set the `InitialResourceVersions`

hzxuzhonghu: For the versions, we can set empty "", because the server side does not check them now

hzxuzhonghu: right, `ResourceNamesSubscribe` unset means client is subscribing all the resources.

Only for on demand xds this can be set to specific values

pkg/controller/workload/workload_controller.go
```diff
@@ -46,14 +46,34 @@ func NewController(workloadMap bpf2go.KmeshCgroupSockWorkloadMaps) *Controller {
...
 	ws.Stream, err = client.DeltaAggregatedResources(ctx)
 	if err != nil {
 		return fmt.Errorf("DeltaAggregatedResources failed, %s", err)
 	}
 
-	if err := ws.Stream.Send(newWorkloadRequest(AddressType, nil)); err != nil {
+	if ws.Processor != nil {
+		cachedServices := ws.Processor.ServiceCache.List()
+		cachedWorkloads := ws.Processor.WorkloadCache.List()
+		initialResourceVersions = make(map[string]string)
```
hzxuzhonghu, line 62: nit: make(map[string]string, len(cachedServices)+len(cachedWorkloads))

hzxuzhonghu (approved):

hzxuzhonghu (approved):
/lgtm

PR #435, opened 2024-06-13
Fix authPolicy remove

hzxuzhonghu:
**What type of PR is this?**

Fix the issue: a authorization deleted, the same namespace scoped authorization will be mis removed

Fixes #

```release-note

```

### DIFF
--- a/pkg/auth/rbac.go
+++ b/pkg/auth/rbac.go
@@ -70,7 +70,7 @@ type rbacConnection struct {
 
 func NewRbac(workloadObj *bpf.BpfKmeshWorkload, workloadCache cache.WorkloadCache) *Rbac {
 	return &Rbac{
-		policyStore:   newPolicystore(),
+		policyStore:   newPolicyStore(),
 		workloadCache: workloadCache,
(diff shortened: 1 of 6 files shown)


supercharge-xsy:
The modification is fine, but seems that the test case needs to be modified accordingly.

pkg/auth/rbac.go
```diff
@@ -160,14 +160,16 @@ func (r *Rbac) RemovePolicy(policyKey string) {
...
-		dstWorkload = r.workloadCache.GetWorkloadByAddr(cache.NetworkAddress{
-			Network: conn.dstNetwork,
-			Address: nets.ConvertIpByteToUint32(conn.dstIp),
-		})
+	dstWorkload := r.workloadCache.GetWorkloadByAddr(cache.NetworkAddress{
+		Network: conn.dstNetwork,
+		Address: nets.ConvertIpByteToUint32(conn.dstIp),
+	})
+	// If no workload found, deny
+	if dstWorkload == nil {
```
hzxuzhonghu, line 168: @supercharge-xsy This is to align with ztunnel

PR #472, opened 2024-07-02
Split E2E test as a separate CI jobs

### LINKED ISSUE #467: Split E2E test as a separate CI jobs
One CI job can only run in searial, with more and more e2e cases added, it will take more time.

And the second reason is that, if one step failed, it will stop.

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Fixes #467 

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,4 +19,4 @@ jobs:
     - name: E2E Test
       shell: bash
       run: |
-        sudo env "PATH=$PATH" bash ./test/e2e/run_test.sh
+        sudo make e2e
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -17,5 +17,6 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
     - name: E2E Test
+      shell: bash
       run: |
-        make e2e
+        sudo env "PATH=$PATH" bash ./test/e2e/run_test.sh
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,21 @@
+name: E2E Test
+on: 
+  pull_request:
+jobs:
+  e2e-test:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        go-version: [ '1.22' ]
+    name: E2E Test
+    timeout-minutes: 40
+
+    steps:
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (approved):
/lgtm

.github/workflows/e2e.yml
```diff
@@ -17,5 +17,6 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
     - name: E2E Test
+      shell: bash
       run: |
-        make e2e
+        sudo env "PATH=$PATH" bash ./test/e2e/run_test.sh
```
hzxuzhonghu, line 22: why set `env "PATH=$PATH"`

hzxuzhonghu: and why not make e2e

LiZhenCheng9527: If run `make e2e` directly, there are permission problems.
image

Set `env "PATH=$PATH"` is refer to https://github.com/kmesh-net/kmesh/blob/da515ea41ea5464992e4cd22140bf28b624f0536/.github/workflows/main.yml#L30

Since the e2e tests also require `Build Kmesh`, it is also straightforward to set up env to run the e2e scripts.

hzxuzhonghu: I guess sudo make e2e works.

LiZhenCheng9527: done

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup Go
+      uses: actions/setup-go@v4.0.0
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: E2E Test
+      shell: bash
+      run: |
+        sudo make e2e
```
YaoZengzeng, line 22: E2E test would build Kmesh image and push it to the local registry and there is also a `Build Kmesh` step in CI. Can the two be combined into one? Simply delete `Build Kmesh` or let E2E use the result of `Build Kmesh`.

LiZhenCheng9527: The two github workflows are running in different virtual environments. So the image built by the other workflow cannot be used directly.

hzxuzhonghu (approved):
/lgtm

PR #482, opened 2024-07-03
fix hash name bug (#479)

### LINKED ISSUE #479: HashName can output different hash value for same str
HashName.StrToNum can allocate different hash value to a str when some previous allocated values not used anymore

Okabe-Rintarou-0:
**What type of PR is this?**
bug

Fixes #479 

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_hash.go
+++ b/pkg/controller/workload/workload_hash.go
@@ -25,25 +25,38 @@ var (
 	hash = fnv.New32a()
 )
 
+const tombstoneMarker = "\x00"
+
 // HashName converts a string to a uint32 integer as the key of bpf map
 type HashName struct {
 	numToStr map[uint32]string
+	// records its tombstone number given a string
+	tombstones map[string]uint32
 }
 
 func NewHashName() *HashName {
 	con := &HashName{}
 	con.numToStr = make(map[uint32]string)
+	con.tombstones = make(map[string]uint32)
 	return con
 }
 
 func (h *HashName) StrToNum(str string) uint32 {
 	var num uint32
 
+	if num, exists := h.tombstones[str]; exists {
+		h.numToStr[num] = str
+		delete(h.tombstones, str)
+		return num
+	}
+
 	hash.Reset()
 	hash.Write([]byte(str))
 
 	// Using linear probing to solve hash conflicts
 	for num = hash.Sum32(); num < math.MaxUint32; num++ {
+		// We will keep searching until we find an unused slot
+		// We won't use the slot with tombstone marker
 		if h.numToStr[num] == "" {
 			h.numToStr[num] = str
 			break
@@ -56,9 +69,16 @@ func (h *HashName) StrToNum(str string) uint32 {
 }
 
 func (h *HashName) NumToStr(num uint32) string {
-	return h.numToStr[num]
+	str := h.numToStr[num]
+	if str != tombstoneMarker {
+		return str
+	}
+	return ""
 }
 
 func (h *HashName) Delete(str string) {
-	delete(h.numToStr, h.StrToNum(str))
+	// instead of directly deleting, we put a tombstone here
+	num := h.StrToNum(str)
+	h.numToStr[num] = tombstoneMarker
+	h.tombstones[str] = num
 }
--- a/pkg/controller/workload/workload_hash.go
+++ b/pkg/controller/workload/workload_hash.go
@@ -17,50 +17,80 @@
 package workload
 
 import (
+	"encoding/json"
 	"hash/fnv"
 	"math"
+	"os"
 )
 
 var (
 	hash = fnv.New32a()
 )
 
-const tombstoneMarker = "\x00"
+const (
+	persistPath = "/mnt/workload_hash_name.json"
+)
 
 // HashName converts a string to a uint32 integer as the key of bpf map
 type HashName struct {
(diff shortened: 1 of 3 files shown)


pkg/controller/workload/workload_hash.go
```diff
@@ -25,25 +25,38 @@ var (
 	hash = fnv.New32a()
 )
 
+const tombstoneMarker = "\x00"
+
 // HashName converts a string to a uint32 integer as the key of bpf map
 type HashName struct {
 	numToStr map[uint32]string
+	// records its tombstone number given a string
+	tombstones map[string]uint32
```
hzxuzhonghu, line 34: This is kind of mem leak. Our requirement is:

1. a str should be allocated with a consistent value
2. after it deleted, and then at some time recreated, the value can be different.
3. It's better to be consistent for the case that kmesh daemon restart, which seems difficult with this implement. Maybe we can persist the maps somewhere.

Okabe-Rintarou-0: > 2\. after it deleted, and then at some time recreated, the value can be different.

Requirement 1 means for a str like "foo", if we call StrToNum we will get number 123. Then at anytime we StrToNum("foo") should always get 123? Even if collision happened?
Can you give an example about requirement 2?

Okabe-Rintarou-0: In the original implemention, there exists another problem.
If A collides with B.
If StrToNum(A) = 123, then StrToNum(B) = 124
Then we delete A, we will delete numToStr[123].
Then we delete B, we will also delete numToStr[123], cause Delete will call StrToNum, where StrToNum(B) will be 123, if A is deleted(since there is no collision any more). In this case, we will leave numToStr[124] = B, which is also a kind of mem leak.

hzxuzhonghu: >after it deleted, and then at some time recreated, the value can be different.


say `foo` at t0 is allcated `1234`, then at t1, it is deleted. In our case, all the info is deleted from bpf map.

Maybe t2, `foo` is recreated, then we can allocate it to `2345`.  It need to be consistent during its lifecycle.

Okabe-Rintarou-0: I got it. We need to persist the numtostr map in case kmesh restarts.

Okabe-Rintarou-0: > In the original implemention, there exists another problem. If A collides with B. If StrToNum(A) = 123, then StrToNum(B) = 124 Then we delete A, we will delete numToStr[123]. Then we delete B, we will also delete numToStr[123], cause Delete will call StrToNum, where StrToNum(B) will be 123, if A is deleted(since there is no collision any more). In this case, we will leave numToStr[124] = B, which is also a kind of mem leak.

to solve this problem, we will still need the  tombstone marker

Okabe-Rintarou-0: We only need to store string that cause collision.
For example, if A collides with B.
We call StrToNum(A), then StrToNum(B).
A got "123", B got "124"
We only need to persist (B, 124). Since after restart, we call StrToNum(B) we can get StrToNum(B) = 124 from persisted map. And StrToNum(A) = 123 according to hash value.

pkg/controller/workload/workload_hash.go
```diff
@@ -42,12 +78,20 @@ func (h *HashName) StrToNum(str string) uint32 {
...
 	for num = hash.Sum32(); num < math.MaxUint32; num++ {
-		if h.numToStr[num] == "" {
+		// Create a new item if we find an empty slot
+		if _, exists := h.numToStr[num]; !exists {
 			h.numToStr[num] = str
-			break
-		} else if h.numToStr[num] == str {
+			h.strToNum[str] = num
+			// Create a new item here, should flush
+			if err := h.flush(); err != nil {
```
hzxuzhonghu, line 92: Can we write delta data? Flush all seems very expensive

pkg/controller/workload/workload_hash.go
```diff
@@ -42,12 +78,20 @@ func (h *HashName) StrToNum(str string) uint32 {
 	hash.Reset()
 	hash.Write([]byte(str))
 
+	if num, exits := h.strToNum[str]; exits {
+		return num
+	}
+
 	// Using linear probing to solve hash conflicts
 	for num = hash.Sum32(); num < math.MaxUint32; num++ {
```
hzxuzhonghu, line 86: not related to your code。

If num overflows, should we iterate from 0

pkg/controller/workload/workload_hash.go
Okabe-Rintarou-0: Sure. What about appending?

pkg/controller/workload/workload_hash.go
Okabe-Rintarou-0: I think it's necessary. It should be a ring.

pkg/controller/workload/workload_hash.go
Okabe-Rintarou-0: > maybe yaml is simpler
> 
> ```
> foo: 123
> bar: 124
> ```

right, that's a better idea.

hzxuzhonghu (approved):
/lgtm

we can use the pkg `sigs.k8s.io/yaml`, which is from k8s ecosystem

PR #490, opened 2024-07-05
Downgrade go version to v1.21.4

tacslon:
**What type of PR is this?**
/kind feature

Downgrade go version to v1.21.4
Fixes #

```release-note
NONE
```


hzxuzhonghu (changes requested):
Downgrade go version is not a common op, unless the current version has serious problem

YaoZengzeng:
Also E2E test framework depends on a very new version of istio, which uses go 1.22, so go 1.21 may not work and CI does fail.

PR #499, opened 2024-07-05
fix flaky e2e test

YaoZengzeng:
**What type of PR is this?**

/kind bug

E2E test case `TestTrafficSplit` was previously incomplete as it was found to be unstable during test.

Now find reason is that if the test application is deployed before Kmesh Ready, it will not managed by Kmesh.

The solution is to ensure Kmesh runs successfully before proceeding with the subsequent steps.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/run_test.sh
+++ b/test/e2e/run_test.sh
@@ -78,11 +78,34 @@ function setup_istio() {
     kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
         { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v1.1.0" | kubectl apply -f -; }
 
-    istioctl install --set profile=ambient --skip-confirmation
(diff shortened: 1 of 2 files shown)


test/e2e/run_test.sh
```diff
@@ -78,11 +78,34 @@ function setup_istio() {
...
 
-    istioctl install --set profile=ambient --skip-confirmation
+    istioctl install --set profile=ambient --set meshConfig.accessLogFile="/dev/stdout" --skip-confirmation
 }
 
 function setup_kmesh() {
     helm install kmesh $ROOT_DIR/deploy/helm -n kmesh-system --create-namespace --set deploy.kmesh.image.repository=localhost:5000/kmesh
+
+    # Wait for all Kmesh pods to be ready.
+    while true; do
```
hzxuzhonghu, line 88: nit: we can abstract a function like this

https://github.com/kurator-dev/kurator/blob/main/hack/util.sh#L105-L120

hzxuzhonghu (approved):
/lgtm

PR #506, opened 2024-07-06
cleanup stopch(#497)

### LINKED ISSUE #497: Cleanup:  eliminating stopCh defined everywhere
This is the searching result.

https://github.com/search?q=repo%3Akmesh-net%2Fkmesh%20make(chan%20struct%7B%7D)&type=code

By passing stop channel to each module's Start/Run method, we can control each module globally. And even more, by doing so, we could further eliminate `defer stop` stuff
```
	c := controller.NewController(configs, bpfLoader.GetBpfKmeshWorkload(), configs.BpfConfig.BpfFsPath, configs.BpfConfig.EnableBpfLog)
	if err := c.Start(); err != nil {
		return err
	}
	log.Info("controller Start successful")
	defer c.Stop()
```

Okabe-Rintarou-0:
**What type of PR is this?**
enhance

Fixes #497 

```release-note

```

### DIFF
--- a/pkg/controller/bypass/bypass_controller.go
+++ b/pkg/controller/bypass/bypass_controller.go
@@ -48,8 +48,7 @@ const (
 	SidecarAnnotation         = "sidecar.istio.io/inject"
 )
 
-func StartByPassController(client kubernetes.Interface) error {
-	stopChan := make(chan struct{})
+func StartByPassController(client kubernetes.Interface, stopChan <-chan struct{}) error {
 	nodeName := os.Getenv("NODE_NAME")
 
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(client, DefaultInformerSyncPeriod,
--- a/daemon/manager/manager.go
+++ b/daemon/manager/manager.go
@@ -80,8 +80,11 @@ func Execute(configs *options.BootstrapConfigs) error {
 	log.Info("bpf Start successful")
 	defer bpfLoader.Stop()
 
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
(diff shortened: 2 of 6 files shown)


pkg/controller/bypass/bypass_controller.go
```diff
@@ -48,8 +48,7 @@ const (
 	SidecarAnnotation         = "sidecar.istio.io/inject"
 )
 
-func StartByPassController(client kubernetes.Interface) error {
-	stopChan := make(chan struct{})
+func StartByPassController(client kubernetes.Interface, stopChan <-chan struct{}) error {
```
LiZhenCheng9527, line 51: Can you replace stopChan with context.Context

Okabe-Rintarou-0: you mean all of the stopch? I am confused about why there exist both ctx and stopCh.

Okabe-Rintarou-0: https://github.com/kmesh-net/kmesh/blob/ba94cbc1be2650d4a8b38adb1aefa62a764735c6/pkg/controller/bypass/bypass_controller.go#L135

informer needs a stopch

hzxuzhonghu (commented):
LGTM, I would do a little bit further, pass in from `func Execute(configs *options.BootstrapConfigs) error {`

pkg/controller/bypass/bypass_controller.go
LiZhenCheng9527: Indeed, the informer need

hzxuzhonghu (approved):
/lgtm

Issue #291, opened 2024-05-02
IPv6 support


PerforMance308:
**What would you like to be added**:
Based on issues #244 , it was found that kmesh does not currently support ipv6
**Why is this needed**:
If using higher version of k8s, JVM will enable IPv6 by default, in this case , kmesh couldn't handle request between java services

hzxuzhonghu:
cc @nlgwcy

supercharge-xsy:
we need to consider the IPv4-mapped IPv6

YaoZengzeng:
Waypoint should support decoding IPV6 address as well.

nlgwcy:
[PR_434](https://github.com/kmesh-net/kmesh/pull/434) covers only IPv6 compatibility reconstruction.
1. workload mode supports IPv6
   - [x] IPv6-compatible reconstruction on the control plane
   - [x] IPv6-compatible reconstruction on the data plane

hzxuzhonghu:
This is supported now, will be released in v0.4.0

PR #530, opened 2024-07-09
waypoint listen both ipv4 and ipv6

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

To support ipv6, waypoint should listen to both ipv4 and ipv6.

Fixes part of #291

```release-note

```

### DIFF
--- a/deploy/helm/templates/l7-envoyfilter.yaml
+++ b/deploy/helm/templates/l7-envoyfilter.yaml
@@ -18,6 +18,12 @@ spec:
             protocol: TCP
             address: 0.0.0.0
             port_value: 15019
+        additional_addresses:
+        - address:
+            socket_address:
+              protocol: TCP
+              address: "::"
+              port_value: 15019
         filter_chains:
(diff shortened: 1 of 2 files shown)


deploy/helm/templates/l7-envoyfilter.yaml
```diff
@@ -18,6 +18,12 @@ spec:
             protocol: TCP
             address: 0.0.0.0
             port_value: 15019
+        additional_addresses:
+        - address:
+            socket_address:
+              protocol: TCP
+              address: "::"
```
hzxuzhonghu, line 25: will it fail to start in ipv4 env

YaoZengzeng: k8s enable dual stack by default starting from 1.21, ref: https://kubernetes.io/docs/concepts/services-networking/dual-stack/

hzxuzhonghu (approved):
/lgtm

Issue #547, opened 2024-07-10
Unused BPF_LOG_LEVEL macro


Okabe-Rintarou-0:
**Task description**:
https://github.com/kmesh-net/kmesh/blob/9c3c136ee3a0afb88ee3b42485e58ac6ed0896e1/bpf/include/bpf_log.h#L12
`BPF_LOG_LEVEL` macro now is unused.

There is another problem. Before, the bpf default log level is **DEBUG**, but it's **ERROR** by default.

**Solution**:

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

After a contributor has successfully completed 1-2 good first issue's,
they should be ready to move on to `help wanted` items, saving the remaining `good first issue` for other new contributors.

**How to join or take the task**:

Just reply on the issue with the message `/assign` in a separate line.

Then, the issue will be assigned to you.

**How to ask for help**:

If you need help or have questions, please feel free to ask on this issue.
The issue author or other members of the community will guide you through the contribution process.

hzxuzhonghu:
BTW, BPF_LOG macro does not work as expected. If we set bpf log level to debug, some log types cannot be printed out.

hzxuzhonghu:
Some intuit thought, we may make `INFO` level as default for all log types. And update the BPF_LOG macro to compare the particular log level with the default value, if it is LE the default value, we will allow logging. 

Unless we want to support fine-grained log type level adjustment, we can keep all log types using same value now

PR #557, opened 2024-07-11
Provide a way  to allow setting all logger level to debug

hzxuzhonghu:
Fix: https://github.com/kmesh-net/kmesh/issues/547


bpf/include/bpf_log.h
```diff
@@ -6,18 +6,19 @@
...
 
-#define BPF_LOG_LEVEL BPF_LOG_DEBUG
+// set all log level to BPF_LOG_INFO by default
+// TODO: support to customize fine-grained log level
+#define BPF_LOGTYPE_SOCKMAP BPF_LOG_LEVEL
+#define BPF_LOGTYPE_KMESH   BPF_LOG_LEVEL
+#define BPF_LOGTYPE_SOCKOPS BPF_LOG_LEVEL
+#define BPF_LOGTYPE_XDP     BPF_LOG_LEVEL
+#define BPF_LOGTYPE_SENDMSG BPF_LOG_LEVEL
+#define BPF_LOGTYPE_PROBE   BPF_LOG_LEVEL
```
hzxuzhonghu, line 19: These are now useless until we support fine-grained log level control

hzxuzhonghu:
@nlgwcy Because of the instruction limit of ebpf, this is the most we can do now. 

Luckily with this, we can set log level to almost all the bpf prog, except `COMMON`

bpf/include/bpf_log.h
```diff
@@ -101,7 +101,7 @@ static inline int map_lookup_log_level()
     int *value = NULL;
     value = kmesh_map_lookup_elem(&bpf_log_level, &zero);
     if (!value)
-        return 0;
+        return BPF_LOG_INFO;
```
hzxuzhonghu, line 104: This is a change, now no debug level log will be printed by default. Which can be tuned by user from kmesh-daemon.

bpf/kmesh/ads/include/kmesh_common.h
```diff
@@ -9,12 +9,12 @@
...
-#define BPF_LOGTYPE_CLUSTER       BPF_DEBUG_OFF
-#define BPF_LOGTYPE_ROUTER        BPF_DEBUG_OFF
-#define BPF_LOGTYPE_ROUTER_CONFIG BPF_DEBUG_OFF
+#define BPF_LOGTYPE_LISTENER      BPF_DEBUG_ON
+#define BPF_LOGTYPE_FILTERCHAIN   BPF_DEBUG_ON
+#define BPF_LOGTYPE_FILTER        BPF_DEBUG_ON
+#define BPF_LOGTYPE_CLUSTER       BPF_DEBUG_ON
+#define BPF_LOGTYPE_ROUTER        BPF_DEBUG_ON
+#define BPF_LOGTYPE_ROUTER_CONFIG BPF_DEBUG_ON
 #define BPF_LOGTYPE_COMMON        BPF_DEBUG_OFF
```
nlgwcy, line 18: image
In debug level, this log will be printed many times, but it does not provide useful information, So I think it would be removed,  and then the switch of BPF_LOGTYPE_COMMON can be changed to BPF_DEBUG_ON.

hzxuzhonghu: Will try

PR #582, opened 2024-07-16
Fix `make gen` problem

### LINKED ISSUE #571: `make gen` generate proto problem
**What happened**:

Run `make gen` locally produces some differences:
image

```
[root@localhost kmesh]# protoc --version
libprotoc 3.17.3
```

**Environment**:
- Kmesh version:
- Others:

Okabe-Rintarou-0:
**What type of PR is this?**
bug

Fix `make gen` problem, upgrate protoc-c to 1.4.x, aligning with the native one in openEuler 23.03.

Fixes #571 

```release-note

```


hack/gen_protoc.sh
```diff
@@ -0,0 +1,24 @@
...
+# if protoc-c is 1.4.x
+if protoc-c --version | grep -q -E "protobuf-c 1\.4\.[0-9]"; then
+    protoc-c --proto_path=$PROTO_PATH --c_out=. $PROTO_SRC
+else
+    echo "Generate proto files in docker"
+    docker run --rm \
+        --privileged=true \
+        -v $ROOT_DIR:/kmesh \
+        --name kmesh-build \
+        ghcr.io/kmesh-net/kmesh-build-${arch}:latest \
```
YaoZengzeng, line 21: Does the build image already support multi arch? @lec-bit

hack/install-proto.sh
```diff
@@ -3,6 +3,7 @@
 function install_tool () {
     tool=${1}
     if command -v apt > /dev/null; then
+        sudo apt-get update
```
hzxuzhonghu, line 6: indents

hack/gen_protoc.sh
hzxuzhonghu: https://github.com/orgs/kmesh-net/packages/container/package/kmesh-build

We have supported both arm and x86， so can remove arch suffix

Okabe-Rintarou-0: > https://github.com/orgs/kmesh-net/packages/container/package/kmesh-build
> 
> We have supported both arm and x86， so can remove arch suffix

then we should also remove the arch preffix in make build?

hzxuzhonghu (commented):
/lgtm

/approve

PR #590, opened 2024-07-16
optimize workload update

### LINKED ISSUE #503: Improve HandleWorkload prevent losing control during workload update

Currently on workload update， it will call `handleDataWithService`, and then first delete `deletePodFrontendData`. 
And at the same time, if a connect request to the pod with pod ip, kmesh definitely lose control on this connect. 

So the right order of updating bpf map should be reconsidered.
We have totally `frontend_map` `service_map` `endpoint_map` and `backend_map`.  The detialed procedure of traffic managing can be found [here](https://github.com/kmesh-net/kmesh/blob/main/docs/proposal/kmesh_support_workload.md#traffic-governance-process)

To prevent inferencing traffic management, we should update bpf map with the opposite orders that bpf prog uses.
From my perspective should be: 1. backend_map 2. endpoint_map (if exists) 3. service_map(if exists) 4. frontend_map

cc @nlgwcy

nlgwcy:
**What type of PR is this?**
/kind enhancement

Even if a workload is associated with a service, the workload can still be directly accessed.
Fixes #503

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -385,14 +385,8 @@ func (p *Processor) handleDataWithService(workload *workloadapi.Workload) error 
 		bk = bpf.BackendKey{}
 		bv = bpf.BackendValue{}
 	)
-	backend_uid := p.hashName.StrToNum(workload.GetUid())
-	// a Pod may be added to a certain service in the future, so it is necessary to delete the Pod info
-	// that was previously added as an independent Pod to the frontend map.
-	if err = p.deletePodFrontendData(backend_uid); err != nil {
-		log.Errorf("deletePodFrontendData failed, err:%s", err)
-		return err
-	}
 
+	backend_uid := p.hashName.StrToNum(workload.GetUid())
 	for serviceName := range workload.GetServices() {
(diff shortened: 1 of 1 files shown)


pkg/controller/workload/workload_processor.go
```diff
@@ -385,14 +385,8 @@ func (p *Processor) handleDataWithService(workload *workloadapi.Workload) error
 		bk = bpf.BackendKey{}
 		bv = bpf.BackendValue{}
 	)
-	backend_uid := p.hashName.StrToNum(workload.GetUid())
-	// a Pod may be added to a certain service in the future, so it is necessary to delete the Pod info
-	// that was previously added as an independent Pod to the frontend map.
-	if err = p.deletePodFrontendData(backend_uid); err != nil {
```
hzxuzhonghu, line 391: should we track the case pod ip changed in a issue

hzxuzhonghu (approved):
/lgtm

pkg/controller/workload/workload_processor.go
nlgwcy: > should we track the case pod ip changed in a issue

yes, it will be supported later.

PR #598, opened 2024-07-17
fix pod management can not be canceled

### LINKED ISSUE #568: Pod management can not be canceled
**What happened**:
after I manage traffic in pod level, management can not be canceled
management should be canceled
kubectl label pod xxx istio.io/dataplane-mode=kmesh
kubectl label pod xxx istio.io/dataplane-mode-

**Environment**:
- Kmesh version:
- Others:

weli-l:
**What type of PR is this?**
fix pod management can not be canceled

Fixes #568 

```release-note

```

### DIFF
--- a/pkg/controller/manage/kmesh_manage.go
+++ b/pkg/controller/manage/kmesh_manage.go
@@ -210,9 +210,6 @@ func shouldEnroll(client kubernetes.Interface, pod *corev1.Pod) bool {
 	if strings.EqualFold(pod.Labels[constants.DataPlaneModeLabel], constants.DataPlaneModeKmesh) {
 		return true
 	}
-	if pod.Annotations[constants.KmeshRedirectionAnnotation] == "enabled" {
-		return true
-	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/manage/kmesh_manage.go
```diff
@@ -210,9 +210,6 @@ func shouldEnroll(client kubernetes.Interface, pod *corev1.Pod) bool {
 	if strings.EqualFold(pod.Labels[constants.DataPlaneModeLabel], constants.DataPlaneModeKmesh) {
 		return true
 	}
-	if pod.Annotations[constants.KmeshRedirectionAnnotation] == "enabled" {
```
hzxuzhonghu, line 213: @lec-bit This fix would break the secret manager assumption, which heavily rely on this to count numbers

hzxuzhonghu:
@weli-l we should first open pr to main branch, then cherry pick back

pkg/controller/manage/kmesh_manage.go
lec-bit: ok，i will test and Modify

Issue #597, opened 2024-07-17
Remove default resync period


hzxuzhonghu:
Currently for the informer we set resync period to 30s, which means every 30s, it will get triggered with an update event.

PR #599, opened 2024-07-18
Fix a miscommit

lec-bit:
The last mistaken submission caused performance degradation and slow log printing.

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/manage/kmesh_manage.go
+++ b/pkg/controller/manage/kmesh_manage.go
@@ -54,7 +54,7 @@ var (
 )
 
 const (
-	DefaultInformerSyncPeriod = 2 * time.Second
+	DefaultInformerSyncPeriod = 30 * time.Second
 	MaxRetries                = 5
 	ActionAddAnnotation       = "add"
 	ActionDeleteAnnotation    = "delete"


lec-bit:
Multiple interactions take up a lot of performance, resulting in slow pod management, which may take more than ten seconds to respond.

pkg/controller/manage/kmesh_manage.go
```diff
@@ -54,7 +54,7 @@ var (
 )
 
 const (
-	DefaultInformerSyncPeriod = 2 * time.Second
+	DefaultInformerSyncPeriod = 30 * time.Second
```
LiZhenCheng9527, line 57: Is there no way to optimise it?

lec-bit: Synchronization here shouldnot be too frequent, and the interval should be as long as possible. In fact, if there are changes, they will be pushed over immediately, and there is no need to refresh them here regularly.

hzxuzhonghu: I would suggest removing resync https://github.com/kmesh-net/kmesh/issues/597

hzxuzhonghu:
not needed anymore

PR #601, opened 2024-07-18
remove resync period

### LINKED ISSUE #597: Remove default resync period

hzxuzhonghu:
**What type of PR is this?**

Fixes #597

```release-note

```

### DIFF
--- a/pkg/controller/bypass/bypass_controller.go
+++ b/pkg/controller/bypass/bypass_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"time"
 
 	netns "github.com/containernetworking/plugins/pkg/ns"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/bypass/bypass_controller.go
```diff
@@ -42,16 +41,14 @@ var (
 )
 
 const (
-	DefaultInformerSyncPeriod = 30 * time.Second
-	LabelSelectorBypass       = "kmesh.net/bypass=enabled"
-	KmeshAnnotation           = "kmesh.net/redirection"
-	SidecarAnnotation         = "sidecar.istio.io/inject"
+	KmeshAnnotation   = "kmesh.net/redirection"
```
LiZhenCheng9527, line 44: There is a `KmeshRedirectionAnnotation` in constant.go

PR #608, opened 2024-07-18
fix dns cannot remove domain bug

PerforMance308:
**What type of PR is this?**
/kind bug

image

`Stow` solution. should send dnsClusters to DNS resolver even if it is empty so that the resolver could remove unwatched domains.
Fixes #

```release-note

```

### DIFF
--- a/pkg/dns/dns_test.go
+++ b/pkg/dns/dns_test.go
@@ -421,3 +421,85 @@ func TestGetPendingResolveDomain(t *testing.T) {
 		})
 	}
 }
+
+func TestRemoveUnwatchedCluster(t *testing.T) {
+	fakeDNSServer := newFakeDNSServer()
+
+	testDNSResolver, err := NewDNSResolver(ads.NewAdsCache())
+	if err != nil {
+		t.Fatal(err)
+	}
+	stopCh := make(chan struct{})
+	testDNSResolver.StartDNSResolver(stopCh)
+	testDNSResolver.resolvConfServers = []string{fakeDNSServer.Server.PacketConn.LocalAddr().String()}
+
+	utClusterWithHost := clusterv3.Cluster{
+		Name: "testCluster",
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
LG

Do we have a case covering the domain can be removed correctly

PerforMance308:
added ut for  checking the domain can be removed correctly

pkg/dns/dns_test.go
```diff
@@ -324,3 +324,86 @@ func (s *fakeDNSServer) setTTL(ttl uint32) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.domain != "" {
+				fakeDNSServer.setHosts(tt.domain, 1)
+			}
+			testDNSResolver.resolveDomains(tt.args.clusters)
+			time.Sleep(2 * time.Second)
```
hzxuzhonghu, line 403: I think this is unneeded

hzxuzhonghu (commented):
I wish to add a test covering the fix

pkg/dns/dns_test.go
```diff
@@ -421,3 +421,84 @@ func TestGetPendingResolveDomain(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.domain != "" {
+				fakeDNSServer.setHosts(tt.domain, 1)
+			}
+			testDNSResolver.resolveDomains(tt.args.clusters)
+			if _, ok := testDNSResolver.cache[tt.domain]; ok != tt.want {
```
hzxuzhonghu, line 499: I donot think this case can cover the removing sceanrio

hzxuzhonghu:
@PerforMance308 Could you update the tests and we can merge

PR #611, opened 2024-07-19
waypoint should not managed by Kmesh

LiZhenCheng9527:
**What type of PR is this?**
/kind bug

waypoint should not managed by Kmesh
Fixes #

```release-note

```


hzxuzhonghu (commented):
/lgtm

I checked CNI plugin, where we use `checkKmesh`. It is quite different from here.

I would like to align with that, it is more resilient

PR #621, opened 2024-07-21
Support ipv6 in e2e test

### LINKED ISSUE #529: [E2E Enchancement] run e2e test in pure ipv6 environment

Now E2E tests only run in ipv4 environment. Now Kmesh already supports ipv6, we should build an ipv6 environment in CI to run E2E test cases to ensure the availability of all functions under ipv6.

**NOTE: dual stack is not supported yet**

noobwei:
**What type of PR is this?**

/kind enhancement

Support pure ipv6 in e2e test, to fit in the new feature in Kmesh

Fixes #529 

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,3 +20,22 @@ jobs:
       shell: bash
       run: |
         sudo make e2e
+  
+  e2e-ipv6-test:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        go-version: [ '1.22' ]
+    name: E2E IPv6 Test
+    timeout-minutes: 40
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup Go
+      uses: actions/setup-go@v4.0.0
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: E2E IPv6 Test
+      shell: bash
+      run: |
+        sudo make e2e-ipv6
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,8 @@ docker.push: docker
 
 e2e:
 	./test/e2e/run_test.sh
+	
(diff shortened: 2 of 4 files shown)


test/e2e/run_test.sh
```diff
@@ -28,7 +28,25 @@ function setup_kind_cluster() {
     fi
 
     # Create KinD cluster.
-    cat <<EOF | kind create cluster --name="${NAME}" -v4 --retain --image "${IMAGE}" --config=-
+
```
YaoZengzeng, line 31: You should also add a line here for ipv6 e2e call:

https://github.com/kmesh-net/kmesh/blob/main/Makefile#L149

test/e2e/run_test.sh
```diff
@@ -28,7 +28,25 @@ function setup_kind_cluster() {
...
+    if [[ -n "${IPV6:-}" ]]; then
```
YaoZengzeng, line 32: This `if` has no corresponding `fi`?

hzxuzhonghu (commented):
Can you add a CI job for ipv6  like .github/workflows/e2e.yml

Makefile
```diff
@@ -148,6 +148,8 @@ docker.push: docker
 
 e2e:
```
YaoZengzeng, line 149: Make ipv6 an env, if we want to run e2e test for ipv6 run:

```
IPFAMILY=ipv6 make e2e
```

.github/workflows/e2e.yml
```diff
@@ -20,3 +20,22 @@ jobs:
       shell: bash
       run: |
         sudo make e2e
+  
+  e2e-ipv6-test:
```
hzxuzhonghu, line 24: why not separate it to e2e-ipv6.yml

hzxuzhonghu (approved):
/lgtm

PR #636, opened 2024-07-25
Fix TestPodSidecarLabelChangeTriggersAddIptablesAction flake

hzxuzhonghu:
**What type of PR is this?**

/kind bug

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/bypass/bypass_test.go
+++ b/pkg/controller/bypass/bypass_test.go
@@ -25,24 +25,39 @@ import (
 
 	"github.com/agiledragon/gomonkey/v2"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"istio.io/api/annotation"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
 )
 
-func TestPodSidecarLabelChangeTriggersAddIptablesAction(t *testing.T) {
-	client := fake.NewSimpleClientset()
-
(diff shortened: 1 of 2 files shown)


pkg/controller/bypass/bypass_test.go
```diff
@@ -25,24 +25,39 @@ import (
...
+			Name: namespaceName,
+			Labels: map[string]string{
+				"istio-injection": "enabled",
+			},
+		},
+	}
+	client := fake.NewSimpleClientset(namespace)
 	c := NewByPassController(client)
 	go c.Run(stopCh)
+	cache.WaitForCacheSync(stopCh, c.pod.HasSynced)
```
hzxuzhonghu, line 59: This is the root cause

Okabe-Rintarou-0: you can directly call `c.Run(stopCh)`, just remove `go`

Okabe-Rintarou-0:
i also found that the NODE_NAME env is not aligned with the one in the test pod's `spec.nodeName`.
One is `test_node`, another is `test-node`, will it cause some problem?

hzxuzhonghu:
They should match. Weired, why the test passed

pkg/controller/bypass/bypass_test.go
hzxuzhonghu: Yes,  here we can

hzxuzhonghu:
Oh, it is because fake client does not support field selector, which actually works in kube-apiserver

hzxuzhonghu:
ping @supercharge-xsy

YaoZengzeng:
@hzxuzhonghu maybe this PR could be merged?

hzxuzhonghu:
@nlgwcy I thought it was merged

PR #640, opened 2024-07-26
kmesh restart with config change

lec-bit:
**What type of PR is this?**
/kind feature

Fixes #

```release-note

```


pkg/controller/workload/workload_processor.go
```diff
@@ -141,6 +142,16 @@ func (p *Processor) storePodFrontendData(uid uint32, ip []byte) error {
...
+	for _, uid := range removedResources {
+		p.WorkloadCache.DeleteWorkload(uid)
+		if err := p.removeWorkloadResourceByUid(uid); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (p *Processor) removeWorkloadResourceByUid(uid string) error {
```
hzxuzhonghu, line 154: suggest renaming to removeWorkloadFromBpfMap to distinguish with remove from local cache

pkg/controller/workload/workload_processor.go
```diff
@@ -150,66 +161,63 @@ func (p *Processor) removeWorkloadResource(removedResources []string) error {
...
-		p.WorkloadCache.DeleteWorkload(uid)
-		backendUid := p.hashName.StrToNum(uid)
-		// for Pod to Pod access, Pod info stored in frontend map, when Pod offline, we need delete the related records
-		if err = p.deletePodFrontendData(backendUid); err != nil {
-			log.Errorf("deletePodFrontendData failed: %s", err)
-			goto failed
-		}
+	backendUid := p.hashName.StrToNum(uid)
+	// for Pod to Pod access, Pod info stored in frontend map, when Pod offline, we need delete the related records
+	if err = p.deletePodFrontendData(backendUid); err != nil {
```
hzxuzhonghu, line 166: cc @nlgwcy Is fixing this, but it think any one can merge first

pkg/controller/workload/workload_processor.go
```diff
@@ -587,10 +603,46 @@ func (p *Processor) handleAddressTypeResponse(rsp *service_discovery_v3.DeltaDis
...
+// When processing the workload's response for the first time,
+// fetch the data from the /mnt/workload_hash_name.yaml file
+// and compare it with the data in the cache.
+func (p *Processor) compareWorkloadAndServiceWithHashName() {
+	var (
+		bk = bpf.BackendKey{}
+		bv = bpf.BackendValue{}
+	)
+
+	if kmeshbpf.GetStartType() == kmeshbpf.Normal {
```
hzxuzhonghu, line 620: Since we haven't support upgrade, to be safer shall we compare with `RESTART`  only?

pkg/controller/workload/workload_processor.go
```diff
@@ -587,10 +603,46 @@ func (p *Processor) handleAddressTypeResponse(rsp *service_discovery_v3.DeltaDis
...
+		return
+	}
+
+	log.Infof("reload workload config from last start")
+	kmeshbpf.SetStartType(kmeshbpf.Normal)
+	for str, num := range p.hashName.strToNum {
+		if p.WorkloadCache.GetWorkloadByUid(str) == nil && p.ServiceCache.GetService(str) == nil {
+			log.Debugf("GetWorkloadByUid and GetService nil:%v", str)
```
hzxuzhonghu, line 628: We cannot assume all the strs are service or pod, can we?

hzxuzhonghu:
cannot see how you make `compareWorkloadAndServiceWithHashName` only run once? And please add some cases covering different scenarios on changes during restart

hzxuzhonghu:
And you did not handle service delete

pkg/controller/workload/workload_processor.go
lec-bit: yes, I was careless here. I should add a check process in bpf_map.

lec-bit:
> cannot see how you make `compareWorkloadAndServiceWithHashName` only run once? And please add some cases covering different scenarios on changes during restart

When restarting, Status=Restart, and set it to Normal after execution to ensure that the function is only executed once.
And current use case is that there is a pod and a service in hashName, which are then processed in `compareWorkloadAndServiceWithHashName` function.

pkg/controller/workload/workload_processor.go
lec-bit: In fact, the same interface is called, I split it out and reused it

pkg/controller/workload/workload_processor.go
lec-bit: yes, i will change it

hzxuzhonghu:
@lec-bit Address the comments and we can merge

pkg/controller/workload/workload_processor.go
```diff
@@ -587,10 +603,52 @@ func (p *Processor) handleAddressTypeResponse(rsp *service_discovery_v3.DeltaDis
...
+		bv = bpf.BackendValue{}
+		sk = bpf.ServiceKey{}
+		sv = bpf.ServiceValue{}
+	)
+
+	if kmeshbpf.GetStartType() != kmeshbpf.Restart {
+		return
+	}
+
+	log.Infof("reload workload config from last start")
```
hzxuzhonghu, line 626:
```suggestion
	log.Infof("reload workload config from last epoch")
```

pkg/controller/workload/workload_processor.go
```diff
@@ -587,10 +603,52 @@ func (p *Processor) handleAddressTypeResponse(rsp *service_discovery_v3.DeltaDis
...
+	kmeshbpf.SetStartType(kmeshbpf.Normal)
+
+	// The record exists in the hashName file, exists in Backend or Service bpfmap,
+	// and does not exist in cache.
```
hzxuzhonghu, line 630: Confusing

hzxuzhonghu: I guess you mean when the str exists in hash file:
1. if not in the userspace cache, we should delete it from bpf

hzxuzhonghu:
/lgtm
/approve

Thanks, this is a great improvement.

Issue #669, opened 2024-08-02
User mode log error


weli-l:
**What happened**:
When I locate #628 this issue. I found that the log printed in user mode is 0. But actually when I use bpf_trace_printk to print logs, the error code is -17, not 0
**What you expected to happen**:
Print out the actual error code
**How to reproduce it (as minimally and precisely as possible)**:
follow #628 
**Anything else we need to know?**:

**Environment**:
- Kmesh version:
- Others:

weli-l:
ping @bfforever ，pls confirm if this is the problem

PR #673, opened 2024-08-02
Update bpf log

hzxuzhonghu:
**What type of PR is this?**

bpf_ringbuffer_output need extra memory copy  FYI https://arthurchiao.art/blog/bpf-ringbuf-zh/

Fixes #

```release-note

```

### DIFF
--- a/bpf/include/bpf_log.h
+++ b/bpf/include/bpf_log.h
@@ -81,15 +81,14 @@ lower than 22.09, compile would report an error of bpf_snprintf dont exist */
     ({                                                                                                                 \
         struct log_event *e;                                                                                           \
         __u32 ret = 0;                                                                                                 \
-        int zero = 0;                                                                                                  \
-        e = bpf_map_lookup_elem(&tmp_log_buf, &zero);                                                                  \
+        e = bpf_ringbuf_reserve(&kmesh_events, sizeof(struct log_event), 0);                                           \
         if (!e)                                                                                                        \
             break;                                                                                                     \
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
https://docs.kernel.org/6.6/bpf/ringbuf.html

```c
bpf_ringbuf_output() has disadvantage of incurring extra memory copy, because record has to be prepared in some other place first. But it allows to submit records of the length that's not known to verifier beforehand. It also closely matches bpf_perf_event_output(), so will simplify migration significantly.

bpf_ringbuf_reserve() avoids the extra copy of memory by providing a memory pointer directly to ring buffer memory. In a lot of cases records are larger than BPF stack space allows, so many programs have use extra per-CPU array as a temporary heap for preparing sample. bpf_ringbuf_reserve() avoid this needs completely. But in exchange, it only allows a known constant size of memory to be reserved, such that verifier can verify that BPF program can't access memory outside its reserved record space. bpf_ringbuf_output(), while slightly slower due to extra memory copy, covers some use cases that are not suitable for bpf_ringbuf_reserve().
```

supercharge-xsy:
maybe this issue can be resolved by this pr ? https://github.com/kmesh-net/kmesh/issues/669

hzxuzhonghu:
I am not sure, will test it and report back

hzxuzhonghu:
>maybe this issue can be resolved by this pr ? https://github.com/kmesh-net/kmesh/issues/669

This doesnot change anything to [reported issue](https://github.com/kmesh-net/kmesh/issues/669)

PR #696, opened 2024-08-06
fix bug that may occur in Update mode

lec-bit:
**What type of PR is this?**
/kind bug

The current startup strategy of kmesh does not support Update scenarios, so there will be bugs when starting in Update mode. Now the Update startup mode will be forcibly changed to Normal mode, and the Update startup mode will be supported in the future.
Fixes #

```release-note

```

### DIFF
--- a/pkg/bpf/bpf.go
+++ b/pkg/bpf/bpf.go
@@ -220,15 +220,20 @@ func NewVersionMap(config *options.BpfConfig) *ebpf.Map {
 		}
 	}
 
-	if GetStartType() == Restart {
+	switch GetStartType() {
+	case Restart:
 		return versionMap
+	case Update:
+		// TODO : update mode has not been fully developed and is currently consistent with normal mode
+		log.Warnf("Update mode support is under development, Will be started in Normal mode.")
+	default:
 	}
 
 	// Make sure the directory about to use is clean
-	kmeshBpfPath := filepath.Dir(filepath.Dir(versionPath))
+	kmeshBpfPath := filepath.Dir(versionPath)
 	err = os.RemoveAll(kmeshBpfPath)
 	if err != nil {
-		log.Errorf("Clean bpf Path failed, err is:%v", err)
+		log.Errorf("Clean bpf maps and progs failed, err is:%v", err)
 		return nil
 	}
 
--- a/pkg/bpf/bpf.go
+++ b/pkg/bpf/bpf.go
@@ -33,6 +33,7 @@ import (
 	"github.com/cilium/ebpf/rlimit"
 
 	"kmesh.net/kmesh/daemon/options"
+	"kmesh.net/kmesh/pkg/constants"
 	"kmesh.net/kmesh/pkg/logger"
 	"kmesh.net/kmesh/pkg/version"
 )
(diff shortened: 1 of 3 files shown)


pkg/bpf/bpf.go
```diff
@@ -204,21 +204,33 @@ func (l *BpfLoader) Stop() {
...
+		versionPath = filepath.Join(config.BpfFsPath + "/bpf_kmesh_workload/map/kmesh_version")
 	}
 
 	_, err := os.Stat(versionPath)
 	if err == nil {
-		m := recoverVersionMap(config, versionPath)
-		if m != nil {
-			SetStartStatus(m)
-			return m
+		versionMap = recoverVersionMap(config, versionPath)
```
LiZhenCheng9527, line 216: Don't you handle errors when they occur?

pkg/bpf/bpf.go
```diff
@@ -204,21 +204,33 @@ func (l *BpfLoader) Stop() {
 
 func NewVersionMap(config *options.BpfConfig) *ebpf.Map {
 	var versionPath string
+	var versionMap *ebpf.Map
 	if config.AdsEnabled() {
-		versionPath = filepath.Join(config.BpfFsPath + "/bpf_kmesh/map/")
+		versionPath = filepath.Join(config.BpfFsPath + "/bpf_kmesh/map/kmesh_version")
```
hzxuzhonghu, line 209: since you are touching these hardcoded string, i would suggest defining some constants for them

pkg/bpf/bpf.go
```diff
@@ -204,21 +204,33 @@ func (l *BpfLoader) Stop() {
...
 
+	if GetStartType() == Restart {
+		return versionMap
+	}
+
+	// Make sure the directory about to use is clean
+	kmeshBpfPath := filepath.Dir(filepath.Dir(versionPath))
+	err = os.RemoveAll(kmeshBpfPath)
+	if err != nil {
+		log.Errorf("Clean bpf Path failed, err is:%v", err)
```
hzxuzhonghu, line 230: make the log clearer

pkg/bpf/bpf_restart.go
```diff
@@ -96,6 +96,8 @@ func SetStartStatus(versionMap *ebpf.Map) {
 		SetStartType(Restart)
 	} else {
 		log.Infof("kmesh start with Update")
+		// TODO : update mode has not been fully developed and is currently consistent with normal mode
+		log.Infof("Update mode support is under development, Will be started in Normal mode.")
```
YaoZengzeng, line 100: It's more appropriate to adjust the log level to warn

pkg/bpf/bpf.go
lec-bit: Only the successful recovery from the pin directory needs to be processed. In other cases, it will be started in the Normal state.

pkg/bpf/bpf_restart.go
hzxuzhonghu: I would prefer hidding this

pkg/bpf/bpf.go
```diff
@@ -204,21 +205,38 @@ func (l *BpfLoader) Stop() {
 
 func NewVersionMap(config *options.BpfConfig) *ebpf.Map {
 	var versionPath string
+	var versionMap *ebpf.Map
 	if config.AdsEnabled() {
-		versionPath = filepath.Join(config.BpfFsPath + "/bpf_kmesh/map/")
+		versionPath = filepath.Join(config.BpfFsPath + constants.VersionPath)
```
hzxuzhonghu, line 210: We can up[date later
```suggestion
		versionPath = filepath.Join(config.BpfFsPath, constants.VersionPath)
```

hzxuzhonghu (approved):
/lgtm

PR #707, opened 2024-08-08
FIX: go test timeout after running benchmark test

hzxuzhonghu:
…timeout

**What type of PR is this?**

Without cleaning up hash file, test will fail with timeout error.  By default vs plugin set 30s timeout.

Fixes #

```release-note

```


hzxuzhonghu:
@nlgwcy @supercharge-xsy Can you please take a look

PR #709, opened 2024-08-09
Support hostname network

### LINKED ISSUE #631: We canot assume workload ip address are unique
Since istio is supporting accessing service with hostnetwork pods. It can build workload with hostnetwork with same ip now.  In kmesh we need to adapt to hostnetwork pods.

I haven't looked into what will happend if we kmesh do no change.  Want to hear more voices first.

Okabe-Rintarou-0:
**What type of PR is this?**
feature

Fixes #631 

```release-note

```

### DIFF
--- a/pkg/controller/workload/cache/workload_cache.go
+++ b/pkg/controller/workload/cache/workload_cache.go
@@ -93,10 +93,15 @@ func (w *cache) AddWorkload(workload *workloadapi.Workload) {
 	}
 
 	w.byUid[uid] = workload
-	for _, ip := range workload.Addresses {
-		addr, _ := netip.AddrFromSlice(ip)
-		networkAddress := composeNetworkAddress(workload.Network, addr)
-		w.byAddr[networkAddress] = workload
+
+	// We should exclude the workloads that use host network mode
+	// Since they are using the host ip, we can not use address to identify them
+	if workload.NetworkMode != workloadapi.NetworkMode_HOST_NETWORK {
+		for _, ip := range workload.Addresses {
+			addr, _ := netip.AddrFromSlice(ip)
+			networkAddress := composeNetworkAddress(workload.Network, addr)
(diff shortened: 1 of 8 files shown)


pkg/controller/workload/cache/workload_cache.go
```diff
@@ -93,10 +93,15 @@ func (w *cache) AddWorkload(workload *workloadapi.Workload) {
...
-		networkAddress := composeNetworkAddress(workload.Network, addr)
-		w.byAddr[networkAddress] = workload
+
+	// We should exclude the workloads that use host network mode
+	// Since they are using the host ip, we can not use address to identify them
+	if workload.NetworkMode != workloadapi.NetworkMode_HOST_NETWORK {
+		for _, ip := range workload.Addresses {
+			addr, _ := netip.AddrFromSlice(ip)
+			networkAddress := composeNetworkAddress(workload.Network, addr)
+			w.byAddr[networkAddress] = workload
```
hzxuzhonghu, line 103: This may have some influence on telemetry, but since we cannot distinguish workload by only host ip. So this make sense to me

pkg/controller/workload/workload_processor.go
```diff
@@ -427,9 +431,12 @@ func (p *Processor) handleDataWithoutService(workload *workloadapi.Workload) err
 			return err
 		}
 
-		if err = p.storePodFrontendData(uid, ip); err != nil {
-			log.Errorf("storePodFrontendData failed, err:%s", err)
-			return err
+		// we should not store frontend data of hostname network mode pods
```
hzxuzhonghu, line 434: add: because host network pod cannot be managed by kmesh

hzxuzhonghu (commented):
LGTM, please add some ut cases

Okabe-Rintarou-0:
ok

hzxuzhonghu (approved):
/lgtm

PR #723, opened 2024-08-13
Improve cni uninstall

hzxuzhonghu:
**What type of PR is this?**

Fix option 1 of https://github.com/kmesh-net/kmesh/issues/730

> Do clean up the cni config at the end of https://github.com/kmesh-net/kmesh/blob/main/build/docker/start_kmesh.sh#L54

This may cleanup cni configs twice, but doesnot matter.

If without this, and kmesh crashed unexpected, the k8s cluster will be influenced, the blast radius is very large

Fixes #

```release-note

```

### DIFF
--- a/daemon/manager/manager.go
+++ b/daemon/manager/manager.go
@@ -80,11 +80,15 @@ func Execute(configs *options.BootstrapConfigs) error {
 	}
 
 	bpfLoader := bpf.NewBpfLoader(configs.BpfConfig)
+<<<<<<< HEAD
 	if err = bpfLoader.Start(configs.BpfConfig); err != nil {
+=======
+	defer bpfLoader.Stop()
+	if err := bpfLoader.Start(configs.BpfConfig); err != nil {
+>>>>>>> e6e5bf1 (Improve cni uninstall)
 		return err
 	}
-	log.Info("bpf Start successful")
-	defer bpfLoader.Stop()
+	log.Info("bpf loader start successfully")
 
 	stopCh := make(chan struct{})
 	defer close(stopCh)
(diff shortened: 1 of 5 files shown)


pkg/cni/install.go
```diff
@@ -70,17 +70,22 @@ func NewInstaller(mode string,
 
 func (i *Installer) Start() error {
 	if i.Mode == constants.AdsMode || i.Mode == constants.WorkloadMode {
```
YaoZengzeng, line 72: Are there other modes?

pkg/cni/install.go
```diff
@@ -70,17 +70,22 @@ func NewInstaller(mode string,
...
+		if err != nil {
+			i.Stop()
+			return err
+		}
 	}
 	return nil
 }
 
 func (i *Installer) Stop() {
 	if i.Mode == constants.AdsMode || i.Mode == constants.WorkloadMode {
```
YaoZengzeng, line 84: ditto

daemon/manager/manager.go
```diff
@@ -67,6 +68,7 @@ func NewCommand() *cobra.Command {
 	cmd.AddCommand(version.NewCmd())
 	cmd.AddCommand(dump.NewCmd())
 	cmd.AddCommand(logcmd.NewCmd())
+	cmd.AddCommand(uninstall.NewUninstallCmd())
```
YaoZengzeng, line 71:
```suggestion
	cmd.AddCommand(uninstall.NewCmd())
```

Be consistent with other subcommands and the pkg name already explains the purpose.

pkg/cni/install.go
hzxuzhonghu: can not specify any mode, like if we want to enable bypass

pkg/cni/install.go
```diff
@@ -70,17 +70,22 @@ func NewInstaller(mode string,
 
 func (i *Installer) Start() error {
 	if i.Mode == constants.AdsMode || i.Mode == constants.WorkloadMode {
-		log.Info("start write CNI config\n")
-		return i.addCniConfig()
+		log.Info("start write CNI config")
+		err := i.addCniConfig()
+		if err != nil {
+			i.Stop()
```
bitcoffeeiux, line 76: Is this the cause of cni remove twice?

nlgwcy:
@supercharge-xsy

PR #724, opened 2024-08-13
Increased stability of metric asynchronous execution

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Fixes #

```release-note

```


pkg/controller/telemetry/metric.go
```diff
@@ -163,37 +171,50 @@ func (m *MetricController) Run(ctx context.Context, mapOfTcpInfo *ebpf.Map) {
...
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		default:
+			rec := ringbuf.Record{}
 			if err := reader.ReadInto(&rec); err != nil {
 				log.Errorf("ringbuf reader FAILED to read, err: %v", err)
-				continue
+				return
```
YaoZengzeng, line 182: why did it change from `continue` to `return`? 

`getConnectDataFromRingBuf` run as goroutine and don't handle errors. Event if it fails, only the goroutine exits and the rest of the system is completely unaware. 

We need a fault-tolerant mechanism or an overall exit after failure

pkg/controller/telemetry/metric.go
```diff
@@ -163,37 +171,50 @@ func (m *MetricController) Run(ctx context.Context, mapOfTcpInfo *ebpf.Map) {
...
 			default:
-				log.Errorf("get connection info failed: %v", err)
-				continue
+				log.Errorf("Unsupported IP family")
+				return
+			}
+
+			if err != nil {
+				log.Debugf("get data failed: %v", err)
+				return
```
YaoZengzeng, line 217: ditto

pkg/controller/telemetry/metric.go
```diff
@@ -297,7 +318,7 @@ func (m *MetricController) getWorkloadByAddress(address []byte) (*workloadapi.Wo
 	networkAddr.Address, _ = netip.AddrFromSlice(address)
 	workload := m.workloadCache.GetWorkloadByAddr(networkAddr)
 	if workload == nil {
-		log.Warnf("get workload from ip %v FAILED", address)
+		log.Debugf("get workload from ip %v FAILED", address)
```
YaoZengzeng, line 321: Why did it change from `Warnf` to `Debugf`? Does this happen frequently? And is it normal that workload can't be found from IP?

hzxuzhonghu (commented):
We still need this async processing, there are crirical section protected during metric building

pkg/controller/telemetry/metric.go
```diff
@@ -350,14 +371,12 @@ func buildServiceMetric(dstWorkload, srcWorkload *workloadapi.Workload, dstPort
 			}
 		}
 		if namespacedhost == "" {
-			log.Infof("can't find service correspond workload: %s", dstWorkload.Name)
+			log.Debugf("can't find service correspond workload: %s", dstWorkload.Name)
```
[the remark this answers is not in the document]
LiZhenCheng9527: This error message has been removed from the main branch.

pkg/controller/telemetry/metric.go
LiZhenCheng9527: i think this is not error

LiZhenCheng9527:
That's no longer necessary.

PR #726, opened 2024-08-14
Rm coverage threshold

hzxuzhonghu:
With the hard limit, may even prevent developer from fixing typo, which is not covered before.


hzxuzhonghu:
@LiZhenCheng9527

LiZhenCheng9527:
/lgtm
There's nothing wrong with removing the threshold.

PR #731, opened 2024-08-14
make bypass idempotent

### LINKED ISSUE #637: Make bypass idempotent and can startup alone

Kmesh is rely on pod informer to get enable/disable bypass sidecar. As we know the k8s informer can drop some events if the client is very slow receive and process the pod event. 

In order to better process bypass, we should make iptables add/delete idempotent.  

And to separate from ads and workload mode, it should be run separately. 

/assign @weli-l

weli-l:
**What type of PR is this?**

Fixes #637

```release-note

```


pkg/utils/informer.go
hzxuzhonghu, line 1: shall we move to pkg/kube package

pkg/controller/bypass/bypass_controller.go
```diff
@@ -153,6 +146,14 @@ func addIptables(ns string) error {
 
 	execFunc := func(netns.NetNS) error {
 		log.Infof("Running add iptables rule in namespace:%s", ns)
+		//To avoid iptables rules being added multiple times due to problems with k8s resource synchronization
+		delIptArgs := [][]string{
+			{"-t", "nat", "-D", "PREROUTING", "-j", "RETURN"},
```
hzxuzhonghu, line 151: Is this safe if the no such rule exist

pkg/utils/informer.go
weli-l: > shall we move to pkg/kube package

done

pkg/controller/bypass/bypass_controller.go
weli-l: > Is this safe if the no such rule exist

yes, when this rule is not present, an error will occur, but it will not be thrown out in the log and will have no effect on the function, so you can just ignore it.

hzxuzhonghu (approved):
/lgtm

PR #739, opened 2024-08-15
Modify observe connection shutdown probe log

jiayoukun:
For the bpf_sk_storage_get function, for an sk that may have been closed to get the storage associated with it, it usually returns NULL, and this part of the judgement log level should be changed to INFO, and the subsequent `tcp_report(sk, tcp_sock, storage, BPF_TCP_ CLOSE)` also reports on this connection closure.

### DIFF
--- a/bpf/kmesh/probes/probe.h
+++ b/bpf/kmesh/probes/probe.h
@@ -61,8 +61,7 @@ static inline void observe_on_close(struct bpf_sock *sk)
 
     storage = bpf_sk_storage_get(&map_of_sock_storage, sk, 0, 0);
     if (!storage) {
-        BPF_LOG(ERR, PROBE, "close bpf_sk_storage_get failed\n");
-        return;
+        BPF_LOG(INFO, PROBE, "TCP Connection closed or map_of_sock_storage cleaned up\n");
     }
 
     tcp_report(sk, tcp_sock, storage, BPF_TCP_CLOSE);


jiayoukun:
@LiZhenCheng9527 @nlgwcy

bpf/kmesh/probes/probe.h
```diff
@@ -61,10 +61,9 @@ static inline void observe_on_close(struct bpf_sock *sk)
 
     storage = bpf_sk_storage_get(&map_of_sock_storage, sk, 0, 0);
     if (!storage) {
-        BPF_LOG(ERR, PROBE, "close bpf_sk_storage_get failed\n");
-        return;
+        BPF_LOG(INFO, PROBE, "TCP Connection closed or map_of_sock_storage cleaned up\n");
```
LiZhenCheng9527, line 64: Are you able to give an actual example?
Because storage is created when the link is established, this is handled when the link is closed. If the link is closed before this processing, then the whole processing logic is problematic

jiayoukun: > Are you able to give an actual example? Because storage is created when the link is established, this is handled when the link is closed. If the link is closed before this processing, then the whole processing logic is problematic

The ‘bpf_sk_storage_get ’ function is a helper function for getting bpf-local-storage from a socket (sk). It is similar to getting a value from a map with sk as the key, but bpf-local-storage is actually stored locally on the sk, not on the map.
I read in [https://docs.kernel.org/bpf/map_sk_storage.html](https://docs.kernel.org/bpf/map_sk_storage.html) that the ‘BPF_MAP_TYPE_SK_STORAGE’ type of map is recycled when the socket is deleted, and in the case that occurs here it should mostly return NULL, and the logs will print the err logs repeatedly.

jiayoukun:
@LiZhenCheng9527 like this
87CF2CB0-A8D9-4039-8F11-82E9726DC327

bpf/kmesh/probes/probe.h
```diff
@@ -61,10 +61,38 @@ static inline void observe_on_close(struct bpf_sock *sk)
...
     storage = bpf_sk_storage_get(&map_of_sock_storage, sk, 0, 0);
     if (!storage) {
-        BPF_LOG(ERR, PROBE, "close bpf_sk_storage_get failed\n");
+        __u32 src_ip4 = sk->src_ip4;
+        __u32 dst_ip4 = sk->dst_ip4;
+        __u32 src_port = sk->src_port;
+        __u32 dst_port = (__u32)bpf_ntohs(sk->dst_port);
+
+        if (sk->family == AF_INET) {
+            // IPv4 连接
```
YaoZengzeng, line 70: It would be better to use English to comment :)

bpf/kmesh/probes/probe.h
```diff
@@ -61,10 +61,38 @@ static inline void observe_on_close(struct bpf_sock *sk)
...
+            BPF_LOG(INFO, PROBE, "TCP connection closed, src_ip=%u.%u.%u.%u, src_port=%u, dst_ip=%u.%u.%u.%u, dst_port=%u\n",
+                    (src_ip4 >> 24) & 0xFF, (src_ip4 >> 16) & 0xFF, (src_ip4 >> 8) & 0xFF, src_ip4 & 0xFF,
+                    src_port,
+                    (dst_ip4 >> 24) & 0xFF, (dst_ip4 >> 16) & 0xFF, (dst_ip4 >> 8) & 0xFF, dst_ip4 & 0xFF,
+                    dst_port);
+        } else if (sk->family == AF_INET6) {
+            // IPv6 连接
```
YaoZengzeng, line 77: ditto

jiayoukun: thanks.

bpf/kmesh/probes/probe.h
```diff
@@ -61,10 +61,38 @@ static inline void observe_on_close(struct bpf_sock *sk)
 
     storage = bpf_sk_storage_get(&map_of_sock_storage, sk, 0, 0);
     if (!storage) {
-        BPF_LOG(ERR, PROBE, "close bpf_sk_storage_get failed\n");
+        __u32 src_ip4 = sk->src_ip4;
```
LiZhenCheng9527, line 64: Previously it was found that the src and dst of inbound traffic is reversed with outbound traffic, don't we need to consider this over here?

LiZhenCheng9527:
> @LiZhenCheng9527 like this 87CF2CB0-A8D9-4039-8F11-82E9726DC327

thanks

bpf/kmesh/probes/probe.h
```diff
@@ -61,10 +61,38 @@ static inline void observe_on_close(struct bpf_sock *sk)
...
+        __u32 dst_ip4 = sk->dst_ip4;
+        __u32 src_port = sk->src_port;
+        __u32 dst_port = (__u32)bpf_ntohs(sk->dst_port);
+
+        if (sk->family == AF_INET) {
+            // IPv4 connection
+            BPF_LOG(INFO, PROBE, "TCP connection closed, src_ip=%u.%u.%u.%u, src_port=%u, dst_ip=%u.%u.%u.%u, dst_port=%u\n",
```
hzxuzhonghu, line 71: donot we already have a way to print ip address

hzxuzhonghu: ip2str

PR #741, opened 2024-08-16
separation

lec-bit:
**What type of PR is this?**
/kind feature

Separate control plane and data plane compilation

1. Compile the kmesh-devel software package, including some compilation dependencies
2. Split the Makefile and execute make kmesh-controller to compile the complete kmesh-daemon using the files provided by the kmesh-devel package.

Fixes #

```release-note

```


Makefile
```diff
@@ -73,22 +73,26 @@ TMP_FILES := bpf/kmesh/bpf2go/bpf2go.go \
 	bpf/kmesh/ads/include/config.h \
 	bpf/include/bpf_helper_defs_ext.h
 
-.PHONY: all install uninstall clean build docker
+.PHONY: all install uninstall clean build docker dataplane controller
 
-all:
-	$(QUIET) find $(ROOT_DIR)/mk -name "*.pc" | xargs sed -i "s#^prefix=.*#prefix=${ROOT_DIR}#g"
+all: dataplane controller
```
YaoZengzeng, line 78: What are the benefits of doing so? Compile the control plane go program separately without having some dependencies on the kernel, I guess?

Also we should also update dev doc, tell users that they can do this.

LiZhenCheng9527: This is supposed to make it easier to test without having to compile a daemon.

lec-bit: This is a user requirement. They need the kernel capabilities related to the ebpf of the kmesh. Then, they make some customized modifications on the control plane. Therefore, the compilation is split.

PR #743, opened 2024-08-16
passthrough go test flag

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

The E2E test will eventually call the `go test` command, which has a variety of flags for adjusting the test.

Obviously, it's very tedious to encapsulate each flag. Therefore we adopt the direct passthrough unknown flags to make full use of existing capabilities of `go test`.

Although it's not very robust, I think it's ok for running test cases.

For example, now we could directly run the follow command:

```bash
./test/e2e/run_test.sh --only-run-tests -run "TestKmeshRestart"  -count 10 -timeout 30m
``` 

Fixes #

```release-note

```


YaoZengzeng:
@hzxuzhonghu PTAL

hzxuzhonghu (approved):

PR #776, opened 2024-08-27
Check codecov but not fail github checks

hzxuzhonghu:
**What type of PR is this?**

Just want codcov to give use a hint, not to block PR merging. Not sure if this works.

Also investigated on prow bot to ignore the codecov, but not found how to congure it

Fixes #

```release-note

```

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,11 +1,12 @@
 coverage:
   status:
-    project: #add everything under here, more options at https://docs.codecov.com/docs/commit-status
+    project: off
+    # https://docs.codecov.com/docs/commit-status#patch-status
+    patch:
       default:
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -7,4 +7,5 @@ coverage:
         base: auto 
 comment:                  # this is a top-level key
   layout: "header, files, footer" # remove "new" from "header" and "footer"
-  hide_project_coverage: false # set to false
\ No newline at end of file
+  hide_project_coverage: false # set to false
```
LiZhenCheng9527, line 10: can refer to https://github.com/kubeedge/kubeedge/blob/master/.github/codecov.yml

PR #790, opened 2024-08-29
e2e test to include bookinfo as test application

### LINKED ISSUE #766: [E2E Enhancement] include bookinfo as test appliaction

Include [bookinfo](https://istio.io/latest/docs/examples/bookinfo/) as test application in the e2e test set to cover complex application scenarios with multiple language applications and multiple hops. In this way, problem similar to #756 could be convered.

**Why is this needed**:

The test application in the e2e framework is just a multi-instance deployment of a single go application, which acts as both a client and a server, and has only a single-hop access. When doing manual testing, we often use bookinfo, but it often happens that all e2e test cases pass and the bookinfo test fails.

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

Fixes #766

```release-note

```

### DIFF
--- a/test/e2e/baseline_test.go
+++ b/test/e2e/baseline_test.go
@@ -626,11 +626,11 @@ func TestBookinfo(t *testing.T) {
 		})
 
 		// Install sleep as client.
-		if _, err := shell.Execute(true, fmt.Sprintf("kubectl apply -f https://raw.githubusercontent.com/istio/istio/release-1.21/samples/sleep/sleep.yaml -n %s", namespace)); err != nil {
+		if _, err := shell.Execute(true, fmt.Sprintf("kubectl apply -f https://raw.githubusercontent.com/istio/istio/release-1.22/samples/sleep/sleep.yaml -n %s", namespace)); err != nil {
 			t.Fatalf("failed to install sleep as client of bookinfo: %v", err)
 		}
 		t.Cleanup(func() {
-			if _, err := shell.Execute(true, fmt.Sprintf("kubectl delete -f https://raw.githubusercontent.com/istio/istio/release-1.21/samples/sleep/sleep.yaml -n %s", namespace)); err != nil {
+			if _, err := shell.Execute(true, fmt.Sprintf("kubectl delete -f https://raw.githubusercontent.com/istio/istio/release-1.22/samples/sleep/sleep.yaml -n %s", namespace)); err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/baseline_test.go
```diff
@@ -610,6 +612,77 @@ func TestMixNsAndServiceWaypoint(t *testing.T) {
...
 
+func TestBookinfo(t *testing.T) {
+	framework.NewTest(t).Run(func(t framework.TestContext) {
+		namespace := apps.Namespace.Name()
+		// Install bookinfo.
+		if _, err := shell.Execute(true, fmt.Sprintf("kubectl apply -f https://raw.githubusercontent.com/istio/istio/release-1.22/samples/bookinfo/platform/kube/bookinfo.yaml -n %s", namespace)); err != nil {
+			t.Fatalf("failed to install bookinfo: %v", err)
+		}
+		t.Cleanup(func() {
+			if _, err := shell.Execute(true, fmt.Sprintf("kubectl delete -f https://raw.githubusercontent.com/istio/istio/release-1.22/samples/bookinfo/platform/kube/bookinfo.yaml -n %s", namespace)); err != nil {
```
LiZhenCheng9527, line 623: Does this need to change with istio version changes

test/e2e/baseline_test.go
```diff
@@ -610,6 +612,77 @@ func TestMixNsAndServiceWaypoint(t *testing.T) {
...
+				t.Fatalf("failed to delete bookinfo: %v", err)
+			}
+		})
+
+		// Install sleep as client.
+		if _, err := shell.Execute(true, fmt.Sprintf("kubectl apply -f https://raw.githubusercontent.com/istio/istio/release-1.21/samples/sleep/sleep.yaml -n %s", namespace)); err != nil {
```
LiZhenCheng9527, line 629: Why is it release 1.22 above and release 1.21 here?

test/e2e/baseline_test.go
YaoZengzeng: > Does this need to change with istio version changes

It should not be needed at present. We only use very basic functions here and bookinfo itself will not change much with upgrade of istio.

hzxuzhonghu (approved):
This is very hack,  we can use call options provided by istio


But i wish this test to cover ipv4 mapped v6 asap, so

/lgtm

PR #791, opened 2024-08-29
istio 1.23 for e2e

### LINKED ISSUE #782: [E2E Enhancement] support multiple versions of istio

E2E tests could be run on multiple versions of istio, and some cases only run on specific istio version, such as some newly added features like #697 

**Why is this needed**:

Currently E2E test cases will only run on istio 1.22, but as subsequent istio versions are released and new features are added, we will hope to add new E2E test cases for new features while ensuring compatibilities with older version.

**What needs to be discussed is how many istio versions should we support and how long to retire the oldest version?**

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

Fixes #782

```release-note

```

### DIFF
--- a/.github/workflows/e2e-ipv6-istio-1.23.yml
+++ b/.github/workflows/e2e-ipv6-istio-1.23.yml
@@ -1,4 +1,4 @@
-name: E2E IPv6 Test/istio-1.23
+name: E2E IPv6 Test(istio 1.23)
 on: 
   pull_request:
 jobs:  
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e-ipv6-istio-1.23.yml
```diff
@@ -0,0 +1,22 @@
+name: E2E IPv6 Test/istio-1.23
```
hzxuzhonghu, line 1:
```suggestion
name: E2E IPv6 Test(istio 1.23)
```

hzxuzhonghu:
Yeah, much clear now

`
E2E IPv6 Test(istio 1.23) / E2E IPv6 Test (1.22) (pull_request) ` not sure what is the last 1.22 mean, could be go version. We may hide it

PR #796, opened 2024-08-30
some cleanups and rename on start type

hzxuzhonghu:
**What type of PR is this?**

/kind cleanup

Fixes #

**Special notes for you're reviewer**:

```release-note

```

### DIFF
--- a/pkg/bpf/bpf_restart.go
+++ b/pkg/bpf/bpf_restart.go
@@ -52,7 +52,10 @@ const (
 	Update
 )
 
+// kmeshStartType is used during bootstrap to indicate how to initiate bpf prog and maps
 var kmeshStartType StartType
+
+// kmeshStartType is used during exit to indicate how to cleanup bpf prog and maps
 var kmeshExitType StartType
 
 func GetStartType() StartType {
--- a/pkg/bpf/bpf_restart.go
+++ b/pkg/bpf/bpf_restart.go
@@ -38,10 +38,10 @@ import (
  *		Normal: a normal new start
  *		Restart: reusing the previous kmesh configuration
  *		Update: upgrading kmesh and reusing part of previous kmesh configuration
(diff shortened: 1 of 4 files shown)


pkg/bpf/bpf_restart.go
```diff
@@ -53,6 +53,7 @@ const (
 )
 
 var kmeshStartType StartType
+var kmeshExitType StartType
```
YaoZengzeng, line 56: Add comments about when to use `kmeshExitType` and when to use `kmeshStartType`, it is also very confusing that they are both of type `StartType`, maybe we need better naming.

pkg/bpf/bpf_restart_test.go
```diff
@@ -89,9 +89,11 @@ func runTestRestart(t *testing.T) {
 		assert.ErrorIsf(t, err, nil, "bpfLoader start failed %v", err)
 	}
 	assert.Equal(t, Normal, GetStartType(), "set kmesh start status:Normal failed")
-	SetStartType(Restart)
+
+	// Test Kmesh Restart
+	kmeshExitType = Restart
 	bpfLoader.Stop()
-	assert.Equal(t, Restart, GetStartType(), "set kmesh close status:Restart failed")
+	assert.Equal(t, Restart, GetExitType(), "set kmesh close status:Restart failed")
```
YaoZengzeng, line 96:
```suggestion
	assert.Equal(t, Restart, GetExitType(), "set kmesh exit type to Restart failed")
```

pkg/bpf/bpf_restart.go
hzxuzhonghu: kmeshStartType is used during bootsttap

kmeshExitType  is used during exit, to indicate whether keep bpf prog/maps or not

Previously one var is used to indicate both, confusing to me

PR #800, opened 2024-08-31
fix verifier log omitted

weli-l:
**What type of PR is this?**
fix verifier log omitted

Fixes #

```release-note

```

### DIFF
--- a/pkg/bpf/bpf.go
+++ b/pkg/bpf/bpf.go
@@ -21,6 +21,7 @@ package bpf
 // #include "cluster/cluster.pb-c.h"
 import "C"
 import (
+	"errors"
 	"fmt"
 	"hash/fnv"
 	"os"
@@ -68,6 +69,7 @@ func NewBpfLoader(config *options.BpfConfig) *BpfLoader {
 }
 
 func (l *BpfLoader) StartAdsMode() (err error) {
+	var ve *ebpf.VerifierError
 	if l.obj, err = NewBpfKmesh(l.config); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/bpf/bpf.go
```diff
@@ -79,7 +81,11 @@ func (l *BpfLoader) StartAdsMode() (err error) {
 	}()
 
 	if err = l.obj.Load(); err != nil {
-		return fmt.Errorf("bpf Load failed, %s", err)
+		l.Stop()
```
hzxuzhonghu, line 84: this is not needed there is a defer above

pkg/bpf/bpf_kmesh_l4_workload.go
```diff
@@ -54,14 +57,24 @@ func newWorkloadBpf(cfg *options.BpfConfig) (*BpfKmeshWorkload, error) {
...
 
-	if err = l.workloadObj.Load(); err != nil {
+	defer func() {
+		if err != nil {
+			l.Stop()
+		}
+	}()
+
+	if err = l.obj.Load(); err != nil {
 		l.Stop()
```
hzxuzhonghu, line 73: remove

hzxuzhonghu (approved):
/lgtm

pkg/bpf/bpf.go
```diff
@@ -68,6 +69,7 @@ func NewBpfLoader(config *options.BpfConfig) *BpfLoader {
 }
 
 func (l *BpfLoader) StartAdsMode() (err error) {
+	var ve *ebpf.VerifierError
```
hzxuzhonghu, line 72: This should be var ve ebpf.VerifierError, otherwise the below &ve will panic

hzxuzhonghu (changes requested):
Missed there is a bug

pkg/bpf/bpf.go
weli-l: > This should be var ve ebpf.VerifierError, otherwise the below &ve will panic

The function we finally call is 
```shell
func ErrorWithLog(source string, err error, log []byte, truncated bool) *VerifierError {
       ... ...
}
```
So, this ve should be *ebpf.VerifierError, And actually when I change  *ebpf.VerifierError to  ebpf.VerifierError, it will panic

hzxuzhonghu (approved):
/lgtm

Issue #730, opened 2024-08-14
Improve CNI installer


hzxuzhonghu:
**What would you like to be added**:

Currently cni installer is within the kmesh daemon, however when the daemon panic unexpected, the cni does not have a chance to uninstall all the cni configs.

If the kubeconfig kmesh-cni using is not out of date, then after kmesh-daemon updated, it can start up successfully.
However if the kubeconfig's token is out dated,  the the above procedure would fail.

Several solutions come up to my mind:

1. Do clean up the cni config at the end of https://github.com/kmesh-net/kmesh/blob/main/build/docker/start_kmesh.sh#L54
2. Add a go routine in cni installer, watch the token file, update the kubeconfig when token modified. This would make sure the kubeconfig can not expire easily.
3. Make kmesh-cni `CmdAdd` never fail, even it failed, we return success. And depend on manage controller to manage workload eventually.

**Why is this needed**:

hzxuzhonghu:
cc @YaoZengzeng

PR #801, opened 2024-08-31
update kubeconfig when token was updated

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

Fixes part of #730 

```release-note

```

### DIFF
--- a/daemon/options/cni.go
+++ b/daemon/options/cni.go
@@ -33,7 +33,6 @@ func (c *cniConfig) AttachFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&c.CniMountNetEtcDIR, "cni-etc-path", "/etc/cni/net.d", "cni etc path")
 	cmd.PersistentFlags().StringVar(&c.CniConfigName, "conflist-name", "", "cni conflist name")
 	cmd.PersistentFlags().BoolVar(&c.CniConfigChained, "plugin-cni-chained", true, "kmesh cni plugins chained to another cni")
-	cmd.PersistentFlags().StringVar(&c.ServiceAccountPath, "service-account-token", "/var/run/secrets/kubernetes.io/serviceaccount", "service account path")
 }
 
 func (c *cniConfig) ParseConfig() error {
@@ -45,6 +44,8 @@ func (c *cniConfig) ParseConfig() error {
 	if _, err = os.Stat(c.CniMountNetEtcDIR); err != nil {
 		return err
 	}
+
+	c.ServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount"
 
 	return nil
 }
--- a/daemon/options/cni.go
+++ b/daemon/options/cni.go
@@ -23,15 +23,17 @@ import (
 )
 
 type cniConfig struct {
-	CniMountNetEtcDIR string
-	CniConfigName     string
-	CniConfigChained  bool
+	CniMountNetEtcDIR  string
+	CniConfigName      string
+	CniConfigChained   bool
+	ServiceAccountPath string
 }
 
 func (c *cniConfig) AttachFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&c.CniMountNetEtcDIR, "cni-etc-path", "/etc/cni/net.d", "cni etc path")
 	cmd.PersistentFlags().StringVar(&c.CniConfigName, "conflist-name", "", "cni conflist name")
 	cmd.PersistentFlags().BoolVar(&c.CniConfigChained, "plugin-cni-chained", true, "kmesh cni plugins chained to another cni")
+	cmd.PersistentFlags().StringVar(&c.ServiceAccountPath, "service-account-token", "/var/run/secrets/kubernetes.io/serviceaccount", "service account path")
 }
 
 func (c *cniConfig) ParseConfig() error {
--- a/pkg/cni/install.go
+++ b/pkg/cni/install.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/fsnotify/fsnotify"
 	"istio.io/istio/pkg/filewatcher"
 
 	"kmesh.net/kmesh/pkg/constants"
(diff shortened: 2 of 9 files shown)


pkg/cni/install.go
```diff
@@ -76,6 +82,43 @@ func (i *Installer) Start() error {
...
+				select {
+				case _, ok := <-watcher.Events:
+					if !ok {
+						log.Info("events channel of fsnotify watcher was closed")
+						return
+					}
+
+					// NOTE: due to the update mechanism of service account path, try to update
+					// kubeconfig of Kmesh cni regardless of any events.
+					if err := maybeWriteKubeConfigFile(filepath.Join(i.CniMountNetEtcDIR, kmeshCniKubeConfig)); err != nil {
```
hzxuzhonghu, line 110: This can be improved, when sa changed, there can be multi events received. We should only run rewrite once.

pkg/cni/install.go
```diff
@@ -76,6 +82,43 @@ func (i *Installer) Start() error {
...
+						log.Errorf("failed try to update kubeconfig of Kmesh cni: %v", err)
```
LiZhenCheng9527, line 111:
```suggestion
						log.Errorf("failed try to update kmesh cni kubeconfig: %v", err)
```

daemon/options/cni.go
```diff
@@ -23,15 +23,17 @@ import (
...
+	CniConfigName      string
+	CniConfigChained   bool
+	ServiceAccountPath string
 }
 
 func (c *cniConfig) AttachFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&c.CniMountNetEtcDIR, "cni-etc-path", "/etc/cni/net.d", "cni etc path")
 	cmd.PersistentFlags().StringVar(&c.CniConfigName, "conflist-name", "", "cni conflist name")
 	cmd.PersistentFlags().BoolVar(&c.CniConfigChained, "plugin-cni-chained", true, "kmesh cni plugins chained to another cni")
+	cmd.PersistentFlags().StringVar(&c.ServiceAccountPath, "service-account-token", "/var/run/secrets/kubernetes.io/serviceaccount", "service account path")
```
hzxuzhonghu, line 36: When this path can be different?

pkg/cni/install.go
```diff
@@ -50,24 +57,57 @@ func (i *Installer) removeCniConfig() error {
...
+		for {
+			select {
+			case gotEvent := <-i.Watcher.Events(tokenPath):
+				log.Infof("got event %s", gotEvent.String())
+
+				if err := maybeWriteKubeConfigFile(i.ServiceAccountPath, filepath.Join(i.CniMountNetEtcDIR, kmeshCniKubeConfig)); err != nil {
+					log.Errorf("failed try to update Kmesh cni kubeconfig: %v", err)
+				}
+			case err := <-i.Watcher.Errors(tokenPath):
+				log.Infof("error from errors channel of file watcher: %v", err)
```
hzxuzhonghu, line 103: If you donot return here, how can you stop the watcher

pkg/cni/install.go
```diff
@@ -50,24 +57,57 @@ func (i *Installer) removeCniConfig() error {
...
+	// Start listening for events.
+	go func() {
+		log.Infof("start watching file %s", tokenPath)
+
+		for {
+			select {
+			case gotEvent := <-i.Watcher.Events(tokenPath):
+				log.Infof("got event %s", gotEvent.String())
+
+				if err := maybeWriteKubeConfigFile(i.ServiceAccountPath, filepath.Join(i.CniMountNetEtcDIR, kmeshCniKubeConfig)); err != nil {
```
hzxuzhonghu, line 99: you can refer istio cert watcher to reduce the overwrite times
```
func (s *Server) handleCACertsFileWatch() {
	var timerC <-chan time.Time
	for {
		select {
		case <-timerC:
			timerC = nil
			handleEvent(s)

		case event, ok := <-s.cacertsWatcher.Events:
			if !ok {
				log.Debug("plugin cacerts watch stopped")
				return
			}
			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
				if timerC == nil {
					timerC = time.After(100 * time.Millisecond)
				}
			}

		case err := <-s.cacertsWatcher.Errors:
			if err != nil {
				log.Errorf("failed to catch events on cacerts file: %v", err)
				return
			}

		case <-s.internalStop:
			return
		}
	}
}
```

daemon/options/cni.go
YaoZengzeng: No, but want to configure it one place to avoid defining `ServiceAccountPath` constant everywhere.

Maybe don't set this field as flag, but assign it a default value.

YaoZengzeng:
@hzxuzhonghu PTAL

hzxuzhonghu:
i think this [comments](https://github.com/kmesh-net/kmesh/pull/801#discussion_r1741353538) has not been addressed

YaoZengzeng:
@hzxuzhonghu updated

pkg/cni/install_test.go
```diff
@@ -0,0 +1,85 @@
...
+	os.Setenv("KUBERNETES_SERVICE_PORT", "443")
+
+	expectedKC, err := createKubeConfig(serviceAccountPath)
+	if err != nil {
+		t.Fatalf("failed to create expected kubeconfig: %v", err)
+	}
+
+	cniDir := t.TempDir()
+
+	i := NewInstaller(constants.WorkloadMode, cniDir, "conflist-name", true, serviceAccountPath)
```
hzxuzhonghu, line 46: should close the watcher after test

pkg/cni/install_test.go
```diff
@@ -44,6 +44,7 @@ func TestWatchTokenFile(t *testing.T) {
 	cniDir := t.TempDir()
 
 	i := NewInstaller(constants.WorkloadMode, cniDir, "conflist-name", true, serviceAccountPath)
+	defer i.Watcher.Close()
```
hzxuzhonghu, line 47: should use i.Stop

YaoZengzeng: I didn't call `i.Start()`, so the entire cni installation process was not simulated, so calling `i.Stop()` would fail actually.

hzxuzhonghu (approved):
make sense 
/lgtm

YaoZengzeng:
@hzxuzhonghu need to be merged manually because of codecov

PR #813, opened 2024-09-04
ci: add helm target

zirain:
**What type of PR is this?**

/kind enhancement

add `helm-package` and `helm-push` make targets to support push helm into an OCI registry.

Fixes #

```release-note
add `helm-package` and `helm-push` make targets to support push helm into an OCI registry.
```


zirain:
cc @hzxuzhonghu

deploy/charts/kmesh-helm/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: kmesh
+name: kmesh-helm
```
hzxuzhonghu, line 2: why do you change it

zirain: Since there is already an image named kmesh in the registry, you should use a new name for the helm package.
You could porposal with a better name.

hzxuzhonghu: Ah, then if user install kmesh using the oci, it seems not convenient. how does envoygateway do

zirain: see https://hub.docker.com/r/envoyproxy/gateway-helm/tags

zirain: the installation command will looks like:
```
helm install kmesh oci://ghcr.io/kmesh-net/kmesh-helm --version v0.0.0-latest -n kmesh-system --create-namespace
```

YaoZengzeng: The name of Kmesh daemonset will be the same as this name, which causes the e2e test to fail.

We should decouple the two.

zirain: `fullnameOverride` should help.

hzxuzhonghu (commented):
@zirain if we push to the helm registry, how can we install with that chart there

zirain:
> @zirain if we push to the helm registry, how can we install with that chart there

```shell
helm install kmesh oci://ghcr.io/kmesh-net/kmesh-helm --version v0.0.0-latest -n kmesh-system --create-namespace
```

deploy/charts/kmesh-helm/values.yaml
```diff
@@ -13,3 +13,4 @@ deploy:
         cpu: "1"
         memory: 800Mi
 kubernetesClusterDomain: cluster.local
+fullnameOverride: kmesh
```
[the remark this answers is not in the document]
zirain: After we change the helm chart name, the name of daemonset will change to `kmesh-helm` without this.

Makefile
```diff
@@ -195,3 +195,38 @@ clean:
...
+CHARTS_FOLDER := deploy/charts
+CHARTS := $(wildcard $(CHARTS_FOLDER)/*)
+CHART_VERSION ?= v0.0.0-latest
+CHART_OCI_REGISTRY ?= oci://$(HUB)
+CHART_OUTPUT_DIR ?= out/charts
+
+.PHONY: helm-package.%
+helm-package.%: # Package Helm chart
+	$(eval COMMAND := $(word 1,$(subst ., ,$*)))
+	$(eval CHART_NAME := $(COMMAND))
```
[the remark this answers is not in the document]
zirain: exract `CHART_NAME` when you run `make helm-package.kmesh-helm`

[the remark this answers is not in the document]
hzxuzhonghu: can simplify it, we have only one chart

hzxuzhonghu (commented):
Thanks i will test it in my dev env

Makefile
```diff
@@ -195,3 +195,38 @@ clean:
 	$(QUIET) if docker ps -a -q -f name=kmesh-build | grep -q .; then \
 		docker rm -f kmesh-build; \
 	fi
+
+
+##@ Helm
+
+CHARTS_FOLDER := deploy/charts
+CHARTS := $(wildcard $(CHARTS_FOLDER)/*)
```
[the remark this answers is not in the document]
zirain: it will be easier to extend in the future.

hzxuzhonghu:
/lgtm
/approve

We may need to enable github action later

Issue #820, opened 2024-09-06
"/debug/bpf/*" should output human-readable, real bpf information


YaoZengzeng:
**What would you like to be added**:

"/debug/bpf/*" should output the underlying actual eBPF info, including the contents of the map and prog, in a human-readable way to facilitate troubleshooting the problem.

**Why is this needed**:

Now we have `/debug/config_dump/ads` and `/debug/config_dump/workload` output the config content cached in Kmesh daemon.

At the same time, we have `/debug/bpf/ads`, but its output is exactly the same as `/debug/config_dump/ads`.

Now we hope that `/debug/bpf/*` could output the information about the underlying eBPF.

Due to various reasons, the config in Kmesh daemon cache and the actual eBPF may not be completely consistent. If we can get human-readable eBPF info, it will be more helpful for us to troubleshooting.

 This is also an important part of the `kmeshctl`, used to obtain the real eBPF info of each node.

YaoZengzeng:
cc @Okabe-Rintarou-0 would you like to take it :)

Okabe-Rintarou-0:
sure.
/assign

Okabe-Rintarou-0:
maybe we need to lookup all the keys of bpf map in bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c. bpf supports traversing the keys.

Okabe-Rintarou-0:
I am working on this, we can get all elements with a linked list

Issue #702, opened 2024-08-07
The number tag of config_type fields in Filter proto can not exceeds 3.


yuanqijing:
**Please provide an in-depth description of the question you have**:
Hi, I was trying to add a local_rate_limit field inside the Filter message. 
```
message Filter {
  string name = 1;
  oneof config_type {
    filter.TcpProxy tcp_proxy = 2;
    filter.HttpConnectionManager http_connection_manager = 3;
    filter.LocalRateLimit local_rate_limit = 4;
  }
}
```
After  running `make gen`, the config_type_case field obtained in eBPF is 0, but I expect it to be 4. Additionally, I noticed that if the number tag of **any fields is set to >= 4, config_type_case always shows 0 in C side**. I would like to know how to solve this or this is a bug. Below is a test code to reproduce it.

```golang
apiFilter := &listener_v2.Filter{}
fn := func() {
  buf, err := proto.Marshal(apiFilter)
  if err != nil {
    log.Errorf("proto.Marshal failed: %v", err)
    return
  }
  convertToPack := func(buf []byte) *C.uint8_t {
    return (*C.uint8_t)(unsafe.Pointer(&buf[0]))
  }
  cmsg := C.listener__filter__unpack(nil, C.size_t(len(buf)), convertToPack(buf))
  klog.Infof("listener__filter__unpack: %v", spew.Sdump(cmsg))
}
klog.Infof("test local rate limit")
apiFilter.ConfigType = &listener_v2.Filter_LocalRateLimit{}
fn()
klog.Infof("test tcp proxy")
apiFilter.ConfigType = &listener_v2.Filter_TcpProxy{}
fn()
klog.Infof("test http connection manager")
apiFilter.ConfigType = &listener_v2.Filter_HttpConnectionManager{}
fn()
```

**What do you think about this question?**:

**Environment**:
- Kmesh version:
- Others:

nlgwcy:
You can view the [xDS models supported by kmesh](https://github.com/kmesh-net/kmesh/tree/main/api) in the code repository. Currently, kmesh supports [filter.TcpProxy and filter.HttpConnectionManager](https://github.com/kmesh-net/kmesh/blob/ee20a00d5a226cb3d7e6a77bc140c4052f289080/api/listener/listener_components.proto#L10).

yuanqijing:
[...]

PR #827, opened 2024-09-08
Support ads bpf map lookup all

Okabe-Rintarou-0:
**What type of PR is this?**
feature

support cluster lookup all for dump all the kv of bpf map.
It's for #820 

i've met some problem in developing, seems to be related to the deserialization problem(#702 ) talked about before:
image

i update cluster 1 here, but the lookup and the newly written lookupall both got empty cluster name.

the output is:
lookup: cluster: api_status:UPDATE  connect_timeout:1, api_status:UPDATE  name:"ut-cluster-2"  connect_timeout:1

`ut-cluster-2` is shown, but `ut-cluster-1` does not work.

If we only define `Name` field, the test will even panic.

Fixes #

```release-note

```

### DIFF
--- a/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
+++ b/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
@@ -57,6 +57,16 @@ struct op_context {
         (context).curr_fd = (fd);                                                                                      \
     } while (0)
 
+#define append_new_node(elem_list_head, curr_elem_list_node, new_node)                                                 \
+    do {                                                                                                               \
+        if (curr_elem_list_node == NULL) {                                                                             \
+            curr_elem_list_node = elem_list_head = new_node;                                                           \
+        } else {                                                                                                       \
+            curr_elem_list_node->next = new_node;                                                                      \
+            curr_elem_list_node = new_node;                                                                            \
+        }                                                                                                              \
+    } while (0)
+
 #define TASK_SIZE (100)
 struct inner_map_stat {
     int map_fd;
@@ -800,10 +810,9 @@ static void *create_struct_list(struct op_context *ctx, int *err)
         prev_key = ctx->key;
 
         value = create_struct(ctx, err);
-        if (*err != 0) {
+        if (*err) {
             LOG_ERR("create_struct failed, err = %d\n", err);
-            deserial_free_elem_list(elem_list_head);
-            value = NULL;
+            break;
         }
 
         if (value == NULL) {
@@ -812,17 +821,17 @@ static void *create_struct_list(struct op_context *ctx, int *err)
 
         struct element_list_node *new_node = (struct element_list_node *)calloc(1, sizeof(struct element_list_node));
         if (!new_node) {
-            deserial_free_elem_list(elem_list_head);
-            return NULL;
+            *err = -1;
+            break;
         }
+
         new_node->elem = value;
         new_node->next = NULL;
-        if (curr_elem_list_node == NULL) {
-            curr_elem_list_node = elem_list_head = new_node;
-        } else {
-            curr_elem_list_node->next = new_node;
-            curr_elem_list_node = new_node;
-        }
+        append_new_node(elem_list_head, curr_elem_list_node, new_node);
+    }
+    if (*err) {
+        deserial_free_elem_list(elem_list_head);
+        return NULL;
     }
     return elem_list_head;
 }
@@ -870,8 +879,13 @@ static void *create_struct(struct op_context *ctx, int *err)
         if (ret) {
             LOG_INFO("field[%d] query fail\n", i);
             *err = 1;
-            return value;
+            break;
         }
+    }
+
+    if (*err) {
+        deserial_free_elem(value);
+        return NULL;
     }
 
     return value;
@@ -916,8 +930,6 @@ struct element_list_node *deserial_lookup_all_elems(const void *msg_desciptor)
     value_list_head = create_struct_list(&context, &err);
     if (err != 0) {
         LOG_ERR("create_struct_list failed, err = %d", err);
-        deserial_free_elem_list(value_list_head);
-        value_list_head = NULL;
     }
 
 end:
@@ -971,8 +983,7 @@ void *deserial_lookup_elem(void *key, const void *msg_desciptor)
     normalize_key(&context, key, map_name);
     value = create_struct(&context, &err);
     if (err != 0) {
-        deserial_free_elem(value);
-        value = NULL;
+        LOG_ERR("create_struct failed, err = %d\n", err);
     }
 
 end:
--- a/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
+++ b/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
@@ -19,36 +19,42 @@
 #include "deserialization_to_bpf_map.h"
 #include "../../config/kmesh_marcos_def.h"
 
-#define LOG_ERR(fmt, args...)  printf(fmt, ##args)
-#define LOG_WARN(fmt, args...) printf(fmt, ##args)
-#define LOG_INFO(fmt, args...) printf(fmt, ##args)
+#define PRINTF(fmt, args...) \
+    do {                     \
+        printf(fmt, ##args); \
+        fflush(stdout);      \
+    } while (0)
+
+#define LOG_ERR(fmt, args...) PRINTF(fmt, ##args)
+#define LOG_WARN(fmt, args...) PRINTF(fmt, ##args)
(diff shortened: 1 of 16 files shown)


hzxuzhonghu:
>i've met some problem in developing, seems to be related to the deserialization problem talked about before:

How to reproduce? And the proto gen seems outdated, you can re run `make gen`

Okabe-Rintarou-0:
> > i've met some problem in developing, seems to be related to the deserialization problem talked about before:
> 
> How to reproduce? And the proto gen seems outdated, you can re run `make gen`
For the 1st question: you can checkout to my branch and run the newly added unit test `TestClusterLookupAll`.
For the 2nd question: It's because i did not format the c code.

Okabe-Rintarou-0:
In this pr, i mainly add a new function to lookup all elements. it will return a linked list(since the length can be dynamic)
```c
struct element_list_node* deserial_lookup_all_elems(const void* msg_desciptor) {
    int ret, err;
    struct element_list_node* value_list_head = NULL;
    const char* map_name = NULL;
    struct op_context context = {.inner_map_object = NULL};
    const ProtobufCMessageDescriptor* desc;
    struct bpf_map_info outter_info = {0}, inner_info = {0}, info = {0};
    int map_fd, outter_fd = 0, inner_fd = 0;
    unsigned int id, outter_id = 0, inner_id = 0;

    if (msg_desciptor == NULL)
        return NULL;

    desc = (ProtobufCMessageDescriptor*)msg_desciptor;
    if (desc->magic != PROTOBUF_C__MESSAGE_DESCRIPTOR_MAGIC)
        return NULL;

    map_name = desc->short_name;
    ret = get_map_ids(map_name, &id, &outter_id, &inner_id);
    if (ret)
        return NULL;

    ret = get_map_fd_info(id, &map_fd, &info);
    if (ret < 0) {
        LOG_ERR("invalid MAP_ID: %d\n", id);
        return NULL;
    }

    ret = get_map_fd_info(inner_id, &inner_fd, &inner_info);
    ret |= get_map_fd_info(outter_id, &outter_fd, &outter_info);
    if (ret < 0 || map_info_check(&outter_info, &inner_info))
        goto end;

    init_op_context(context, NULL, NULL, desc, outter_fd, map_fd, &outter_info,
                    &inner_info, &info);

    value_list_head = create_struct_list(&context, &err);
    if (err != 0) {
        LOG_ERR("create_struct_list failed, err = %d", err);
        deserial_free_elem_list(value_list_head);
        value_list_head = NULL;
    }

end:
    if (context.key != NULL)
        free(context.key);
    if (map_fd > 0)
        close(map_fd);
    if (outter_fd > 0)
        close(outter_fd);
    if (inner_fd > 0)
        close(inner_fd);
    return value_list_head;
}
```

pkg/cache/v2/maps/cluster.go
```diff
@@ -76,6 +76,35 @@ func ClusterLookup(key string, value *cluster_v2.Cluster) error {
...
 
+func ClusterLookupAll() ([]*cluster_v2.Cluster, error) {
+	var (
+		err      error
+		clusters []*cluster_v2.Cluster
+	)
+
+	cMsg := C.deserial_lookup_all_elems(unsafe.Pointer(&C.cluster__cluster__descriptor))
+	if cMsg == nil {
+		return nil, fmt.Errorf("ClusterLookupAll deserial_lookup_all_elems failed")
```
LiZhenCheng9527, line 87:
```suggestion
		return nil, fmt.Error("ClusterLookupAll deserial_lookup_all_elems failed")
```

pkg/cache/v2/maps/cluster.go
```diff
@@ -76,6 +76,35 @@ func ClusterLookup(key string, value *cluster_v2.Cluster) error {
 	return err
 }
 
+func ClusterLookupAll() ([]*cluster_v2.Cluster, error) {
+	var (
+		err      error
```
LiZhenCheng9527, line 81: nit:No need to define variables in advance

hzxuzhonghu:
We can implement for the workload map first, which is simpler

Okabe-Rintarou-0:
> We can implement for the workload map first, which is simpler

yes, that can be done by using the bpf go user space method.

for the ads map, i've implement a common method for looking up all keys of a given bpf map, which will return a linked list. We can then use it to implement `LookupAll` function for all ads bpf map.

Okabe-Rintarou-0:
FInished all, cc @hzxuzhonghu @YaoZengzeng

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -19,9 +19,15 @@
 #include "deserialization_to_bpf_map.h"
 #include "../../config/kmesh_marcos_def.h"
 
-#define LOG_ERR(fmt, args...)  printf(fmt, ##args)
-#define LOG_WARN(fmt, args...) printf(fmt, ##args)
-#define LOG_INFO(fmt, args...) printf(fmt, ##args)
+#define PRINTF(fmt, args...)                                                                                           \
+    do {                                                                                                               \
+        printf(fmt, ##args);                                                                                           \
+        fflush(stdout);                                                                                                \
```
hzxuzhonghu, line 25: why do we need this

hzxuzhonghu (commented):
Generally lg, what does the output look like

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
Okabe-Rintarou-0: without fflush we cannot see the log.
the c printf uses a buffer.

I met this when i am debugging.

Okabe-Rintarou-0:
> Generally lg, what does the output look like

will add a ut

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -771,6 +777,56 @@ static int repeat_field_query(struct op_context *ctx, const ProtobufCFieldDescri
...
+    void *value;
+    struct element_list_node *elem_list_head = NULL;
+    struct element_list_node *curr_elem_list_node = NULL;
+
+    *err = 0;
+    ctx->key = calloc(1, ctx->curr_info->key_size);
+    while (!bpf_map_get_next_key(ctx->curr_fd, prev_key, ctx->key)) {
+        prev_key = ctx->key;
+
+        value = create_struct(ctx, err);
```
nlgwcy, line 802: When `create_struct` returns error, value may not be NULL. `create_struct` needs to be optimized. When an error is returned, the value is released in the function instead of depending on the caller. Currently, leakage may occur in `create_struct_list`.

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -771,6 +777,56 @@ static int repeat_field_query(struct op_context *ctx, const ProtobufCFieldDescri
...
+        if (*err != 0) {
+            LOG_ERR("create_struct failed, err = %d\n", err);
+            deserial_free_elem_list(elem_list_head);
```
nlgwcy, line 805: When an error occurs, resources must be reclaimed and returned in a unified manner. The recommended format is as follows:
```c
while (!bpf_map_get_next_key(ctx->curr_fd, prev_key, ctx->key)) {
    value = create_struct(ctx, err);
        if (*err != 0)
           break;
    ......
    struct element_list_node *new_node = (struct element_list_node *)calloc(1, sizeof(struct element_list_node));
        if (!new_node) {
        *err = -1;
        break;
    }
    .......
}

if (*err) {
            deserial_free_elem_list(elem_list_head);
            return NULL;
    }
return elem_list_head;
```

bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -771,6 +777,56 @@ static int repeat_field_query(struct op_context *ctx, const ProtobufCFieldDescri
...
+        if (value == NULL) {
+            continue;
+        }
+
+        struct element_list_node *new_node = (struct element_list_node *)calloc(1, sizeof(struct element_list_node));
+        if (!new_node) {
+            deserial_free_elem_list(elem_list_head);
+            return NULL;
+        }
+        new_node->elem = value;
```
nlgwcy, line 818: These are basic operations for updating list. It is recommended that you encapsulate them into independent functions or macros or call existing libraries.

Okabe-Rintarou-0: thanks, will modify it.

hzxuzhonghu (approved):
we should support dump this from kmeshctl now

PR #828, opened 2024-09-09
refactor mode judgement

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

Fixes #

```release-note

```

### DIFF
--- a/daemon/options/bpf.go
+++ b/daemon/options/bpf.go
@@ -17,6 +17,7 @@
 package options
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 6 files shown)


daemon/options/bpf.go
```diff
@@ -44,6 +45,10 @@ func (c *BpfConfig) AttachFlags(cmd *cobra.Command) {
 func (c *BpfConfig) ParseConfig() error {
 	var err error
 
+	if c.Mode != constants.AdsMode && c.Mode != constants.WorkloadMode {
+		return fmt.Errorf("invalid mode value, should be `ads` or `workload`")
```
hzxuzhonghu, line 49: should allow set empty when user only want to use bypass capability

pkg/controller/client.go
```diff
@@ -58,6 +58,9 @@ func NewXdsClient(mode string, bpfWorkload *bpf.BpfKmeshWorkload) *XdsClient {
 		client.WorkloadController = workload.NewController(bpfWorkload)
 	} else if mode == constants.AdsMode {
 		client.AdsController = ads.NewController()
+	} else {
+		// Shouldn't happen here.
+		panic("invalid start mode, should be \"ads\" or \"workload\"")
```
hzxuzhonghu, line 63: donot call panic everywhere since you have validated above

PR #838, opened 2024-09-10
Add auto build and push kmesh-build image

### LINKED ISSUE #824: Kmesh-build image should be public

Currently this builder image is built manually, add the dockerfile and automatically build it after dependency updated.

**Why is this needed**:

Otherwise we may miss some important upgrade, which is risky

hzxuzhonghu:
**What type of PR is this?**

Fixes #824

```release-note
Added kmesh-build dockerfile and added github action to auto push image
```

### DIFF
--- a/build/docker/builder.dockerfile
+++ b/build/docker/builder.dockerfile
@@ -1,0 +1,35 @@
+# Dockerfile for building kmesh-build docker image.
+#
+# Usage:
+# docker build -f builder.dockerfile -t kmesh-build:latest .
+#
+
+# base image
+FROM openeuler/openeuler:23.09
+
+# Setup Go
+COPY --from=golang:1.22.7 /usr/local/go/ /usr/local/go/
+RUN mkdir -p /go
+ENV GOROOT /usr/local/go
+ENV GOPATH /go
+ENV PATH "${GOROOT}/bin:${GOPATH}/bin:${PATH}"
+
+WORKDIR /prepare
+COPY go.mod ./
+
+# install pkg dependencies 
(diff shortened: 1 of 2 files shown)


build/docker/builder.dockerfile
```diff
@@ -0,0 +1,35 @@
...
+#
+# Usage:
+# docker build -f builder.dockerfile -t kmesh-build:latest .
+#
+
+# base image
+FROM openeuler/openeuler:23.09
+
+# Setup Go
+COPY --from=golang:1.22.7 /usr/local/go/ /usr/local/go/
```
lec-bit, line 11: Do we need to hardcode the golang version? Can we write the golang version as latest?

build/docker/builder.dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# install pkg dependencies 
+# RUN yum install -y kmod util-linux
+# install package in online-compile image
+RUN yum install -y kmod \
+    && yum install -y util-linux iptables
+
+RUN go env -w GO111MODULE=on \
+    && export GONOSUMDB=* \
+    && go env -w GOPROXY=https://goproxy.cn,direct \
```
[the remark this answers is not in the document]
hzxuzhonghu: Yes, i think we can remove it. If run make docker in China, we can add a var when run `make docker`

build/docker/builder.dockerfile
hzxuzhonghu: I think we can do this manually. To be determined

PR #839, opened 2024-09-10
Add proposal for circuitbreaker(JUST A PLACE HOLDER FOR OSPP)

Okabe-Rintarou-0:
**What type of PR is this?**

Fixes #

```release-note

```


docs/proposal/circuit_breaker.md
```diff
@@ -0,0 +1,374 @@
...
+}
+```
+
+Envoy also supports outlier detection. If an endpoint produces too many exceptions (e.g., returning 5xx HTTP status), it will be temporarily removed from the connection pool.
+
+<div align="center">
+    <img src="./pics/outlier_detection.png" />
+</div>
+
+After some time, it will be added back, but if it continues to fail, it will be removed again. Each time it's removed, the wait time to be added back increases.
```
LiZhenCheng9527, line 121:
```suggestion
After some time, it will be added back, but if it continues to fail, it will be removed again. The wait time for re-adding increases with each removal.
```

docs/proposal/circuit_breaker.md
```diff
@@ -0,0 +1,374 @@
...
+| Connection Pool Settings | L4, connection statistic, traffic control            |
+| Outlier Detection        | L4 & L7, http status code statistic, traffic control |
+
+#### Implement the connection pool settings
+
+Here are some counters and gauges in Envoy:
+
++ Host Stats
+
+    | Variable        | Type    |
```
LiZhenCheng9527, line 136: nit: you can explain what these variables mean

docs/proposal/circuit_breaker.md
```diff
@@ -0,0 +1,374 @@
...
+            return;
+        }
+
+        data = bpf_sk_storage_get(&map_of_cluster_sock, sk, 0, BPF_LOCAL_STORAGE_GET_F_CREATE);
+        if (!data) {
+            BPF_LOG(ERR, KMESH, "record_cluster_sock call bpf_sk_storage_get failed\n");
+            return;
+        }
+
+        bpf_strncpy(data->cluster_name, BPF_DATA_MAX_LEN, (char *)cluster_name);
```
hzxuzhonghu, line 279: This is introduced in kernel 5.17, but we should support 5.10

hzxuzhonghu (approved):
Mostly lgtm

docs/proposal/circuit_breaker.md
Okabe-Rintarou-0: https://github.com/kmesh-net/kmesh/blob/cce29bcdc8e709d53d8c08d66bbe9875c40b0453/bpf/kmesh/ads/include/tail_call.h#L88-L92

but i found it also use `bpf_strncpy`, maybe we need an alternative

hzxuzhonghu: IC, this is defined with kernel patch

PR #859, opened 2024-09-15
ads: support envoy filter local ratelimit.

yuanqijing:
**What type of PR is this?**
support local rate limit.

Fixes #

```release-note

```


api/filter/ratelimit.proto
```diff
@@ -0,0 +1,57 @@
...
+      "local_rate_limit_per_downstream_connection": ...,
+      "local_cluster_rate_limit": {...},
+      "enable_x_ratelimit_headers": ...,
+      "vh_rate_limits": ...,
+      "always_consume_default_token_bucket": {...},
+      "rate_limited_as_resource_exhausted": ...
+  }
+*/
+message LocalRateLimit {
+  reserved 1 to 2;
```
hzxuzhonghu, line 49: why reserve?

api/filter/ratelimit.proto
```diff
@@ -0,0 +1,57 @@
...
+      "request_headers_to_add_when_not_enforced": [],
+      "response_headers_to_add": [],
+      "descriptors": [],
+      "stage": ...,
+      "local_rate_limit_per_downstream_connection": ...,
+      "local_cluster_rate_limit": {...},
+      "enable_x_ratelimit_headers": ...,
+      "vh_rate_limits": ...,
+      "always_consume_default_token_bucket": {...},
+      "rate_limited_as_resource_exhausted": ...
```
hzxuzhonghu, line 45: if other fields not supported, please remove them. Add once we support it

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,195 @@
...
+        struct {
+            __u32 ipv4;
+            __u32 port;
+            __u32 family;
+        } sk_skb;
+    } key;
+};
+
+struct ratelimit_value {
+    __u64 last_topup;
```
hzxuzhonghu, line 23: can you add a comment about what is this

hzxuzhonghu (commented):
First we need a proposal to better understand this implement

yuanqijing:
> First we need a proposal to better understand this implement

hi i have created a proposal pr in #873

api/filter/ratelimit.proto
yuanqijing: reserve for
{
 "stat_prefix": ...,
 "status": {...},
 }
 should i start token_bucket with number 1, or keep the number tag ?

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: updated.

yuanqijing:
@hzxuzhonghu hi, ratelimit use the same way to cut tcp connection as circuit breaker,  it can be merged after #570.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,195 @@
...
+    __u64 tokens;     /* Current number of available tokens */
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, struct ratelimit_key);
+    __type(value, struct ratelimit_value);
+    __uint(max_entries, 1000);
+    __uint(map_flags, BPF_F_NO_PREALLOC);
+} ratelimit_map SEC(".maps");
```
hzxuzhonghu, line 33: please keep consistent with other map name.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,195 @@
...
+#include "bpf_common.h"
+#include "kmesh_common.h"
+#include "listener/listener.pb-c.h"
+
+struct ratelimit_key {
+    union {
+        struct {
+            __u32 ipv4;   /* Destination IPv4 address. */
+            __u32 port;   /* Destination port. */
+            __u32 family; /* Address family (e.g., AF_INET) */
```
hzxuzhonghu, line 17: So is this node level ratelimit, compared to sidecar, it is on all node local clients requeests

yuanqijing: yes, global ratelimit needs a external service, i think it is not feasible for ebpf.

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: updated.

api/filter/ratelimit.proto
hzxuzhonghu: You can start from 1

bpf/kmesh/ads/include/filter.h
```diff
@@ -181,6 +182,10 @@ int filter_chain_manager(ctx_buff_t *ctx)
     if (filter_chain == NULL) {
         return KMESH_TAIL_CALL_RET(-1);
     }
+
+    /* ratelimit check */
+    Local_rate_limit__check_and_take(filter_chain, &addr, ctx);
```
hzxuzhonghu, line 187: This is ratelimit on filterchain, where did you support rate limit on listener

yuanqijing: Yes, the ratelimit takes on filterchain, maybe i express it incorrect. The EnvoyFilter YAML object applies to the filter chain, and the rate limit configuration is defined inside the filter. A listener includes multiple filter chains. In eBPF, the listener_manager tail calls the filter_chain_manager.

api/filter/ratelimit.proto
yuanqijing: updated.

pkg/controller/ads/extensions/local_ratelimit.go
```diff
@@ -0,0 +1,56 @@
+package extensions
```
YaoZengzeng, line 1: Add copyright

pkg/controller/ads/extensions/local_ratelimit.go
```diff
@@ -0,0 +1,56 @@
...
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"kmesh.net/kmesh/api/v2/filter"
+	"kmesh.net/kmesh/api/v2/listener"
+)
+
+const LocalRateLimit = "envoy.filters.tcp.local_ratelimit"
+
+// NewLocalRateLimit constructs a new LocalRateLimit filter wrapper.
+func NewLocalRateLimit(filter *listenerv3.Filter) (*listener.Filter_LocalRateLimit, error) {
```
YaoZengzeng, line 19: Any UT or E2E test?

yuanqijing: hi, i added ut tests.

pkg/controller/ads/extensions/local_ratelimit.go
yuanqijing: updated

tacslon:
lgtm, and your test report(google doc) can be pasted here to show your test results @yuanqijing

yuanqijing:
> lgtm, and your test report(google doc) can be pasted here to show your test results @yuanqijing

hi, i have add a [test case](https://docs.google.com/document/d/1SkgIqxCJRYCDqHejtTNCEHUyHecY6UqZPWhIisIR7nw/edit?usp=sharing) here, This PR might only take effect after #570 is merged. Currently, the corresponding log output can be observed in the code. I manually merged the disconnection handling capability from #570 and tested it locally.

hzxuzhonghu:
@yuanqijing Does it support http rate limit

yuanqijing:
> @yuanqijing Does it support http rate limit

No,  it only support tcp connections rate limit as describe in proposal #873.

hzxuzhonghu:
@tacslon Ptal

tacslon:
lgtm, please resolve the conflict @yuanqijing

yuanqijing:
Go test:

bpf_map.go:40: Failed to mount /mnt/kmesh_cgroup2/: device or resource busy
   
cause TestServer_getAndSetBpfLevel Failed, this was wired

yuanqijing:
@tacslon conflict addressed.
 a notice: I add bpf.CleanupBpfMap() in InitBpfMap, the tests passed. Should I create another pr to fix it?

```go
func InitBpfMap(t *testing.T, config options.BpfConfig) (CleanupFn, *bpf.BpfLoader) {
+	bpf.CleanupBpfMap()

	err := os.MkdirAll("/mnt/kmesh_cgroup2", 0755)
        ...
}
```

tacslon:
> @tacslon conflict addressed. a notice: I add bpf.CleanupBpfMap() in InitBpfMap, the tests passed. Should I create another pr to fix it?
> 
> ```go
> func InitBpfMap(t *testing.T, config options.BpfConfig) (CleanupFn, *bpf.BpfLoader) {
> +	bpf.CleanupBpfMap()
> 
> 	err := os.MkdirAll("/mnt/kmesh_cgroup2", 0755)
>         ...
> }
> ```

No problem for me, ptal @hzxuzhonghu

bpf/kmesh/ads/include/filter.h
```diff
@@ -181,6 +182,14 @@ int filter_chain_manager(ctx_buff_t *ctx)
     if (filter_chain == NULL) {
         return KMESH_TAIL_CALL_RET(-1);
     }
+
+    /* ratelimit check */
+    ret = Local_rate_limit__check_and_take(filter_chain, &addr, ctx);
+    if (ret != 0) {
+        BPF_LOG(ERR, FILTERCHAIN, "ratelimit check and take failed, addr=%s\n", ip2str(&addr.ipv4, 1));
```
hzxuzhonghu, line 189:
```suggestion
        BPF_LOG(ERR, FILTERCHAIN, "rate limited, addr=%s\n", ip2str(&addr.ipv4, 1));
```

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+    struct ratelimit_value *value;
+    struct ratelimit_value new_value;
+    __u64 now = bpf_ktime_get_ns();
+    __u64 topup;
+
+    value = bpf_map_lookup_elem(&map_of_local_ratelimit, key);
+    if (!value) {
+        new_value.last_topup = now;
+        new_value.tokens = settings->bucket_size;
+        bpf_map_update_elem(&map_of_local_ratelimit, key, &new_value, BPF_ANY);
```
hzxuzhonghu, line 59: Think of the concurrent case, it could skip the ratelimit. I am not sure if there is a function make lookup and update as atomiclly

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+        return 0;
+    }
+
+    topup = (now - value->last_topup) / settings->topup_interval_ns;
+    if (topup > 0) {
+        value->tokens += topup * settings->tokens_per_topup;
+        if (value->tokens > settings->bucket_size) {
+            value->tokens = settings->bucket_size;
+        }
+        value->last_topup += topup * settings->topup_interval_ns;
```
hzxuzhonghu, line 69: This may greatly influence the performance.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+Local_rate_limit__check_and_take(const Listener__FilterChain *filter_chain, address_t *addr, const ctx_buff_t *ctx)
+{
+    int ret = 0;
+    Listener__Filter *filter = NULL;
+    Filter__LocalRateLimit *rate_limit = NULL;
+    Filter__TokenBucket *token_bucket = NULL;
+
+    ret = Local_rate_limit__filter_chain__match(filter_chain, &filter);
+    if (ret) {
+        BPF_LOG(INFO, FILTERCHAIN, "no local rate limit filter matched\n");
```
hzxuzhonghu, line 103:
```suggestion
        BPF_LOG(DEBUG, FILTERCHAIN, "no local rate limit filter matched\n");
```

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+
+        if (Local_rate_limit__filter__match(filter)) {
+            *filter_ptr = filter;
+            return 0;
+        }
+    }
+    return -1;
+}
+
+#endif
```
hzxuzhonghu, line 197: and a newline

pkg/controller/ads/extensions/local_ratelimit.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return &listener.Filter_LocalRateLimit{
+		LocalRateLimit: localRateLimit,
+	}, nil
+}
+
+// newLocalRateLimit creates a new LocalRateLimit filter.
+func newLocalRateLimit(Filter *listenerv3.Filter) (*filter.LocalRateLimit, error) {
+	unstructured, err := unmarshalToTypedStruct(Filter)
```
hzxuzhonghu, line 48: why donot you import xds and use structured api

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: should I use [bpf_spin_lock](https://docs.ebpf.io/linux/helper-function/bpf_spin_lock/) ?

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: Why would there be a performance impact? Since the interval is typically set to 1 second, the if block runs only about once every second.

bpf/kmesh/ads/include/filter.h
yuanqijing: addressed.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+    __u64 tokens;     /* Current number of available tokens */
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, struct ratelimit_key);
+    __type(value, struct ratelimit_value);
+    __uint(max_entries, 1000);
+    __uint(map_flags, BPF_F_NO_PREALLOC);
+} map_of_local_ratelimit SEC(".maps");
```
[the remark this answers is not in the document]
yuanqijing: addressed.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+
+struct ratelimit_settings {
+    __u64 bucket_size;       /* Maximum capacity of the token bucket */
+    __u64 tokens_per_topup;  /* Number of tokens added per refill */
```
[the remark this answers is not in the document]
yuanqijing: addressed.

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: addressed.

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: addressed.

bpf/kmesh/ads/include/local_ratelimit.h
hzxuzhonghu: We could not assume the interval cannot be smaller. In addition, the tricky point is concurrent connections.

pkg/controller/ads/extensions/local_ratelimit.go
hzxuzhonghu: FYI https://github.com/istio/istio/blob/master/pkg/wasm/convert.go#L145-L167

TypedExtensionConfig can also be unmarshaled

yuanqijing:
``` go
func main() {
	bucket := &v1.TypedStruct{
		TypeUrl: "type.googleapis.com/envoy.extensions.filters.http.local_ratelimit.v3.LocalRateLimit",
		Value: &structpb.Struct{
			Fields: .....
		},
	}

	anyTyped, _ := anypb.New(bucket)
	extConfig := &corev3.TypedExtensionConfig{}
	if err := anyTyped.UnmarshalTo(extConfig); err != nil { // panic
		panic(err)
	}

	typedStruct := &v1.TypedStruct{}
	if err := anyTyped.UnmarshalTo(typedStruct); err != nil { // no panic
		panic(err)
	}
}
```

I tested it, and it seems it cannot unmarshal into TypedExtensionConfig.

Panic: proto: mismatched message type: got "envoy.config.core.v3.TypedExtensionConfig", want "udpa.type.v1.TypedStruct"

Istio is using TypedStruct, I don't understand how to modify it.

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: Or just using [bpf_spin_lock](https://docs.ebpf.io/linux/helper-function/bpf_spin_lock/), but it hurt performance.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,197 @@
...
+        BPF_LOG(ERR, FILTERCHAIN, "get rate_limit failed\n");
+        return -1;
+    }
+    token_bucket = kmesh_get_ptr_val(rate_limit->token_bucket);
+    if (!token_bucket) {
+        BPF_LOG(ERR, FILTERCHAIN, "get token_bucket failed\n");
+        return -1;
+    }
+
+    struct ratelimit_key key = {
```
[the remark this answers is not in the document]
yuanqijing: Why is source service information needed? Currently, the design implements rate limiting based on the destination service, where each destination service maintains two states in map: tokens and last_topup.

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,199 @@
...
+            value->tokens = settings->max_tokens;
+        }
+        value->last_topup += topup * settings->fill_interval;
+    }
+
+    if (value->tokens == 0) {
+        return -1;
+    }
+
+    value->tokens--;
```
nlgwcy, line 78:
```suggestion
    __sync_fetch_and_add(&value->tokens, -1);
```

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,199 @@
...
+struct ratelimit_value {
+    __u64 last_topup; /* Timestamp of the last token refill (nanoseconds) */
+    __u64 tokens;     /* Current number of available tokens */
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __type(key, struct ratelimit_key);
+    __type(value, struct ratelimit_value);
+    __uint(max_entries, 1000);
```
nlgwcy, line 31: Pls define a macro.

bpf/kmesh/ads/include/local_ratelimit.h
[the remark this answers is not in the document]
nlgwcy: Yes. The current design is node-level rate limit, which is different from the traditional local rate limit.

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: Change to MAP_SIZE_OF_LISTENER, because its size will not exceed the number of listeners. Each listener corresponds to at most one ratelimit entry.

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: addressed.

hzxuzhonghu:
We also need to delete `kmesh_ratelimit` whenever a pod deleted, otherwise it cause mem leak

yuanqijing:
> We also need to delete `kmesh_ratelimit` whenever a pod deleted, otherwise it cause mem leak

should i create a global var (option 1) or modify NewCache Function Signiture(option 2) like this [commit](https://github.com/kmesh-net/kmesh/pull/570/commits/f6362ccd5904c06b14f76f5be6630dc701c76158) do.
```go
var BpfAds *BpfAds
var once     sync.Once

// NewBpfAds creates or returns the existing BpfAds instance, ensuring that only one instance is created (singleton).
func NewBpfAds(cfg *options.BpfConfig) (*BpfAds, error) {
	var err error
	once.Do(func() {
		BpfAds = &BpfAds{}
		err = BpfAds.SockConn.NewBpf(cfg)
		if err != nil {
			BpfAds = nil
		}
	})
	return BpfAds, err
}
```
Option 1: Easy to implement, no impact on existing code, but exposing a global variable might introduce risks.
Option 2: Affects the interface. I'd add a new function, like NewCacheWithConfig(cfg *CacheConfig), to avoid breaking changes.

tacslon:
> > We also need to delete `kmesh_ratelimit` whenever a pod deleted, otherwise it cause mem leak
> 
> Thanks, you're right.
> 
> Should i create a global var (option 1) or modify NewCache Function Signiture(option 2) like this [commit](https://github.com/kmesh-net/kmesh/pull/570/commits/f6362ccd5904c06b14f76f5be6630dc701c76158) do.
> 
> ```go
> var BpfAds *BpfAds
> var once     sync.Once
> 
> // NewBpfAds creates or returns the existing BpfAds instance, ensuring that only one instance is created (singleton).
> func NewBpfAds(cfg *options.BpfConfig) (*BpfAds, error) {
> 	var err error
> 	once.Do(func() {
> 		BpfAds = &BpfAds{}
> 		err = BpfAds.SockConn.NewBpf(cfg)
> 		if err != nil {
> 			BpfAds = nil
> 		}
> 	})
> 	return BpfAds, err
> }
> ```
> 
> Option 1: Easy to implement, no impact on existing code, but exposing a global variable might introduce risks. Option 2: Affects the interface. Or add a new function, like NewCacheWithConfig(cfg *CacheConfig), to avoid breaking changes.

Introducing a global variable may not be a good choice, option 2(add a new function) should be better

tacslon:
> We also need to delete `kmesh_ratelimit` whenever a pod deleted, otherwise it cause mem leak

@hzxuzhonghu Maybe this PR can be merged, then @yuanqijing creates another PR to fix this issue?

hzxuzhonghu:
@yuanqijing I think both @nlgwcy and me are concerned this is a per node ratelimit, while with istio generated xds, it is per pod.
 https://github.com/kmesh-net/kmesh/pull/859#discussion_r1810778940

yuanqijing:
> @yuanqijing I think both @nlgwcy and me are concerned this is a per node ratelimit, while with istio generated xds, it is per pod. [#859 (comment)](https://github.com/kmesh-net/kmesh/pull/859#discussion_r1810778940)

Yes, this is a per node ratelimit, it does not ratelimit per pod connection rates. It ratelimit downstream pods to upstream service. [The ducument](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting#arch-overview-global-rate-limit):
> Envoy will call the rate limit service for every new connection on the listener where the filter is installed. The configuration specifies a specific domain and descriptor set to rate limit on.

seems ratelimit does not specifiy which downstream pods should be processed, i dont get it. I understand that traditional rate limiting can restrict each pod, which makes sense. However, this might not align with Envoy's documentation.

yuanqijing:
> @yuanqijing I think both @nlgwcy and me are concerned this is a per node ratelimit, while with istio generated xds, it is per pod. [#859 (comment)](https://github.com/kmesh-net/kmesh/pull/859#discussion_r1810778940)

Regarding this point
> while with Istio generated xDS, it is per pod' 

I don't quite understand which rate limit configuration case the statement refers to. Could you provide a specific example for discussion?

hzxuzhonghu:
In kmesh, we are a little different, we can only ratelimit on the client side now. So take the example here, though it is http. 

https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#local-rate-limit

```
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: filter-local-ratelimit-svc
  namespace: istio-system
spec:
  workloadSelector:
    labels:
      app: productpage
  configPatches:
    - applyTo: NETWORK_FILTER
      match:
        context: SIDECAR_OUTBOUND
```


It should be applied on each client instance, not on all the instances of a node.

yuanqijing:
> In kmesh, we are a little different, we can only ratelimit on the client side now. So take the example here, though it is http.
> 
> https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/#local-rate-limit
> 
> ```
> apiVersion: networking.istio.io/v1alpha3
> kind: EnvoyFilter
> metadata:
>   name: filter-local-ratelimit-svc
>   namespace: istio-system
> spec:
>   workloadSelector:
>     labels:
>       app: productpage
>   configPatches:
>     - applyTo: NETWORK_FILTER
>       match:
>         context: SIDECAR_OUTBOUND
> ```
> 
> It should be applied on each client instance, not on all the instances of a node.

Understood, so the key point here is whether intances in a node share a token bucket.
option 1:  client_A->Service_C: 5 token, client_B->Service_C: 6 token
option 2: Service_C: 6 token.

Envoy using option 2, but in kmesh we use option1 ? Is my understanding correct?

hzxuzhonghu:
> Envoy using option 2, but in kmesh we use option1 ? Is my understanding correct?

Envoy can use both, kmesh we can only use option 1

yuanqijing:
> > Envoy using option 2, but in kmesh we use option1 ? Is my understanding correct?
> 
> Envoy can use both, kmesh we can only use option 1

I have added a new field to distinguish different pods:
```
struct ratelimit_key {
    union {
        struct {
            __u64 netns;  /* Network namespace. */
            __u32 ipv4;   /* Destination IPv4 address. */
            __u32 port;   /* Destination port. */
            __u32 family; /* Address family (e.g., AF_INET) */
        } sk_skb;
    } key;
};
```

yuanqijing:
@hzxuzhonghu  Hi I think there are 2 pending Issues:
1. When cleaning up the listener, also clean up the ratelimit configuration. Can I create another PR to fix this?
2. The concurrent issue here may not have been fully discussed yet.

hzxuzhonghu:
> 1. When cleaning up the listener, also clean up the ratelimit configuration. Can I create another PR to fix this?

Yes, i think its ok

> 2. The concurrent issue here may not have been fully discussed yet.

@nlgwcy suggested using atomic update

yuanqijing:
> > 1. When cleaning up the listener, also clean up the ratelimit configuration. Can I create another PR to fix this?
> 
> Yes, i think its ok
> 
> > 2. The concurrent issue here may not have been fully discussed yet.
> 
> @nlgwcy suggested using atomic update

suggested atomic updates all addressed. https://github.com/kmesh-net/kmesh/pull/859#discussion_r1810771130 https://github.com/kmesh-net/kmesh/pull/859#discussion_r1810771130

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,201 @@
...
+        value = bpf_map_lookup_elem(&kmesh_ratelimit, key);
+        if (!value)
+            return 0;
+    }
+
+    topup = (now - value->last_topup) / settings->fill_interval;
+    if (topup > 0) {
+        value->tokens += topup * settings->tokens_per_fill;
+        if (value->tokens > settings->max_tokens) {
+            value->tokens = settings->max_tokens;
```
hzxuzhonghu, line 70: This is not concurrent safe

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,201 @@
...
+#ifndef __KMESH_LOCAL_RATE_LIMIT_H__
+#define __KMESH_LOCAL_RATE_LIMIT_H__
+
+#include "bpf_log.h"
+#include "bpf_common.h"
+#include "kmesh_common.h"
+#include "listener/listener.pb-c.h"
+
+struct ratelimit_key {
+    union {
```
hzxuzhonghu, line 13: why use union here

yuanqijing: Using a union in the ratelimit_key structure improves scalability. If we need to limit HTTP connections in the future, we can reuse the same key structure. For example, we can add a structure for HTTP connections like this:

```

struct ratelimit_key {
    union {
        struct {
            __u64 netns;  /* Network namespace. */
            __u32 ipv4;   /* Destination IPv4 address. */
            __u32 port;   /* Destination port. */
            __u32 family; /* Address family (e.g., AF_INET) */
        } sk_skb;

        struct {
            __u32 http_method;   /* HTTP method (e.g., GET, POST) */
            __u32 http_version;    /* HTTP version (e.g., 1.1, 2) */
            __u32 http_port;         /* HTTP port */
        } http; // New HTTP connection related fields
    } key;
};

```

bpf/kmesh/ads/include/local_ratelimit.h
yuanqijing: can this ensue concurent safe ? cause only one progress can add the value.

```c
  topup = (now - value->last_topup) / settings->fill_interval;
  if (topup > 0) {
      topup_time = value->last_topup + topup * settings->fill_interval;
      if (__sync_bool_compare_and_swap(&value->last_topup, value->last_topup, topup_time)) {
          delta = topup * settings->tokens_per_fill;
          if (value->tokens + delta > settings->max_tokens) {
              delta = settings->max_tokens - value->tokens;
          }
          __sync_fetch_and_add(&value->tokens, delta);
      }
  }

```

hzxuzhonghu: @nlgwcy is an expert on this

bpf/kmesh/ads/include/local_ratelimit.h
```diff
@@ -0,0 +1,201 @@
...
+    struct ratelimit_settings settings = {
+        .max_tokens = token_bucket->max_tokens,
+        .tokens_per_fill = token_bucket->tokens_per_fill,
+        .fill_interval = token_bucket->fill_interval,
+    };
+
+    if (rate_limit__check_and_take(&key, &settings)) {
+        BPF_LOG(INFO, FILTERCHAIN, "rate limit exceeded\n");
+// TODO: remove this after #570 merged.
+#ifndef MARK_REJECTED
```
nlgwcy, line 137: It's not needed anymore.

PR #866, opened 2024-09-18
Enable non-default cluster

### LINKED ISSUE #524: [E2E Enchancement] running E2E test cases in an existing cluster

Now running the E2E test script will always create a brand new k8s cluster and deploy all of things like istio, Kmesh，etc on it.

But during the development process, many times we already have a test k8s cluster and deployed the modified components on it, and we just want to run all the E2E test cases to make sure everything works fine.

Maybe we could provide command line flag that allow users to specify the cluster they wanto run the E2E test suite on.

@noobwei

noobwei:
**What type of PR is this?**

/kind enhancement

Enable non-default cluster

Fixes #524 

```release-note

```

### DIFF
--- a/test/e2e/run_test.sh
+++ b/test/e2e/run_test.sh
@@ -232,7 +232,7 @@ while (( "$#" )); do
         echo "Error: Cluster '$NAME' does not exist."
         exit 1
       fi
-      shift
+      shift 2
     ;;
     --ipv6)
       IPV6=true
@@ -249,6 +249,8 @@ while (( "$#" )); do
     ;;
     esac
 done
+
+NAME="${NAME:-kmesh-testing}"
 
 if [[ -z "${SKIP_INSTALL_DEPENDENCIES:-}" ]]; then
     install_dependencies
--- a/test/e2e/run_test.sh
+++ b/test/e2e/run_test.sh
@@ -232,8 +232,6 @@ while (( "$#" )); do
         echo "Error: Cluster '$NAME' does not exist."
         exit 1
       fi
-      SKIP_SETUP=true
-      echo "Using existing cluster '$NAME'. Setup will be skipped."
       shift
     ;;
     --ipv6)
--- a/test/e2e/run_test.sh
+++ b/test/e2e/run_test.sh
@@ -20,24 +20,10 @@ function setup_kind_cluster() {
     local NAME="${1:-kmesh-testing}"
     local IMAGE="${2:-"${DEFAULT_KIND_IMAGE}"}"
 
-    local FORCE_RECREATE=false
-
-    if [ "$NAME" == "kmesh-testing" ]; then
-        FORCE_RECREATE=true
-    fi
-    
-
-    if (kind get clusters | grep -q "^${NAME}$"); then
-        if [ "$FORCE_RECREATE" = true ]; then
-            echo "Deleting previous KinD cluster with name=${NAME}"
-            if ! (kind delete cluster --name="${NAME}" -v9) > /dev/null; then
-                echo "Failed to delete existing kind cluster with name ${NAME}. Continuing with creation..."
-            fi
-        else
-            echo "Cluster ${NAME} exists. Skipping creation."
-            kubectl config use-context "$NAME"
-            return
-        fi
+    # Delete any previous KinD cluster.
+    echo "Deleting previous KinD cluster with name=${NAME}"
+    if ! (kind delete cluster --name="${NAME}" -v9) > /dev/null; then
+        echo "No existing kind cluster with name ${NAME}. Continue..."
     fi
 
     # Create KinD cluster.
@@ -219,7 +205,6 @@ function cleanup_docker_registry() {
 }
 
 PARAMS=()
-USE_EXISTING_CLUSTER=false
 
 while (( "$#" )); do
     case "$1" in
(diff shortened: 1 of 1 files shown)


test/e2e/run_test.sh
```diff
@@ -226,6 +241,21 @@ while (( "$#" )); do
       SKIP_BUILD=true
       shift
     ;;
+    --use-cluster)
```
YaoZengzeng, line 244: Name it "--cluster"

test/e2e/run_test.sh
```diff
@@ -226,6 +241,21 @@ while (( "$#" )); do
...
+      if [ -n "$2" ] && [[ "$2" != --* ]]; then
```
YaoZengzeng, line 245: There is no need to check whether the cluster name is correct, because the parameters don't just start with "--". If it is not correct, it will fail directly.

test/e2e/run_test.sh
```diff
@@ -226,6 +241,21 @@ while (( "$#" )); do
...
+        NAME="$2"
+        USE_EXISTING_CLUSTER=true
+        if ! kubectl config get-contexts | grep -q "$NAME"; then
```
YaoZengzeng, line 248: Using grep doesn't seem very rigorous. Why doesnt' it strictly match the result of `kind get clusters`

test/e2e/run_test.sh
```diff
@@ -226,6 +241,21 @@ while (( "$#" )); do
       SKIP_BUILD=true
       shift
     ;;
+    --use-cluster)
+      if [ -n "$2" ] && [[ "$2" != --* ]]; then
+        NAME="$2"
+        USE_EXISTING_CLUSTER=true
```
YaoZengzeng, line 247: No `USE_EXISTING_CLUSTER` is needed. If the user wants to  use an exisiting cluster, just specify `--skip-setup`

test/e2e/run_test.sh
```diff
@@ -246,15 +276,23 @@ if [[ -z "${SKIP_INSTALL_DEPENDENCIES:-}" ]]; then
...
+if [[ -z "${SKIP_SETUP:-}" ]] && [[ "$USE_EXISTING_CLUSTER" = false ]]; then
+    setup_kind_cluster "$NAME"
 fi
 
 if [[ -z "${SKIP_BUILD:-}" ]]; then
     setup_kind_registry
     build_and_push_images
 fi
 
+kubectl config use-context "$NAME"
```
YaoZengzeng, line 288: If name is a kind cluster, the prefix `kind-` should be added here.

test/e2e/run_test.sh
```diff
@@ -246,15 +276,23 @@ if [[ -z "${SKIP_INSTALL_DEPENDENCIES:-}" ]]; then
...
+echo "Running tests in cluster '$NAME'"
+
+if [[ "$USE_EXISTING_CLUSTER" = false ]]; then
```
YaoZengzeng, line 291: This code block is not needed.

test/e2e/run_test.sh
```diff
@@ -20,10 +20,24 @@ function setup_kind_cluster() {
     local NAME="${1:-kmesh-testing}"
     local IMAGE="${2:-"${DEFAULT_KIND_IMAGE}"}"
 
-    # Delete any previous KinD cluster.
-    echo "Deleting previous KinD cluster with name=${NAME}"
-    if ! (kind delete cluster --name="${NAME}" -v9) > /dev/null; then
-        echo "No existing kind cluster with name ${NAME}. Continue..."
+    local FORCE_RECREATE=false
```
YaoZengzeng, line 23: Also no need for logic of `FORCE_RECREATE`, keep `setup_kind_cluster` unchanged.

test/e2e/run_test.sh
noobwei: I guess lines 288 & 289 are not included.

YaoZengzeng: > I guess lines 288 & 289 are not included.

Yes, just `USE_EXISTING_CLUSTER` related code.

test/e2e/run_test.sh
```diff
@@ -226,6 +241,21 @@ while (( "$#" )); do
...
+      if [ -n "$2" ] && [[ "$2" != --* ]]; then
+        NAME="$2"
+        USE_EXISTING_CLUSTER=true
+        if ! kubectl config get-contexts | grep -q "$NAME"; then
+        echo "Error: Cluster context '$NAME' does not exist."
+        exit 1
+        fi
+        shift
+      else
+          echo "Error: Missing cluster name for --use-cluster"
```
LiZhenCheng9527:
```suggestion
          echo "Error: Missing cluster name for --cluster"
```

hzxuzhonghu (dismissed):
/lgtm

But i donot think it worth much to spend time on such a tiny case

test/e2e/run_test.sh
```diff
@@ -226,6 +226,16 @@ while (( "$#" )); do
       SKIP_BUILD=true
       shift
     ;;
+    --cluster)
+      NAME="$2"
+      if ! kind get clusters | grep -qw "$NAME"; then
```
YaoZengzeng, line 231: Don't nest `SKIP_SETUP` logic here, use `--cluster` to get cluster name is enough.

Whether to create a cluster follow the previous logic.

test/e2e/run_test.sh
```diff
@@ -247,14 +255,18 @@ if [[ -z "${SKIP_INSTALL_DEPENDENCIES:-}" ]]; then
...
-    setup_kind_cluster
+    setup_kind_cluster "$NAME"
 fi
 
 if [[ -z "${SKIP_BUILD:-}" ]]; then
     setup_kind_registry
     build_and_push_images
 fi
 
+kubectl config use-context "kind-$NAME"
```
YaoZengzeng, line 266: CI failed:

```bash
error: no context exists with the name: "kind-"
make: *** [Makefile:157: e2e] Error 1
``

If `--cluster` is not specified, NAME would be empty. Maybe we should specify NAME as a global variable.

PR #873, opened 2024-09-19
Add proposal for local rate limit.

yuanqijing:
**What type of PR is this?**

Fixes #

```release-note

```


docs/proposal/local_ratelimit.md
```diff
@@ -0,0 +1,110 @@
...
+- TBD
+
+creation-date: 2024-09-19
+
+---
+
+## Support Local Rate Limit
+
+### Summary
+Envoy supports local rate limiting, but Kmesh currently does not. This proposal aims to add local rate limiting to Kmesh, focusing on per-connection rate limiting.
```
hzxuzhonghu, line 21: what does per-connection mean, doesnot it can be used to limit max connections?

yuanqijing: > what does per-connection mean, doesnot it can be used to limit max connections?
what does per-connection mean
===
Rate limit works at per tcp connection (connect syscall) and decide whether to deny the connection.

doesnot it can be used to limit max connections?
===
No, it limits the rate of TCP connection attempts rather than the total number of concurrent connections. For example, this proposal allows a client to initiate up to four TCP connections to a target machine per minute. A fifth attempt within the same minute will be denied, but a connection attempt in the next minute will be allowed.

hzxuzhonghu: Yeah, understood.

hzxuzhonghu (approved):
@tacslon any other comments

PR #877, opened 2024-09-20
Prevent potential panic when object is nil or empty

### LINKED ISSUE #875: Another potential ads deserial bug
**What happened**:
```go
func socketAddressToClang(goMsg *core_v2.SocketAddress) (*C.Core__SocketAddress, error) {
	buf, err := proto.Marshal(goMsg)
	if err != nil {
		return nil, err
	}

	cMsg := C.core__socket_address__unpack(nil, C.size_t(len(buf)), convertToPack(buf))
	if cMsg == nil {
		return nil, fmt.Errorf("core__socket_address__unpack failed")
	}
	return cMsg, nil
}
```

take the above code as example, if the field of goMsg is all zero, proto.Marshal will produce a buf with zero length.
then convertToPack will cause panic.

cc @hzxuzhonghu @nlgwcy 

**Environment**:
- Kmesh version:
- Others:

hzxuzhonghu:
**What type of PR is this?**

Fixes #875

```release-note

```

### DIFF
--- a/pkg/cache/v2/maps/authz.go
+++ b/pkg/cache/v2/maps/authz.go
@@ -31,6 +31,9 @@ import (
 
 func authorizationToGolang(goMsg *security_v2.Authorization, cMsg *C.Istio__Security__Authorization) error {
 	buf := make([]byte, C.istio__security__authorization__get_packed_size(cMsg))
+	if len(buf) == 0 {
+		return nil
+	}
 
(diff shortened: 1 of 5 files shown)


hzxuzhonghu:
cc @Okabe-Rintarou-0 How about this fix?

Okabe-Rintarou-0:
> cc @Okabe-Rintarou-0 How about this fix?

seems good

PR #890, opened 2024-09-24
make xds request send async

lec-bit:
**What type of PR is this?**
/kind bug

When quickly launching a large number of pods and services in a Kubernetes cluster, the kmesh handadsresponse speed is insufficient. Additionally, the gRPC stream's sending and receiving code runs in the same coroutine, which can lead to deadlock issues. We are now separating the sending and receiving processes of the gRPC stream into different coroutines.

Fixes #

```release-note

```


hzxuzhonghu:
/lgtm

Please fix the lint

PR #900, opened 2024-09-26
Add locality loadbalance to kmesh workload mode.

derekwin:
**What type of PR is this?**
/kind enhancement

~This PR is not ready to be merged; some features are not working properly.~
~1. The priority of the service Pod written to the BPF map is observed to be correct in user space, but the behavior of locality load balancing in kernel space is random. (This issue does not exist in the test version of the code at [this link](https://github.com/derekwin/kmesh/tree/lb-dev-user-bak).)~
~2. The code for the workload has undergone significant changes in the past one to two months. The current PR's code attempts to adapt to the new logic of the workload as much as possible. However, a lot of redundant deletion behaviors have been observed. These parts of the code need to be optimized in conjunction with the existing workload logic.~

This pr is ok.

But i find a bug in new version kubectl and istio work with kmesh(https://github.com/kmesh-net/kmesh/issues/910).
> this pr's enhancement need new version kubectl and istio.

### DIFF
--- a/bpf/kmesh/workload/include/service.h
+++ b/bpf/kmesh/workload/include/service.h
@@ -23,7 +23,7 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
         return 0;
 
     endpoint_k.service_id = service_id;
-    endpoint_k.prio = 0; // for random handle，all endpoints are saved in max priority
+    endpoint_k.prio = 0; // for random handle，all endpoints are saved with highest priority
 
     rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[0] + 1;
     endpoint_k.backend_index = rand_k;
--- a/bpf/kmesh/workload/include/service.h
+++ b/bpf/kmesh/workload/include/service.h
@@ -19,10 +19,13 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
     endpoint_value *endpoint_v = NULL;
     int rand_k = 0;
 
-    endpoint_k.service_id = service_id;
-    endpoint_k.prio = MIN_PRIO; // for random handle，all endpoints are saved in MIN_PRIO
+    if (service_v->prio_endpoint_count[0] == 0)
+        return 0;
 
-    rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[MIN_PRIO] + 1;
+    endpoint_k.service_id = service_id;
+    endpoint_k.prio = 0; // for random handle，all endpoints are saved in max priority
+
+    rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[0] + 1;
     endpoint_k.backend_index = rand_k;
 
     endpoint_v = map_lookup_endpoint(&endpoint_k);
@@ -49,6 +52,8 @@ lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, s
     endpoint_key endpoint_k = {0};
     endpoint_value *endpoint_v = NULL;
     endpoint_k.service_id = service_id;
+    struct ip_addr zero_addr = {0};
+    __u32 zero_port = 0;
 
     // #pragma unroll
     for (int match_prio = 0; match_prio < PRIO_COUNT; match_prio++) {
@@ -72,9 +77,13 @@ lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, s
                     BPF_LOG(ERR, SERVICE, "endpoint_manager failed, ret:%d\n", ret);
                 return ret;
             }
+            BPF_LOG(DEBUG, SERVICE, "locality loadbalance matched backend_uid %d\n", endpoint_v->backend_uid);
             return 0; // find the backend successfully
         }
-        if (is_strict && match_prio == service_v->lb_strict_prio) { // only match lb strict index
+        if (is_strict) { // only match max priority in strict mode
+            kmesh_ctx->dnat_ip = zero_addr;
+            kmesh_ctx->dnat_port = zero_port;
+            BPF_LOG(DEBUG, SERVICE, "locality loadbalance match nothing in STRICT mode\n");
             return -ENOENT;
         }
     }
--- a/bpf/kmesh/workload/include/service.h
+++ b/bpf/kmesh/workload/include/service.h
@@ -17,9 +17,13 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
     int ret = 0;
     endpoint_key endpoint_k = {0};
     endpoint_value *endpoint_v = NULL;
+    int rand_k = 0;
 
     endpoint_k.service_id = service_id;
-    endpoint_k.backend_index = bpf_get_prandom_u32() % service_v->endpoint_count + 1;
+    endpoint_k.prio = MAX_PRIO; // for random handle，all endpoints are saved in MAX_PRIO
+
+    rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[MAX_PRIO] + 1;
+    endpoint_k.backend_index = rand_k;
 
     endpoint_v = map_lookup_endpoint(&endpoint_k);
     if (!endpoint_v) {
@@ -35,6 +39,47 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
     }
 
     return 0;
+}
+
+static inline int
+lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, service_value *service_v, bool is_strict)
+{
+    int ret = 0;
+    uint32_t rand_k = 0;
+    endpoint_key endpoint_k = {0};
+    endpoint_value *endpoint_v = NULL;
+    endpoint_k.service_id = service_id;
+
+    // #pragma unroll
+    for (int match_rank = MAX_PRIO; match_rank >= 0; match_rank--) {
+        endpoint_k.prio = match_rank; // 6->0
+        // if we have endpoints in this prio
+        if (service_v->prio_endpoint_count[match_rank] > 0) {
+            rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[match_rank] + 1;
+            if (rand_k >= MAP_SIZE_OF_BACKEND) {
+                return -ENOENT;
+            }
+            endpoint_k.backend_index = rand_k;
+            endpoint_v = map_lookup_endpoint(&endpoint_k);
+            if (!endpoint_v) {
+                BPF_LOG(
+                    ERR, SERVICE, "find endpoint [%u/%u/%u] failed", service_id, match_rank, endpoint_k.backend_index);
+                return -ENOENT;
+            }
+            ret = endpoint_manager(kmesh_ctx, endpoint_v, service_id, service_v);
+            if (ret != 0) {
+                if (ret != -ENOENT)
+                    BPF_LOG(ERR, SERVICE, "endpoint_manager failed, ret:%d\n", ret);
+                return ret;
+            }
+            return 0; // find the backend successfully
+        }
+        if (is_strict && match_rank == service_v->lb_strict_index) { // only match lb strict index
+            return -ENOENT;
+        }
+    }
+    // no backend matched
+    return -ENOENT;
 }
 
 static inline int service_manager(struct kmesh_context *kmesh_ctx, __u32 service_id, service_value *service_v)
@@ -55,14 +100,17 @@ static inline int service_manager(struct kmesh_context *kmesh_ctx, __u32 service
         return ret;
     }
 
-    if (service_v->endpoint_count == 0) {
-        BPF_LOG(DEBUG, SERVICE, "service %u has no endpoint", service_id);
-        return 0;
-    }
+    BPF_LOG(DEBUG, SERVICE, "service [%u] policy [%u] failed", service_id, service_v->lb_policy);
 
     switch (service_v->lb_policy) {
     case LB_POLICY_RANDOM:
         ret = lb_random_handle(kmesh_ctx, service_id, service_v);
+        break;
+    case LB_POLICY_STRICT:
+        ret = lb_locality_failover_handle(kmesh_ctx, service_id, service_v, true);
+        break;
+    case LB_POLICY_FAILOVER:
+        ret = lb_locality_failover_handle(kmesh_ctx, service_id, service_v, false);
         break;
     default:
         BPF_LOG(ERR, SERVICE, "unsupported load balance type:%u\n", service_v->lb_policy);
--- a/bpf/kmesh/workload/include/service.h
+++ b/bpf/kmesh/workload/include/service.h
@@ -20,9 +20,9 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
     int rand_k = 0;
 
     endpoint_k.service_id = service_id;
-    endpoint_k.prio = MAX_PRIO; // for random handle，all endpoints are saved in MAX_PRIO
+    endpoint_k.prio = MIN_PRIO; // for random handle，all endpoints are saved in MIN_PRIO
 
-    rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[MAX_PRIO] + 1;
+    rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[MIN_PRIO] + 1;
     endpoint_k.backend_index = rand_k;
 
     endpoint_v = map_lookup_endpoint(&endpoint_k);
@@ -51,11 +51,11 @@ lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, s
     endpoint_k.service_id = service_id;
 
     // #pragma unroll
-    for (int match_rank = MAX_PRIO; match_rank >= 0; match_rank--) {
-        endpoint_k.prio = match_rank; // 6->0
+    for (int match_prio = 0; match_prio < PRIO_COUNT; match_prio++) {
+        endpoint_k.prio = match_prio; // 0->6
         // if we have endpoints in this prio
-        if (service_v->prio_endpoint_count[match_rank] > 0) {
-            rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[match_rank] + 1;
+        if (service_v->prio_endpoint_count[match_prio] > 0) {
+            rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[match_prio] + 1;
             if (rand_k >= MAP_SIZE_OF_BACKEND) {
                 return -ENOENT;
             }
@@ -63,7 +63,7 @@ lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, s
             endpoint_v = map_lookup_endpoint(&endpoint_k);
             if (!endpoint_v) {
                 BPF_LOG(
-                    ERR, SERVICE, "find endpoint [%u/%u/%u] failed", service_id, match_rank, endpoint_k.backend_index);
+                    ERR, SERVICE, "find endpoint [%u/%u/%u] failed", service_id, match_prio, endpoint_k.backend_index);
                 return -ENOENT;
             }
             ret = endpoint_manager(kmesh_ctx, endpoint_v, service_id, service_v);
@@ -74,7 +74,7 @@ lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, s
             }
             return 0; // find the backend successfully
         }
-        if (is_strict && match_rank == service_v->lb_strict_index) { // only match lb strict index
+        if (is_strict && match_prio == service_v->lb_strict_prio) { // only match lb strict index
             return -ENOENT;
         }
     }
--- a/bpf/kmesh/workload/include/service.h
+++ b/bpf/kmesh/workload/include/service.h
@@ -61,9 +61,6 @@ lb_locality_failover_handle(struct kmesh_context *kmesh_ctx, __u32 service_id, s
         // if we have endpoints in this prio
         if (service_v->prio_endpoint_count[match_prio] > 0) {
             rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[match_prio] + 1;
-            if (rand_k >= MAP_SIZE_OF_BACKEND) {
-                return -ENOENT;
-            }
             endpoint_k.backend_index = rand_k;
             endpoint_v = map_lookup_endpoint(&endpoint_k);
             if (!endpoint_v) {
--- a/pkg/controller/workload/bpfcache/locality_cache.go
+++ b/pkg/controller/workload/bpfcache/locality_cache.go
@@ -1,3 +1,19 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package bpfcache
 
 import (
@@ -59,7 +75,7 @@ func (l *localityInfo) IsSet(param uint32) bool {
 type LocalityCache struct {
 	LbPolicy               uint32
 	localityInfo           localityInfo
-	LbStrictIndex          uint32 // for failover strict mode
+	LbStrictPrio           uint32 // for failover strict mode
 	isLocalityInfoSet      bool
 	RoutingPreference      []workloadapi.LoadBalancing_Scope
 	isRoutingPreferenceSet bool
@@ -92,7 +108,7 @@ func (l *LocalityCache) SetRoutingPreference(s []workloadapi.LoadBalancing_Scope
 	// notice: s should set by lb.GetRoutingPreference()
 	if len(s) > 0 {
 		l.RoutingPreference = s
-		l.LbStrictIndex = uint32(len(s))
+		l.LbStrictPrio = uint32(MinPrio - len(s))
 		l.isRoutingPreferenceSet = true
 	}
 }
@@ -138,7 +154,7 @@ func (l *LocalityCache) CalcuLocalityLBPrio(wl *workloadapi.Workload) uint32 {
 			}
 		}
 	}
-	return rank
+	return MinPrio - rank
 }
 
 func (l *LocalityCache) SaveToWaitQueue(wl *workloadapi.Workload) {
--- a/pkg/controller/workload/bpfcache/locality_cache.go
+++ b/pkg/controller/workload/bpfcache/locality_cache.go
@@ -116,7 +116,7 @@ func (l *LocalityCache) SetRoutingPreference(s []workloadapi.LoadBalancing_Scope
 	}
 }
 
-func (l *LocalityCache) CanLocalityLB() bool {
+func (l *LocalityCache) IsLocalityInfoSet() bool {
 	log.Debugf("isLocalityInfoSet: %#v, isRoutingPreferenceSet: %#v", l.isLocalityInfoSet, l.isRoutingPreferenceSet)
 	return l.isLocalityInfoSet && l.isRoutingPreferenceSet
 }
--- a/pkg/controller/workload/bpfcache/locality_cache.go
+++ b/pkg/controller/workload/bpfcache/locality_cache.go
@@ -1,0 +1,154 @@
+package bpfcache
+
+import (
+	"kmesh.net/kmesh/api/v2/workloadapi"
+)
+
+const (
+	REGION    = 1 << iota // 000001
+	ZONE                  // 000010
+	SUBZONE               // 000100
+	NODENAME              // 001000
+	CLUSTERID             // 010000
+	NETWORK               // 100000
+)
+
+type localityInfo struct {
+	region    string // init from workload.GetLocality().GetRegion()
+	zone      string // init from workload.GetLocality().GetZone()
+	subZone   string // init from workload.GetLocality().GetSubZone()
+	nodeName  string // init from os.Getenv("NODE_NAME"), workload.GetNode()
+	clusterId string // init from workload.GetClusterId()
+	network   string // workload.GetNetwork()
+	mask      uint32 // mask
+}
+
+func Valid(s string) bool {
+	return s != ""
+}
+
+func (l *localityInfo) Set(s string, param uint32) {
+	if !Valid(s) {
+		return
+	}
+	switch param {
+	case REGION:
+		l.region = s
+	case ZONE:
+		l.zone = s
+	case SUBZONE:
+		l.subZone = s
+	case NODENAME:
+		l.nodeName = s
+	case CLUSTERID:
+		l.clusterId = s
+	case NETWORK:
+		l.network = s
+	}
+	l.mask |= param
+}
+
+func (l *localityInfo) Clear(param uint32) {
+	l.mask &= ^param
+}
+
+func (l *localityInfo) IsSet(param uint32) bool {
+	return l.mask&param != 0
+}
+
+type LocalityCache struct {
+	LbPolicy               uint32
+	localityInfo           localityInfo
+	LbStrictIndex          uint32 // for failover strict mode
+	isLocalityInfoSet      bool
+	RoutingPreference      []workloadapi.LoadBalancing_Scope
+	isRoutingPreferenceSet bool
+	workloadWaitQueue      map[*workloadapi.Workload]struct{}
+}
+
+func NewLocalityCache() *LocalityCache {
+	return &LocalityCache{
+		localityInfo:           localityInfo{},
+		isLocalityInfoSet:      false,
+		RoutingPreference:      make([]workloadapi.LoadBalancing_Scope, 0),
+		isRoutingPreferenceSet: false,
+		workloadWaitQueue:      make(map[*workloadapi.Workload]struct{}),
+	}
+}
+
+func (l *LocalityCache) SetLocality(nodeName, clusterId, network string, locality *workloadapi.Locality) {
+	// notice: nodeName should set by processor or os.Getenv("NODE_NAME"),
+	l.localityInfo.Set(nodeName, NODENAME)
+	l.localityInfo.Set(locality.GetRegion(), REGION)
+	l.localityInfo.Set(locality.GetSubzone(), SUBZONE)
+	l.localityInfo.Set(locality.GetZone(), ZONE)
+	l.localityInfo.Set(clusterId, CLUSTERID)
+	l.localityInfo.Set(network, NETWORK)
+
+	l.isLocalityInfoSet = true
+}
+
+func (l *LocalityCache) SetRoutingPreference(s []workloadapi.LoadBalancing_Scope) {
+	// notice: s should set by lb.GetRoutingPreference()
+	if len(s) > 0 {
+		l.RoutingPreference = s
+		l.LbStrictIndex = uint32(len(s))
+		l.isRoutingPreferenceSet = true
+	}
+}
+
+func (l *LocalityCache) CanLocalityLB() bool {
+	log.Debugf("isLocalityInfoSet: %#v, isRoutingPreferenceSet: %#v", l.isLocalityInfoSet, l.isRoutingPreferenceSet)
+	return l.isLocalityInfoSet && l.isRoutingPreferenceSet
+}
+
+func (l *LocalityCache) CalcuLocalityLBPrio(wl *workloadapi.Workload) uint32 {
+	var rank uint32 = 0
+	for scope := range l.RoutingPreference {
+		switch scope {
+		case int(workloadapi.LoadBalancing_REGION):
+			log.Debugf("l.localityInfo.IsSet(REGION) %#v, Valid(wl.GetLocality().GetRegion()) %#v, l.localityInfo.region %#v, wl.GetLocality().GetRegion() %#v", l.localityInfo.IsSet(REGION), Valid(wl.GetLocality().GetRegion()), l.localityInfo.region, wl.GetLocality().GetRegion())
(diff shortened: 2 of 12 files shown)


pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,154 @@
...
+	l.localityInfo.Set(network, NETWORK)
+
+	l.isLocalityInfoSet = true
+}
+
+func (l *LocalityCache) SetRoutingPreference(s []workloadapi.LoadBalancing_Scope) {
+	// notice: s should set by lb.GetRoutingPreference()
+	if len(s) > 0 {
+		l.RoutingPreference = s
+		l.LbStrictIndex = uint32(len(s))
```
hzxuzhonghu, line 95: Not sure i understand this, previously i thought this is a flag.

pkg/controller/workload/workload_processor.go
```diff
@@ -179,6 +181,7 @@ func (p *Processor) removeWorkloadFromBpfMap(uid string) error {
 		wpkDelete = bpf.WorkloadPolicyKey{}
 	)
 
+	log.Warnf("== removeWorkloadFromBpfMap: workload uid: %#v, backendUid: %#v", uid, p.hashName.Hash(uid))
```
hzxuzhonghu, line 184: spammy, please use debug

pkg/controller/workload/workload_processor.go
```diff
@@ -275,13 +278,20 @@ func (p *Processor) removeServiceResourceFromBpfMap(svc *workloadapi.Service, na
 		}
 
 		var i uint32
-		for i = 1; i <= svDelete.EndpointCount; i++ {
-			ekDelete := bpf.EndpointKey{
-				ServiceId:    serviceId,
-				BackendIndex: i,
+		var j uint32
```
hzxuzhonghu, line 281: nit: you donot need to define j here, move below

for j:=0; j <= bpf.MaxPrio; j++ {

pkg/controller/workload/workload_processor.go
```diff
@@ -290,16 +300,17 @@ func (p *Processor) removeServiceResourceFromBpfMap(svc *workloadapi.Service, na
 }
 
 // addWorkloadToService update service & endpoint bpf map when a workload has new bound services
-func (p *Processor) addWorkloadToService(sk *bpf.ServiceKey, sv *bpf.ServiceValue, uid uint32) error {
+func (p *Processor) addWorkloadToService(sk *bpf.ServiceKey, sv *bpf.ServiceValue, uid uint32, Prio uint32) error {
```
hzxuzhonghu, line 303:
```suggestion
func (p *Processor) addWorkloadToService(sk *bpf.ServiceKey, sv *bpf.ServiceValue, uid uint32, priority uint32) error {
```

pkg/controller/workload/workload_processor.go
```diff
@@ -359,6 +382,7 @@ func (p *Processor) updateWorkload(workload *workloadapi.Workload) error {
 	)
 
 	uid := p.hashName.Hash(workload.GetUid())
+	log.Warnf("=in= updateWorkload: workload uid: %#v, backendUid: %#v", workload.GetUid(), uid)
```
hzxuzhonghu, line 385: nit: spammy

pkg/controller/workload/workload_processor.go
```diff
@@ -401,6 +425,11 @@ func (p *Processor) handleWorkload(workload *workloadapi.Workload) error {
 	p.WorkloadCache.AddOrUpdateWorkload(workload)
 	p.storeWorkloadPolicies(workload.GetUid(), workload.GetAuthorizationPolicies())
 
+	// update kmesh localityCache
+	if p.nodeName == workload.GetNode() {
+		p.locality.SetLocality(p.nodeName, workload.GetClusterId(), workload.GetNetwork(), workload.GetLocality())
```
hzxuzhonghu, line 430: not sure i understand why only setLocality when p.nodeName == workload.GetNode()

pkg/controller/workload/workload_processor.go
```diff
@@ -476,7 +505,12 @@ func (p *Processor) storeServiceData(serviceName string, waypoint *workloadapi.G
 	sk.ServiceId = p.hashName.Hash(serviceName)
 
 	newValue := bpf.ServiceValue{}
-	newValue.LbPolicy = LbPolicyRandom
+	newValue.LbPolicy = uint32(lb.GetMode()) // set loadbalance mode
+	p.locality.SetRoutingPreference(lb.GetRoutingPreference())
+	p.locality.LbPolicy = newValue.LbPolicy
+	log.Debugf("lbPolicy:%#v, routingPreference:%#v, strictIndex:%#v", newValue.LbPolicy, p.locality.RoutingPreference, p.locality.LbStrictIndex)
```
hzxuzhonghu, line 511:
```suggestion
	log.Debugf("lbPolicy:%v, routingPreference:%v, strictIndex:%v", newValue.LbPolicy, p.locality.RoutingPreference, p.locality.LbStrictIndex)
```

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,154 @@
...
+		l.isRoutingPreferenceSet = true
+	}
+}
+
+func (l *LocalityCache) CanLocalityLB() bool {
+	log.Debugf("isLocalityInfoSet: %#v, isRoutingPreferenceSet: %#v", l.isLocalityInfoSet, l.isRoutingPreferenceSet)
+	return l.isLocalityInfoSet && l.isRoutingPreferenceSet
+}
+
+func (l *LocalityCache) CalcuLocalityLBPrio(wl *workloadapi.Workload) uint32 {
```
hzxuzhonghu, line 105: not sure i understand this: From my perspective, the workload priority has to do with the original node info.

If the node and the workload both reside in same RoutingPreference [], the workload has highest priority.

hzxuzhonghu (commented):
I only looked at the go code part

pkg/controller/workload/bpfcache/locality_cache.go
derekwin: In strict mode, only the workloads that exactly match every item in the routePreference are considered. In other words, only the workloads with a priority equal to len(routePreference) are taken into account.　We set LbStrictIndex so that the kernel BPF program filters out priorities other than lbStrictIndex in strict mode.

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: Because the kmesh process retrieves information about all workloads from xDS, and the locality associated with the kmesh process should be carried by the workloads on the same node, we only update locality information from workloads that have the same nodeName as the current one.

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/bpfcache/locality_cache.go
derekwin: Yes, in the earliest design version, it was designed this way. During a kmesh community meeting, you suggested considering the current design format. Given that the kmesh process exists on every node as a DaemonSet, it is possible to pre-calculate the priority of each workload within a service relative to that node on each node. This way, when subsequent traffic arrives and accesses the service, the backend to serve can be selected based on the pre-calculated priorities of the workloads for that service.

derekwin:
After resolving the issue mentioned in https://github.com/kmesh-net/kmesh/issues/910, functional testing was conducted on the current code, and the functionality was found to be working correctly.

bpf/kmesh/workload/include/service.h
```diff
@@ -37,6 +41,47 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
...
+            }
+            ret = endpoint_manager(kmesh_ctx, endpoint_v, service_id, service_v);
+            if (ret != 0) {
+                if (ret != -ENOENT)
+                    BPF_LOG(ERR, SERVICE, "endpoint_manager failed, ret:%d\n", ret);
+                return ret;
+            }
+            return 0; // find the backend successfully
+        }
+        if (is_strict && match_prio == service_v->lb_strict_prio) { // only match lb strict index
```
hzxuzhonghu, line 77: can you explain a little bit more, i am confused about this

bpf/kmesh/workload/include/service.h
```diff
@@ -37,6 +41,47 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
...
+            return -ENOENT;
```
hzxuzhonghu, line 78: This is not align with ztunnel behavior, this return value would make the traffic goes through like directly accessing the clusterIP, but not blocking the traffic

hzxuzhonghu (commented):
Will try to review the control plane part later

pkg/status/status_server_test.go
```diff
@@ -388,7 +388,7 @@ func TestServer_dumpWorkloadBpfMap(t *testing.T) {
 			{ServiceId: 1}, {ServiceId: 2},
 		}
 		testServiceVals := []bpfcache.ServiceValue{
-			{EndpointCount: 1234}, {EndpointCount: 5678},
+			{EndpointCount: [7]uint32{1234, 1234, 1234, 1234, 1234, 1234, 1234}}, {EndpointCount: [7]uint32{5678, 5678, 5678, 5678, 5678, 5678, 5678}},
```
hzxuzhonghu, line 391: not sure i understand

pkg/controller/workload/workload_processor.go
```diff
@@ -179,6 +181,7 @@ func (p *Processor) removeWorkloadFromBpfMap(uid string) error {
 		wpkDelete = bpf.WorkloadPolicyKey{}
 	)
 
+	log.Debugf("removeWorkloadFromBpfMap: workload %s, backendUid %d", uid, p.hashName.Hash(uid))
```
hzxuzhonghu, line 184: nit: move after L185

pkg/controller/workload/workload_processor_test.go
hzxuzhonghu, line 1: We do need a test coverage for service lb policy update

pkg/controller/workload/workload_processor.go
```diff
@@ -341,9 +351,21 @@ func (p *Processor) handleWorkloadNewBoundServices(workload *workloadapi.Workloa
 		sk.ServiceId = svcUid
 		// the service already stored in map, add endpoint
 		if err = p.bpf.ServiceLookup(&sk, &sv); err == nil {
-			if err = p.addWorkloadToService(&sk, &sv, workloadId); err != nil {
-				log.Errorf("addWorkloadToService workload %d service %d failed: %v", workloadId, sk.ServiceId, err)
-				return err
+			if sv.LbPolicy == LbPolicyRandom { // random mode
+				if err = p.addWorkloadToService(&sk, &sv, workloadId, bpf.MinPrio); err != nil { // In random mode, we save all workload to minprio
```
hzxuzhonghu, line 355: I would suggest store with highest prio

bpf/kmesh/workload/include/service.h
```diff
@@ -17,9 +17,13 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
     int ret = 0;
     endpoint_key endpoint_k = {0};
     endpoint_value *endpoint_v = NULL;
+    int rand_k = 0;
 
     endpoint_k.service_id = service_id;
-    endpoint_k.backend_index = bpf_get_prandom_u32() % service_v->endpoint_count + 1;
+    endpoint_k.prio = MIN_PRIO; // for random handle，all endpoints are saved in MIN_PRIO
+
+    rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[MIN_PRIO] + 1;
```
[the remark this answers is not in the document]
hzxuzhonghu: short circuit

pkg/status/status_server_test.go
derekwin: I think this is just a dummy code.

pkg/controller/workload/workload_processor.go
```diff
@@ -341,9 +351,21 @@ func (p *Processor) handleWorkloadNewBoundServices(workload *workloadapi.Workloa
...
+					log.Errorf("addWorkloadToService workload %d service %d failed: %v", workloadId, sk.ServiceId, err)
+					return err
+				}
+			} else { // locality mode
+				if p.locality.CanLocalityLB() {
+					prio := p.locality.CalcuLocalityLBPrio(workload)
+					if err = p.addWorkloadToService(&sk, &sv, workloadId, prio); err != nil {
+						log.Errorf("addWorkloadToService workload %d service %d failed: %v", workloadId, sk.ServiceId, err)
```
YaoZengzeng, line 363:
```suggestion
						log.Errorf("addWorkloadToService workload %d service %d pirority %d failed: %v", workloadId, sk.ServiceId, prio, err)
```

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,170 @@
...
+			log.Debugf("l.localityInfo.IsSet(CLUSTERID) %#v, Valid(wl.GetClusterId()) %#v, l.localityInfo.clusterId %#v, wl.GetClusterId() %#v", l.localityInfo.IsSet(CLUSTERID), Valid(wl.GetClusterId()), l.localityInfo.clusterId, wl.GetClusterId())
+			if l.localityInfo.IsSet(CLUSTERID) && Valid(wl.GetClusterId()) && l.localityInfo.clusterId == wl.GetClusterId() {
+				rank++
+			}
+		}
+	}
+	return MinPrio - rank
+}
+
+func (l *LocalityCache) SaveToWaitQueue(wl *workloadapi.Workload) {
```
YaoZengzeng, line 160: Doesn't need a lock?

hzxuzhonghu:
This pr needs a big rework, can mark wip now

pkg/controller/workload/workload_processor.go
```diff
@@ -604,6 +711,13 @@ func (p *Processor) handleAddressTypeResponse(rsp *service_discovery_v3.DeltaDis
 		}
 	}
 
+	workloadInQueue := p.locality.GetFromWaitQueue() // locality LB mode
+	for workload := range workloadInQueue {
+		if err = p.handleWorkload(workload); err != nil {
+			log.Errorf("handle workload %s failed, err: %v", workload.ResourceName(), err)
```
hzxuzhonghu, line 717: looks a little unefficient, you can do this iteration only when locality info has been set.

And can we only call `handleWorkloadNewBoundServices`


where do you clear workloads from queue?

pkg/status/status_server_test.go
hzxuzhonghu: I mean in this test i donot see you set the routing preference, why expect 7 priority groups

bpf/kmesh/workload/include/service.h
```diff
@@ -17,9 +17,16 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
...
     endpoint_key endpoint_k = {0};
     endpoint_value *endpoint_v = NULL;
+    int rand_k = 0;
+
+    if (service_v->prio_endpoint_count[0] == 0)
+        return 0;
 
     endpoint_k.service_id = service_id;
-    endpoint_k.backend_index = bpf_get_prandom_u32() % service_v->endpoint_count + 1;
+    endpoint_k.prio = 0; // for random handle，all endpoints are saved in max priority
```
hzxuzhonghu, line 26:
```suggestion
    endpoint_k.prio = 0; // for random handle，all endpoints are saved with highest priority
```

derekwin:
not ready

bpf/kmesh/workload/include/service.h
```diff
@@ -37,6 +44,53 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
...
+    struct ip_addr zero_addr = {0};
+    __u32 zero_port = 0;
+
+    // #pragma unroll
+    for (int match_prio = 0; match_prio < PRIO_COUNT; match_prio++) {
+        endpoint_k.prio = match_prio; // 0->6
+        // if we have endpoints in this prio
+        if (service_v->prio_endpoint_count[match_prio] > 0) {
+            rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[match_prio] + 1;
+            if (rand_k >= MAP_SIZE_OF_BACKEND) {
```
hzxuzhonghu, line 64: how could this happen

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,178 @@
...
+	l.localityInfo.Set(locality.GetRegion(), REGION)
+	l.localityInfo.Set(locality.GetSubzone(), SUBZONE)
+	l.localityInfo.Set(locality.GetZone(), ZONE)
+	l.localityInfo.Set(clusterId, CLUSTERID)
+	l.localityInfo.Set(network, NETWORK)
+
+	l.isLocalityInfoSet = true
+}
+
+func (l *LocalityCache) SetRoutingPreference(s []workloadapi.LoadBalancing_Scope) {
```
hzxuzhonghu, line 109: I think we donot need this method,. we have already passed workload to `CalcuLocalityLBPrio`

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,178 @@
...
+	return l.isLocalityInfoSet && l.isRoutingPreferenceSet
+}
+
+func (l *LocalityCache) CalcuLocalityLBPrio(wl *workloadapi.Workload) uint32 {
+	var rank uint32 = 0
+	for scope := range l.RoutingPreference {
+		switch scope {
+		case int(workloadapi.LoadBalancing_REGION):
+			log.Debugf("l.localityInfo.IsSet(REGION) %#v, Valid(wl.GetLocality().GetRegion()) %#v, l.localityInfo.region %#v, wl.GetLocality().GetRegion() %#v", l.localityInfo.IsSet(REGION), Valid(wl.GetLocality().GetRegion()), l.localityInfo.region, wl.GetLocality().GetRegion())
+			if l.localityInfo.IsSet(REGION) && Valid(wl.GetLocality().GetRegion()) && l.localityInfo.region == wl.GetLocality().GetRegion() {
```
hzxuzhonghu, line 129:
```suggestion
			if l.localityInfo.region == wl.GetLocality().GetRegion() {
```

hzxuzhonghu: This looks not right, think of a case only `LoadBalancing_REGION` is set, if the workload has same region with local node, the rank(priority) should be 0, otherwise 1

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,178 @@
...
+func (l *LocalityCache) SetRoutingPreference(s []workloadapi.LoadBalancing_Scope) {
+	// notice: s should set by lb.GetRoutingPreference()
+	if len(s) > 0 {
+		l.RoutingPreference = s
+		l.isRoutingPreferenceSet = true
+		l.routePreferenceCount = uint32(len(s))
+	}
+}
+
+func (l *LocalityCache) CanLocalityLB() bool {
```
hzxuzhonghu, line 118: please rename: suggest IsLocalityInfoSet

Why do you need `isRoutingPreferenceSet`, whose RoutingPreference is it?

pkg/controller/workload/bpfcache/locality_cache.go
derekwin: This funcion is for saving localityInfo to kmesh processor with `locality`. We need kmesh processor hold the localityInfo(because it represents the localityInfo which Node is kmesh processor belongs to) so that we can calculate the prio between the kmesh processor and upcoming workload.

pkg/controller/workload/bpfcache/locality_cache.go
derekwin: We should only deal with fields that have been setup successfully. The default value of locality such as Region is "". If we don't check weather the value has been set, We will get rank=1 while workload comes with unset resion filed.  Also, If the Node is label with locality, the upcomming workload will have this fileds set, and `l.localityInfo.IsSet(REGION)` is always been set.

pkg/controller/workload/bpfcache/locality_cache.go
derekwin: I got it and will remove it from locality_cache, handle related logic by service cache.

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,178 @@
...
+	log.Debugf("isLocalityInfoSet: %#v, isRoutingPreferenceSet: %#v", l.isLocalityInfoSet, l.isRoutingPreferenceSet)
+	return l.isLocalityInfoSet && l.isRoutingPreferenceSet
+}
+
+func (l *LocalityCache) CalcuLocalityLBPrio(wl *workloadapi.Workload) uint32 {
```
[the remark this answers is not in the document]
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: ok

bpf/kmesh/workload/include/service.h
derekwin: If left unrestricted, the BPF program may exhibit anomalies (the exact cause of errors has not been observed yet, but adding this restriction is necessary).

derekwin:
I just submitted the latest version of the code, which fixes the bugs that existed before.

derekwin:
**Special note on the redesign of `deleteEndpointRecords`:** 
The original `deleteEndpointRecords` function in kMesh had a flaw, as it was only suitable for deleting multiple endpoints corresponding to a single workload (these endpoints must belong to different services). This is because when `p.bpf.EndpointSwap` runs, it changes the `BackendIndex` of the endpoint in the BPF map. When trying to delete multiple endpoints that all belong to the same service, the `BackendIndex` dynamically changes, leading to errors. This commit redesigns `deleteEndpointRecords`, adding logic to dynamically maintain the `BackendIndex`, thus avoiding the aforementioned issue.
Additionally, passing the `workloadId` to this function was unnecessary, so it has been removed.

hzxuzhonghu:
> When trying to delete multiple endpoints that all belong to the same service, the BackendIndex dynamically changes, leading to errors.

How could we delete endpoints belong to same service? The caller is actually deleting a workload

pkg/controller/workload/bpfcache/service.go
```diff
@@ -32,9 +32,9 @@ type ServicePorts [MaxPortNum]uint32
 type TargetPorts [MaxPortNum]uint32
 
 type ServiceValue struct {
-	EndpointCount uint32       // endpoint count of current service
-	LbPolicy      uint32       // load balancing algorithm, currently only supports random algorithm
-	ServicePort   ServicePorts // ServicePort[i] and TargetPort[i] are a pair, i starts from 0 and max value is MaxPortNum-1
+	EndpointCount [PrioCount]uint32 // endpoint count of current service
```
hzxuzhonghu, line 35: map

pkg/controller/workload/workload_processor.go
```diff
@@ -57,6 +56,8 @@ type Processor struct {
 	nodeName      string
 	WorkloadCache cache.WorkloadCache
 	ServiceCache  cache.ServiceCache
+	EndpointCache cache.EndpointCache
```
hzxuzhonghu, line 59: why do you add it back>?

pkg/controller/workload/workload_processor.go
```diff
@@ -341,9 +352,22 @@ func (p *Processor) handleWorkloadNewBoundServices(workload *workloadapi.Workloa
...
-				return err
+			if sv.LbPolicy == uint32(workloadapi.LoadBalancing_UNSPECIFIED_MODE) { // random mode
+				if err, _ = p.addWorkloadToService(&sk, &sv, workloadId, 0); err != nil { // In random mode, we save all workload to max priority
+					log.Errorf("addWorkloadToService workload %d service %d failed: %v", workloadId, sk.ServiceId, err)
+					return err
+				}
+			} else { // locality mode
+				if p.locality.IsLocalityInfoSet() {
+					service := p.ServiceCache.GetService(p.hashName.NumToStr(svcUid))
+					prio := p.locality.CalcuLocalityLBPrio(workload, service.LoadBalancing.GetRoutingPreference())
```
hzxuzhonghu, line 363: nit CalcLBPrio

hzxuzhonghu: if the service is nill, this will panic

pkg/controller/workload/workload_processor.go
```diff
@@ -501,8 +580,42 @@ func (p *Processor) storeServiceData(serviceName string, waypoint *workloadapi.G
...
+			// there might not be any workload with the highest priority, and directly switching the service's LB policy could
+			// lead to unexpected service disruptions.
+			if err = p.updateEndpoint(sk.ServiceId, false); err != nil { // this will change bpf map totally
+				log.Errorf("UpdateEndpoint failed, err:%s", err)
+			}
+			updateValue := bpf.ServiceValue{}
+			if err = p.bpf.ServiceLookup(&sk, &updateValue); err != nil {
+				log.Warnf("Lookup Service failed, err:%s", err)
+			}
+			updateValue.LbPolicy = newValue.LbPolicy
```
hzxuzhonghu, line 597: donot we need to update the EndpointCount?

pkg/controller/workload/workload_processor.go
```diff
@@ -501,8 +580,42 @@ func (p *Processor) storeServiceData(serviceName string, waypoint *workloadapi.G
...
+			if err = p.bpf.ServiceUpdate(&sk, &updateValue); err != nil {
+				log.Errorf("Update Service failed, err:%s", err)
+			}
+			return nil
+		} else if oldValue.LbPolicy == uint32(workloadapi.LoadBalancing_UNSPECIFIED_MODE) { // from random to locality loadbalance
+			// In random mode, the workloads are stored with the highest priority. When switching from random mode to locality
+			// load balancing, we first update the service map to quickly initiate the transition of the strategy. Subsequently,
+			// we update the endpoint map. During this update process, the load balancer may briefly exhibit abnormal random behavior,
+			// after which it will fully transition to the locality load balancing mode.
+			if err = p.bpf.ServiceUpdate(&sk, &newValue); err != nil { // update service first
```
hzxuzhonghu, line 607: ditto EndpointCount not updated?

pkg/controller/workload/workload_processor.go
```diff
@@ -466,7 +496,55 @@ func (p *Processor) storeServiceFrontendData(serviceId uint32, service *workload
 	return nil
 }
 
-func (p *Processor) storeServiceData(serviceName string, waypoint *workloadapi.GatewayAddress, ports []*workloadapi.Port) error {
+func (p *Processor) updateEndpoint(serviceId uint32, toLLb bool) error {
```
hzxuzhonghu, line 499: please add some comments  on `toLLb `

pkg/controller/workload/workload_processor.go
```diff
@@ -466,7 +496,55 @@ func (p *Processor) storeServiceFrontendData(serviceId uint32, service *workload
...
+			BackendIndex: ep.BackendIndex,
+		}
+		ev := bpf.EndpointValue{}
+		if err := p.bpf.EndpointLookup(&ek, &ev); err != nil { // get backend Uid
+			return err
+		}
+		backendUids = append(backendUids, ev.BackendUid)
+		eksDelete = append(eksDelete, ek)
+	}
+	if err := p.deleteEndpointRecords(eksDelete); err != nil {
```
hzxuzhonghu, line 519: I am concerned about this may cause outage

pkg/controller/workload/workload_processor.go
```diff
@@ -466,7 +496,55 @@ func (p *Processor) storeServiceFrontendData(serviceId uint32, service *workload
...
+		log.Errorf("delete [%#v] from endpoint map failed: %s", eksDelete, err)
+	}
+
+	sKey := bpf.ServiceKey{ServiceId: serviceId}
+	sValue := bpf.ServiceValue{}
+	for _, uid := range backendUids {
```
hzxuzhonghu, line 525: Not sure if there is a more efficient update

hzxuzhonghu (changes requested):
Please add some tests

derekwin:
> > When trying to delete multiple endpoints that all belong to the same service, the BackendIndex dynamically changes, leading to errors.
> 
> How could we delete endpoints belong to same service? The caller is actually deleting a workload

When updating a service, it's likely that the associated endpoints will be deleted in batches. Although currently the updates to the LbPolicy are implemented by deleting them one by one, we should not rule out the possibility of batch deletions in the future. Additionally, the function name deleteEndpointRecords suggests a general operation and should have a more robust implementation.

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,164 @@
...
+
+	l.isLocalityInfoSet = true
+}
+
+func (l *LocalityCache) IsLocalityInfoSet() bool {
+	log.Debugf("isLocalityInfoSet: %#v", l.isLocalityInfoSet)
+	return l.isLocalityInfoSet
+}
+
+func (l *LocalityCache) CalcuLocalityLBPrio(wl *workloadapi.Workload, rp []workloadapi.LoadBalancing_Scope) uint32 {
```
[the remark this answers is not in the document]
derekwin: rp is included in WorkloadService by filed *LoadBalancing.

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (l *localityInfo) IsSet(param uint32) bool {
+	return l.mask&param != 0
+}
+
+type LocalityCache struct {
+	mutex             sync.RWMutex
+	localityInfo      localityInfo
+	isLocalityInfoSet bool
```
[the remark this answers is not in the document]
derekwin: ok

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,164 @@
...
+	l.localityInfo.Set(locality.GetSubzone(), SUBZONE)
+	l.localityInfo.Set(locality.GetZone(), ZONE)
+	l.localityInfo.Set(clusterId, CLUSTERID)
+	l.localityInfo.Set(network, NETWORK)
+
+	l.isLocalityInfoSet = true
+}
+
+func (l *LocalityCache) IsLocalityInfoSet() bool {
+	log.Debugf("isLocalityInfoSet: %#v", l.isLocalityInfoSet)
```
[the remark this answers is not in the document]
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/workload_processor.go
derekwin: The EndpointCount is already the newest.
In the current implementation, p.updateEndpoint synchronously maintains the state of the bpfmap to the latest when updating an endpoint. We should maintain the independence of the updateEndpoint function in managing the bpfmap state (since we currently lack a mechanism to ensure consistency between user space and kernel space, which we might consider implementing later). If the serviceValue passed internally and externally to the updateEndpoint function is the same, it could lead to complex state changes during the update process, increasing the risk of state confusion errors.

pkg/controller/workload/workload_processor.go
derekwin: when we changing lb policy, its means that we are in service updating, so the total count number of oldValue and new Value is same.

pkg/controller/workload/workload_processor.go
derekwin: ok

pkg/controller/workload/bpfcache/service.go
derekwin: If using map here, how to update this value to bpf map? maintain a new bpf map stuct?

pkg/controller/workload/workload_processor.go
derekwin: I have using *cache.EndpointCache before, and found it may occur some segament error, so i decided to use stuct rather than pointer of it. EndpointCache is a new cache type introduced this time.

pkg/controller/workload/workload_processor.go
derekwin: deleteEndpointRecords maintains newest bpf map state in it.

pkg/controller/workload/workload_processor.go
derekwin: We can lookup sValue only once time. I have updated it.

bpf/kmesh/workload/include/service.h
```diff
@@ -37,6 +41,47 @@ static inline int lb_random_handle(struct kmesh_context *kmesh_ctx, __u32 servic
...
+    endpoint_value *endpoint_v = NULL;
+    endpoint_k.service_id = service_id;
+
+    // #pragma unroll
+    for (int match_prio = 0; match_prio < PRIO_COUNT; match_prio++) {
+        endpoint_k.prio = match_prio; // 0->6
+        // if we have endpoints in this prio
+        if (service_v->prio_endpoint_count[match_prio] > 0) {
+            rand_k = bpf_get_prandom_u32() % service_v->prio_endpoint_count[match_prio] + 1;
+            if (rand_k >= MAP_SIZE_OF_BACKEND) {
```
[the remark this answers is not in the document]
derekwin: removed it

bpf/kmesh/workload/include/service.h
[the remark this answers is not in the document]
derekwin: Yes, the current EndpointSwap implementation relies on the index starting from 1. If we change it to start from 0, EndpointSwap would encounter issues. I believe this can be addressed as a separate optimization point, or we could simply keep the index starting from 1.

derekwin:
> Please add some tests

TestLBPolicyUpdate and TestCalcLocalityLbPrio have been added in workload_processor_test.go

pkg/controller/workload/bpfcache/locality_cache.go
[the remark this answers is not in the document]
hzxuzhonghu: right

hzxuzhonghu:
>>When trying to delete multiple endpoints that all belong to the same service, the BackendIndex dynamically changes, leading to errors.

>>How could we delete endpoints belong to same service? The caller is actually deleting a workload

>When updating a service, it's likely that the associated endpoints will be deleted in batches. we should not rule out the possibility of batch deletions in the future. Additionally, the function name deleteEndpointRecords suggests a general operation and should have a more robust implementation.

How are the endpoints be deleted in batch? I cannot understand

pkg/controller/workload/bpfcache/locality_cache.go
```diff
@@ -0,0 +1,141 @@
...
+				rank++
+			}
+		case workloadapi.LoadBalancing_CLUSTER:
+			log.Debugf("l.LocalityInfo.clusterId %#v, wl.GetClusterId() %#v", l.LocalityInfo.clusterId, wl.GetClusterId())
+			if l.LocalityInfo.clusterId == wl.GetClusterId() {
+				rank++
+			}
+		}
+	}
+	return uint32(len(rp)) - rank
```
hzxuzhonghu, line 140: This method is not right. Keep in mind, we shoud match rp []workloadapi.LoadBalancing_Scope in order, if one doesnot match, we should stop.   


So i am super sure this result now is nort right.

derekwin: fixed it

derekwin:
> > > When trying to delete multiple endpoints that all belong to the same service, the BackendIndex dynamically changes, leading to errors.
> 
> > > How could we delete endpoints belong to same service? The caller is actually deleting a workload
> 
> > When updating a service, it's likely that the associated endpoints will be deleted in batches. we should not rule out the possibility of batch deletions in the future. Additionally, the function name deleteEndpointRecords suggests a general operation and should have a more robust implementation.
> 
> How are the endpoints be deleted in batch? I cannot understand

In function `func (p *Processor) updateEndpoint(serviceId uint32, toLLb bool)` [line 500], we get all endpoints needed to be deleted and delete them in batch using `deleteEndpointRecords` [line 520].
p.s. In the previous meeting, we agreed on the plan to update the endpoints one by one. However, during the subsequent implementation, I found that deleting one endpoint causes the indices of other endpoints to change, leading to confusion during deletion. Therefore, I believe we should delete all the endpoints to be removed at once and then re-add each new endpoint.

pkg/controller/workload/bpfcache/locality_cache.go
hzxuzhonghu: There is not test coverage, i will write a test later to test your function.  I am feeling this function is still not correct

pkg/controller/workload/workload_processor.go
```diff
@@ -494,17 +496,15 @@ func (p *Processor) storeServiceFrontendData(serviceId uint32, service *workload
...
-// it represents a random strategy, in which case we just set the priority to 0.
-func (p *Processor) updateEndpoint(serviceId uint32, toLLb bool) error {
-	var (
-		err  error
-		prio uint32
-	)
-	eksDelete := []bpf.EndpointKey{}
-	backendUids := []uint32{}
-	for _, ep := range p.EndpointCache.List(serviceId) {
+func (p *Processor) updateEndpointOneByOne(serviceId uint32, epsDelete []cache.Endpoint, toLLb bool) error {
```
hzxuzhonghu, line 499: `epsDelete` may look confusing, indeed we are trying to re orgnize them

hzxuzhonghu (dismissed):
LGTM.

pkg/controller/workload/workload_processor.go
```diff
@@ -494,17 +496,15 @@ func (p *Processor) storeServiceFrontendData(serviceId uint32, service *workload
...
+	var prio uint32
+
+	if len(epsDelete) == 0 {
+		return nil
+	}
+
+	for i := len(epsDelete) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
derekwin: updated it

pkg/controller/workload/workload_processor.go
derekwin: using `epsUpdate` now

hzxuzhonghu (approved):
/lgtm

Issue #821, opened 2024-09-06
Kmesh cli tool


YaoZengzeng:
**What would you like to be added**:

Kmesh cli tool called `kmeshctl`, the commands initially included are as follows:

- [] kmeshctl install/uninstall: to install/uninstall Kmesh #552
- [] kmeshctl config-dump: to dump config cache in Kmesh daemon
- [] kmeshctl bpf: to output the underlying bpf config corresponding to the Kmesh daemon in a human-readable way #820 
- [] kmeshctl log: adjust Kmesh log level
- [] kmeshctl waypoint: to config waypoint for namespace/service/pods

to be added later on demand.

**Why is this needed**:

`kmeshctl` is a great tool to for users to install, configure, observe the running status of Kmesh and debug.

PR #901, opened 2024-09-26
`kmeshctl waypoint` framework

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

Implement `kmeshctl waypoint` subcommand, so we could use it to better manage waypoints.

Fixes part of #821

```release-note

```

### DIFF
--- a/ctl/waypoint/waypoint.go
+++ b/ctl/waypoint/waypoint.go
@@ -43,7 +43,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	gateway "sigs.k8s.io/gateway-api/apis/v1"
 
 	"kmesh.net/kmesh/ctl/utils"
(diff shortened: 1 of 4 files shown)


ctl/waypoint/waypoint.go
```diff
@@ -0,0 +1,618 @@
...
+
+	trafficType       = ""
+	validTrafficTypes = sets.New(constants.ServiceTraffic, constants.WorkloadTraffic, constants.AllTraffic, constants.NoTraffic)
+
+	DataplaneModeKmesh               = "Kmesh"
+	KmeshUseWaypointLabel            = "istio.io/use-waypoint"
+	KmeshWaypointForTrafficTypeLabel = "istio.io/waypoint-for"
+
+	WaypointImageAnnotation = "sidecar.istio.io/proxyImage"
+	KmeshWaypointImage      = "ghcr.io/kmesh-net/waypoint:latest"
```
LiZhenCheng9527, line 68: nit：If it in 0.5.0 release, is waypoint version also fixed to 0.5.0?

YaoZengzeng: Good catch!

It should be the same as the release version, but it should also be configurable, because we prefer to use the latest image during dev.

YaoZengzeng:
@hzxuzhonghu @LiZhenCheng9527 PTAL

ctl/waypoint/waypoint.go
```diff
@@ -0,0 +1,628 @@
...
+}
+
+func getKmeshWaypointImage() string {
+	if image != "" {
+		return image
+	}
+
+	ver := version.Get().GitVersion
+
+	return fmt.Sprintf("ghcr.io/kmesh-net/waypoint:v%s", ver)
```
hzxuzhonghu, line 627: `ghcr.io/kmesh-net/waypoint` can be customized

YaoZengzeng: You could use `--image ` flag to customize waypoint image.

YaoZengzeng:
@hzxuzhonghu updated

hzxuzhonghu (approved):

PR #918, opened 2024-09-30
add xdp authz proposal

supercharge-xsy:
**What type of PR is this?**

Fixes #

```release-note

```


docs/proposal/layer4_authorization.md
```diff
@@ -56,5 +54,32 @@ struct {
...
+map_of_authz: records the authz rules of policies.
 
+```.c
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(key_size, sizeof(__u32));
+    __uint(value_size, sizeof(wl_policies_v));
+    __uint(map_flags, BPF_F_NO_PREALLOC);
+    __uint(max_entries, MAP_SIZE_OF_AUTH_POLICY);
+}    SEC(".maps");
```
hzxuzhonghu, line 76: no name present

docs/proposal/layer4_authorization.md
```diff
@@ -56,5 +54,32 @@ struct {
    - 3.2: If the matched record shows `value=deny`, it alters the message flag, sends an RST message to the server, clears the corresponding `auth_map` record. If no record is matched, implying authorization is allowed, the message is passed through.
 4. **client retry**: The client attempts to send another message, but because the server has closed the connection, the client receives a "reset by peer" signal and subsequently closes its own channel.
 
+### Xdp authentication
+
+#### Design detail
+
+![l4_authz_xdp](pics/kmesh_l4_authorization_xdp.svg#pic_center)
```
LiZhenCheng9527, line 61: The control plane is istiod.

docs/proposal/layer4_authorization.md
```diff
@@ -56,5 +54,32 @@ struct {
...
+}    SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(key_size, sizeof(__u32));
+    __uint(value_size, sizeof(Istio__Security__Authorization));
+    __uint(map_flags, BPF_F_NO_PREALLOC);
+    __uint(max_entries, MAP_SIZE_OF_AUTH_POLICY);
+} map_of_authz SEC(".maps");
+```
```
YaoZengzeng, line 85: No more processing details?

bitcoffeeiux, line 85: Why xdp program return XDP_DROP? 
This causes packet loss and timeout errors on the client.

hzxuzhonghu: Any other good suggestion?

bitcoffeeiux: > Any other good suggestion?

There doesn't seem to be a particularly good solution to replace the timeout here.
XDP is a packet-base handler, you can only interfere with the connection by modifying the packet state.
But the connection status can reflect very little information.
Only one RST is returned to the peer end when the connection is terminated. The peer end, however, does not know the cause of the RST.

hzxuzhonghu: Yeah, how can we make it rest fast instead of waiting for timeout now

docs/proposal/layer4_authorization.md
```diff
@@ -56,5 +54,39 @@ struct {
...
+
+#### Processing logic
 
+![l4_authz_xdp](pics/kmesh_l4_authorization_match_chain.svg#pic_center)
 
+1. Message Parsing: When a packet enters the XDP processing logic on the server side, the tuple information of the packet is parsed. The corresponding workload instance is then located based on the destination IP, and the authentication rules configured on that workload are retrieved.
+2. Rule Matching: As shown in the figure, XDP implements a matching chain logic. First, it determines whether to allow or deny the packet based on the port  info, if the result is deny, the packet is intercepted, and the process ends. If the result is allow, the next matching logic (e.g., IP matching) is called using a BPF tail call. This process is repeated until the last link in the chain. If the final result is allow, XDP\_PASS is returned, and the packet is forwarded through the kernel network stack to the server.
```
hzxuzhonghu, line 92: As i talked with @weli-l the tailcall mechanism seems not feasible, because a workload may have several authz policies set, tailcall does not return.

hzxuzhonghu, line 92:
```suggestion
2. Rule Matching: As shown in the figure, XDP implements a matching chain logic. First, it determines whether to allow or deny the packet based on the port  info, if the result is deny, the packet is intercepted, and the process ends. If the result is allow, the next matching logic (e.g., IP matching) is called using a BPF tail call. This process is repeated until the last link in the chain. If the final result is allow, XDP_PASS is returned, and the packet is forwarded through the kernel network stack to the server.
```

docs/proposal/layer4_authorization.md
supercharge-xsy: fix

PR #935, opened 2024-10-08
enable/disable accesslog with kmeshctl

### LINKED ISSUE #925: Starting and shutting down accesslog with kmeshctl

**Why is this needed**:
The startup and shutdown of accesslog is now determined by the startup parameters. We should optimise it by passing parameters via kmeshctl.

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Fixes #925 

```release-note

```

### DIFF
--- a/ctl/accesslog/accesslog.go
+++ b/ctl/accesslog/accesslog.go
@@ -1,0 +1,124 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accesslog
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/spf13/cobra"
+	"istio.io/istio/pkg/kube"
+
+	"kmesh.net/kmesh/ctl/utils"
+	"kmesh.net/kmesh/pkg/logger"
+)
+
+const (
+	patternAccesslog = "/accesslog"
+)
+
+var log = logger.NewLoggerScope("kmeshctl/accesslog")
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "accesslog",
+		Short: "Enable or disable Kmesh's accesslog",
+		Example: `# Enable Kmesh's accesslog:
+kmeshctl accesslog <kmesh-daemon-pod> enable
+
+# Disable Kmesh's accesslog:
+kmeshctl accesslog <kmesh-daemon-pod> disable`,
+		Args: cobra.MinimumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			SetAccesslog(cmd, args)
+		},
+	}
+	return cmd
+}
+
+func SetAccesslog(cmd *cobra.Command, args []string) {
+	var info string
+	accesslogFlag := args[len(args)-1]
+	if accesslogFlag == "enable" {
+		info = "true"
+	} else if accesslogFlag == "disable" {
+		info = "false"
+	} else {
+		log.Errorf("Error: Argument must be 'enable' or 'disable'")
+		os.Exit(1)
+	}
+
+	cli, err := utils.CreateKubeClient()
+	if err != nil {
+		log.Errorf("failed to create cli client: %v", err)
+		os.Exit(1)
+	}
+
+	if len(args) == 1 {
+		// Perform operations on all kmesh daemons.
+		podList, err := cli.PodsForSelector(context.TODO(), utils.KmeshNamespace, utils.KmeshLabel)
+		if err != nil {
+			log.Errorf("failed to get kmesh podList: %v", err)
+			os.Exit(1)
+		}
+		for _, pod := range podList.Items {
+			SetAccesslogPerKmeshDaemon(cli, pod.GetName(), info)
+		}
+	} else {
+		// Processes accesslog triggers for specified kmesh daemon.
+		for _, podname := range args[:len(args)-1] {
+			SetAccesslogPerKmeshDaemon(cli, podname, info)
+		}
+	}
+}
+
+func SetAccesslogPerKmeshDaemon(cli kube.CLIClient, podName, info string) {
+	fw, err := utils.CreateKmeshPortForwarder(cli, podName)
+	if err != nil {
+		log.Errorf("failed to create port forwarder for Kmesh daemon pod %s: %v", podName, err)
+		os.Exit(1)
+	}
+	if err := fw.Start(); err != nil {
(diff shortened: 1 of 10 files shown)


pkg/status/status_server.go
```diff
@@ -236,6 +238,30 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (s *Server) accesslogHandler(w http.ResponseWriter, r *http.Request) {
+	var accesslogInfo string
+
+	defer r.Body.Close()
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "\t%s: %v\n", "Error reading request body", err)
```
YaoZengzeng, line 248:
```suggestion
		fmt.Fprintf(w, "\tError reading request body: %v\n", err)
```

ctl/accesslog/accesslog.go
```diff
@@ -0,0 +1,105 @@
...
+	patternAccesslog = "/accesslog"
+)
+
+var log = logger.NewLoggerScope("kmeshctl/accesslog")
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "accesslog",
+		Short:   "enable or disable Kmesh's accesslog",
+		Example: `# enable Kmesh's accesslog: kmeshctl accesslog <kmesh-daemon-pod> enable`,
```
YaoZengzeng, line 43: According to the code, `enable` should be `true`?

YaoZengzeng: Also the format should be consistent wiht other subcommands, including line breaks, upper and lower case, etc.

LiZhenCheng9527: This should be `true`. I will fix it.

ctl/accesslog/accesslog.go
```diff
@@ -0,0 +1,105 @@
...
+
+	data, err := json.Marshal(accesslogFlag)
+	if err != nil {
+		log.Errorf("Error marshaling accesslog info: %v", err)
+		return
+	}
+
+	url := fmt.Sprintf("http://%s%s", fw.Address(), patternAccesslog)
+
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(data))
```
YaoZengzeng, line 79: If there is only one bool flag, can it be placed directly in the URL, like `http://../accesslog?enable=true`?

Putting it in the body is more cubersome.

LiZhenCheng9527: Done

ctl/accesslog/accesslog.go
```diff
@@ -0,0 +1,90 @@
...
+		log.Errorf("failed to create port forwarder for Kmesh daemon pod %s: %v", podName, err)
+		os.Exit(1)
+	}
+	if err := fw.Start(); err != nil {
+		log.Errorf("failed to start port forwarder for Kmesh daemon pod %s: %v", podName, err)
+		os.Exit(1)
+	}
+	defer fw.Close()
+
+	url := fmt.Sprintf("http://%s%s?enable=%s", fw.Address(), patternAccesslog, accesslogFlag)
```
YaoZengzeng, line 69: "http://%s%s?enable=enable" or "http://%s%s?enable=disable"? 

Looks weird, it's better to turn accesslogFlag to `true` or `false`, and change the url to "http://%s%s?enable=true" or "http://%s%s?enable=false"?

pkg/status/status_server.go
```diff
@@ -236,6 +238,18 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) accesslogHandler(w http.ResponseWriter, r *http.Request) {
```
hzxuzhonghu, line 241: we should limit only POST is supported here

pkg/status/status_server.go
```diff
@@ -236,6 +238,18 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
...
+	accesslogInfo := r.URL.Query().Get("enable")
+
+	if accesslogInfo == "enable" {
+		s.xdsClient.WorkloadController.MetricController.EnableAccesslog = true
+	} else {
+		s.xdsClient.WorkloadController.MetricController.EnableAccesslog = false
```
hzxuzhonghu, line 247: mutating a var which is being used by another routine can cause data race in golang

pkg/status/status_server.go
LiZhenCheng9527: Done

pkg/status/status_server.go
LiZhenCheng9527: Add a lock protection for data race

ctl/accesslog/accesslog.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+var log = logger.NewLoggerScope("kmeshctl/accesslog")
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "accesslog",
+		Short: "Enable or disable Kmesh's accesslog",
+		Example: `# Enable Kmesh's accesslog:
+kmeshctl accesslog <kmesh-daemon-pod> enable`,
```
hzxuzhonghu, line 41: I would think we can enable/disable accesslog for all nodes, this is the normal use case

hzxuzhonghu: And give a example for disable

pkg/controller/telemetry/metric.go
```diff
@@ -50,13 +50,18 @@ const (
 var osStartTime time.Time
 
 type MetricController struct {
-	enableAccesslog     bool
+	EnableAccesslog     enableTrigger
```
hzxuzhonghu, line 53: can use atomic.Bool type

pkg/status/status_server.go
```diff
@@ -236,6 +238,23 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	var info bool
+	accesslogInfo := r.URL.Query().Get("enable")
+	if accesslogInfo == "true" {
+		info = true
+	} else {
+		info = false
+	}
+	s.xdsClient.WorkloadController.MetricController.EnableAccesslog.SetAccesslogTrigger(info)
```
hzxuzhonghu, line 254: The calling stack is too deep, why not provide a method to `MetricController` or even to `WorkloadController`

ctl/accesslog/accesslog.go
LiZhenCheng9527: Done

ctl/accesslog/accesslog.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"kmesh.net/kmesh/ctl/utils"
+	"kmesh.net/kmesh/pkg/logger"
+)
+
+const (
+	patternAccesslog = "/accesslog"
+)
+
+var log = logger.NewLoggerScope("kmeshctl/accesslog")
```
hzxuzhonghu, line 36: Everytime we use a logger we need to init one, i do think we can provide a default logger in the package. So in some common place, it can be used

ctl/utils/utils.go
```diff
@@ -55,3 +41,18 @@ func CreateKubeClient() (kube.CLIClient, error) {
 
 	return cli, nil
 }
+
+// Create a new PortForwarder configured for the given Kmesh daemon pod.
+func CreateKmeshPortForwarder(cliClient kube.CLIClient, podName string) (kube.PortForwarder, error) {
+	// cli, err := CreateKubeClient()
```
hzxuzhonghu, line 47: remove

Issue #944, opened 2024-10-11
[release 0.5]Binary in Kmesh arm image not exec properly


LiZhenCheng9527:
**What happened**:
```sh
./start_kmesh.sh: line 29: /usr/bin/kmesh-daemon: cannot execute binary file: Exec format error
./start_kmesh.sh: line 54: /usr/bin/kmesh-daemon: cannot execute binary file: Exec format error
```
**What you expected to happen**:
Can use Kmesh in an arm architecture OS.
**How to reproduce it (as minimally and precisely as possible)**:
The release 0.5.0 image was built via github action. We may need to validate workflow further.
**Anything else we need to know?**:

**Environment**:
- Kmesh version: 0.5.0
- Others:

hzxuzhonghu:
what os did you run that?

LiZhenCheng9527:
oe2303

lec-bit:
This issue exists in latest, too.

PR #742, opened 2024-08-16
modify dockerfile

weli-l:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #948, opened 2024-10-12
Fix github action arm image built

hzxuzhonghu:
**What type of PR is this?**

: without this, though the docker image is ar,64 arch but  the kmesh-daemon within it is just x86-64.

Fixes https://github.com/kmesh-net/kmesh/issues/944

```release-note

```

### DIFF
--- a/build/docker/github.dockerfile
+++ b/build/docker/github.dockerfile
@@ -1,0 +1,26 @@
+FROM ghcr.io/kmesh-net/kmesh-build:latest AS builder
+COPY . /kmesh
+WORKDIR /kmesh
+RUN ./build.sh && ./build.sh -i
+# Copy the kmesh.ko to /kmesh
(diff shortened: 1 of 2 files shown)


build/docker/github.dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ghcr.io/kmesh-net/kmesh-build:latest AS builder
```
LiZhenCheng9527, line 1: Why need a new dockerfile?Why not modify the original dockerfile?

hzxuzhonghu: This is a temparary fix. Becase that @weli-l found before https://github.com/kmesh-net/kmesh/pull/742, it failed to build using docker 18.09

hzxuzhonghu:
I will merge this manually, since this is influcencing arm deploy

PR #958, opened 2024-10-15
check whether pods should be managed when ns is managed

### LINKED ISSUE #956: check whether pods should be managed when ns is managed
**What happened**:

When I read the code, I found that for newly managed ns, all pods under the ns were directly managed without necessary checks.

For example, pods containing the label "istio.io/dataplane-mode: none" should not be managed.

ref: https://github.com/kmesh-net/kmesh/blob/main/pkg/controller/manage/manage_controller.go#L260

**Environment**:
- Kmesh version:
- Others:

YaoZengzeng:
**What type of PR is this?**

/kind bug

Fixes #956

```release-note

```

### DIFF
--- a/pkg/controller/manage/manage_controller.go
+++ b/pkg/controller/manage/manage_controller.go
@@ -209,12 +209,12 @@ func (kmc *KmeshManageController) handleNamespaceUpdateFunc(oldObj, newObj inter
 	// Compare labels to check if they have actually changed
 	if !utils.ShouldEnroll(nil, oldNS) && utils.ShouldEnroll(nil, newNS) {
 		log.Infof("Enabling Kmesh for all pods in namespace: %s", newNS.Name)
-		kmc.enableKmeshForPodsInNamespace(newNS.Name)
+		kmc.enableKmeshForPodsInNamespace(newNS)
 	}
 
 	if utils.ShouldEnroll(nil, oldNS) && !utils.ShouldEnroll(nil, newNS) {
 		log.Infof("Disabling Kmesh for all pods in namespace: %s", newNS.Name)
-		kmc.disableKmeshForPodsInNamespace(newNS.Name)
+		kmc.disableKmeshForPodsInNamespace(newNS)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/manage/manage_controller.go
```diff
@@ -209,12 +209,12 @@ func (kmc *KmeshManageController) handleNamespaceUpdateFunc(oldObj, newObj inter
 	// Compare labels to check if they have actually changed
 	if !utils.ShouldEnroll(nil, oldNS) && utils.ShouldEnroll(nil, newNS) {
 		log.Infof("Enabling Kmesh for all pods in namespace: %s", newNS.Name)
-		kmc.enableKmeshForPodsInNamespace(newNS.Name)
+		kmc.enableKmeshForPodsInNamespace(newNS)
```
LiZhenCheng9527, line 212: When there is no manage label on the ns and there is a manage label on the pod, this pod will be managed by Kmesh. So when there is no manage label on the pod but there is a manag label on the ns, will the pod be managed in this case? 
Would this be a scenario where something goes wrong and should we report an error?

YaoZengzeng: If the ns is managed by Kmesh, the pods with no manage label or with label of "istio.io/dataplane-mode: Kmesh" should be managed. But if the label of pods is "istio.io/dataplane-mode: none", they should not be managed.

pkg/controller/manage/manage_controller.go
```diff
@@ -250,27 +250,29 @@ func (kmc *KmeshManageController) disableKmeshManage(pod *corev1.Pod) {
...
+	pods, err := kmc.podLister.Pods(namespace.Name).List(labels.Everything())
 	if err != nil {
-		log.Errorf("Error listing pods: %v", err)
+		log.Errorf("Error listing pods in namespace %s: %v", namespace.Name, err)
 		return
 	}
 
 	for _, pod := range pods {
-		kmc.enableKmeshManage(pod)
+		if utils.ShouldEnroll(pod, namespace) {
```
hzxuzhonghu, line 261: It is redundant, namespace already enabled

Issue #955, opened 2024-10-14
Test_restart with authz error log


hzxuzhonghu:
It looks very weird, because actually no authzPolicy set at all, this is because workload is deleted

```
time="2024-10-14T12:47:10Z" level=warning msg="Auth denied for connection: [10 244 1 75] because destination workload not found" subsys=auth
time="2024-10-14T12:47:10Z" level=info msg="Auth denied for connection: {srcIdentity:{trustDomain: namespace: serviceAccount:} dstNetwork: srcIp:[10 244 1 1] dstIp:[10 244 1 75] dstPort:8080}" subsys=auth
time="2024-10-14T12:47:10Z" level=warning msg="Auth denied for connection: [10 244 1 77] because destination workload not found" subsys=auth
time="2024-10-14T12:47:10Z" level=info msg="Auth denied for connection: {srcIdentity:{trustDomain: namespace: serviceAccount:} dstNetwork: srcIp:[10 244 1 1] dstIp:[10 244 1 77] dstPort:8080}" subsys=auth
time="2024-10-14T12:47:10Z" level=warning msg="Auth denied for connection: [10 244 1 76] because destination workload not found" subsys=auth
time="2024-10-14T12:47:10Z" level=info msg="Auth denied for connection: {srcIdentity:{trustDomain: namespace: serviceAccount:} dstNetwork: srcIp:[10 244 1 1] dstIp:[10 244 1 76] dstPort:8080}" subsys=auth
time="2024-10-14T12:47:10Z" level=warning msg="Auth denied for connection: [10 244 1 78] because destination workload not found" subsys=auth

```

LiZhenCheng9527:
How to reproduce the error？
Does delete httpbin when restarting Kmesh?

hzxuzhonghu:
I did run e2e test with https://github.com/kmesh-net/kmesh/pull/950, i believe the upgrade is not the cause, this issue should exist long since beginning. 

The root cause is that there is a race between workload delete and kmesh-daemon do auth check. What's weired to me is that there is no auth policy at all, but we do send that auth req to userspace via ringbuffer.

cc @supercharge-xsy @weli-l as well

PR #959, opened 2024-10-15
adjust log level for userspace authz

hzxuzhonghu:
**What type of PR is this?**

/kind enhancement

Fixes https://github.com/kmesh-net/kmesh/issues/955

```release-note

```

### DIFF
--- a/pkg/auth/rbac.go
+++ b/pkg/auth/rbac.go
@@ -98,19 +98,20 @@ func NewRbac(workloadCache cache.WorkloadCache) *Rbac {
 }
 
 func (r *Rbac) Run(ctx context.Context, mapOfTuple, mapOfAuth *ebpf.Map) {
-	if r == nil || mapOfTuple == nil {
-		log.Error("r or mapOfTuple is nil")
+	if r == nil {
+		return
+	}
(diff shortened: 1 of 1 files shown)


pkg/auth/rbac.go
```diff
@@ -179,7 +180,7 @@ func (r *Rbac) doRbac(conn *rbacConnection) bool {
 	dstWorkload := r.workloadCache.GetWorkloadByAddr(networkAddress)
 	// If no workload found, deny
 	if dstWorkload == nil {
```
hzxuzhonghu, line 182: @LiZhenCheng9527 I thought about this again, we cannot deny if src workload is not found. And also cannot simply allow if src not found,

LiZhenCheng9527: So we need to discuss different scenarios of source workload not found.
But how should we distinguish between these scenarios?

hzxuzhonghu: not needed from here, auth should not care whether src workload exist or not if the auth policy us based on ip or port

https://github.com/kmesh-net/kmesh/issues/961 is the way to ignore kubelet probe traffic

PR #973, opened 2024-10-21
fix cannot get RDS after grpc reconnect

### LINKED ISSUE #964: RDS Not Updating after kmesh long periods of Inactivity
**What happened**:

After running KMesh for an long period without any operations, RDS stops updating. The following observations were made:
1. KMesh is left running overnight without any operations.
2. The next day, modifications to the existing VirtualService do not take effect. For example, changing the match prefix in a VirtualService from` /test` to `/test-echo` does not update.
3. Restarting KMesh resolves the issue temporarily.
4. Debugging logs indicate that only CDS and EDS updates are happening, while RDS is not updating.

Modifications to VirtualService should be applied without needing to restart KMesh.

1. Start KMesh and configure a VirtualService with match prefix /test.
1. Allow KMesh to run overnight without any operations.
1. Modify the VirtualService to match prefix /test-echo.
1. Observe that the update does not take effect.

**Environment**:
- Kmesh version: 0.5
- K8S Node OS: HCE 2.0 with enhanced kernel
- Others:

lec-bit:
**What type of PR is this?**
/kind bug

fix cannot get RDS after grpc reconnect
after grpc reconnect,we did not clear the cache, and we forget to compare lastNonce in RDS,  so the judgment logic in the subsequent process believes that RDS does not need to be updated. 
Fixes #964 

```release-note

```

### DIFF
--- a/pkg/controller/ads/ads_controller.go
+++ b/pkg/controller/ads/ads_controller.go
@@ -64,7 +64,7 @@ func (c *Controller) AdsStreamCreateAndSend(client service_discovery_v3.Aggregat
 		stopCh:       make(chan struct{}),
 	}
 
-	c.Processor = newProcessor()
+	c.Processor.Reset()
 	if err := stream.Send(newAdsRequest(resource_v3.ClusterType, nil, "")); err != nil {
 		return fmt.Errorf("send request failed, %s", err)
 	}
--- a/pkg/controller/ads/ads_controller.go
+++ b/pkg/controller/ads/ads_controller.go
@@ -64,7 +64,7 @@ func (c *Controller) AdsStreamCreateAndSend(client service_discovery_v3.Aggregat
 		stopCh:       make(chan struct{}),
 	}
 
-	c.Processor.Reset()
+	c.Processor = newProcessor()
 	if err := stream.Send(newAdsRequest(resource_v3.ClusterType, nil, "")); err != nil {
 		return fmt.Errorf("send request failed, %s", err)
 	}
--- a/pkg/controller/ads/ads_processor.go
+++ b/pkg/controller/ads/ads_processor.go
@@ -320,6 +320,7 @@ func (p *processor) Reset() {
 		return
 	}
 	p.lastNonce = &lastNonce{}
+	p.Cache = NewAdsCache()
 }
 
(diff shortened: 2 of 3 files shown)


pkg/controller/ads/ads_controller.go
```diff
@@ -64,7 +64,7 @@ func (c *Controller) AdsStreamCreateAndSend(client service_discovery_v3.Aggregat
 		stopCh:       make(chan struct{}),
 	}
 
-	c.Processor.Reset()
+	c.Processor = newProcessor()
```
hzxuzhonghu, line 67: We donot want to start from zero. So donot init a totally new processor

pkg/controller/ads/ads_processor.go
```diff
@@ -184,12 +184,6 @@ func (p *processor) handleCdsResponse(resp *service_discovery_v3.DiscoveryRespon
 	// Note eds typed cluster, we do not flush to bpf map here, we need to wait for eds update.
 	p.Cache.ClusterCache.Flush()
 
-	if p.lastNonce.edsNonce == "" {
```
LiZhenCheng9527, line 187: why delete this?

pkg/controller/ads/ads_controller.go
lec-bit: I perform the operation of resetting the Cache in Reset now

pkg/controller/ads/ads_processor.go
```diff
@@ -326,6 +320,7 @@ func (p *processor) Reset() {
 		return
 	}
 	p.lastNonce = &lastNonce{}
+	p.Cache = NewAdsCache()
```
hzxuzhonghu, line 323: I donot think we should reinit a cache, because in ads handler, we depend on the cache to do optimization,

Ycan reset p.Cache.edsClusterNames and routeNames

pkg/controller/ads/ads_processor.go
lec-bit: This case can be solved after the reset edsClusterNames command is executed.

hzxuzhonghu:
Please fix the ut, then will approve

hzxuzhonghu (approved):
/lgtm

Issue #949, opened 2024-10-14
 kmeshctl version does not work as expected


hzxuzhonghu:
**What happened**:

It always return `v0.0.0-master`

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kmesh version:
- Others:

hzxuzhonghu:
Fixed by https://github.com/kmesh-net/kmesh/pull/1114

Issue #892, opened 2024-09-25
`make build` supports selectively compiling some of the binaries


YaoZengzeng:
**Task description**:

Now we have 4 binaries in Kmesh project:
1. kmesh-daemon
2. mdacore
3. kmesh-cni
4. kmeshctl

ref: https://github.com/kmesh-net/kmesh/blob/main/Makefile#L45

It would be better if we could selectively compile some of the binaries, which would be very beneficial to improving our development efficiency.

**Solution**:

We need to adjust the Makefile structure and then we can pass parameters to `make build`, for example, `BINARIES=kmeshctl make build`, so that only `kmeshctl` is compiled.

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

After a contributor has successfully completed 1-2 good first issue's,
they should be ready to move on to `help wanted` items, saving the remaining `good first issue` for other new contributors.

**How to join or take the task**:

Just reply on the issue with the message `/assign` in a separate line.

Then, the issue will be assigned to you.

**How to ask for help**:

If you need help or have questions, please feel free to ask on this issue.
The issue author or other members of the community will guide you through the contribution process.

YaoZengzeng:
Actullay you can't execute `make all` direclty, it need some prepration.

We usually use `make build` or `make docker` to compile, AFAIK, finally they will call `make` which is actually `make all`.

So you could try passing the `BINARIES` into `kmesh_compile.sh` and have it take effect somewhere. Perhaps some refactor is needed.

hzxuzhonghu:
cc @SpongeBob0318

hzxuzhonghu:
Great, not urgent, just confirm you are on this issue

abhi02max:
Hi team, I'd like to take this up. I have experience with build systems and can refactor the Makefile to accept target arguments for selective compilation. Could you assign this to me?

PR #978, opened 2024-10-23
kmeshctl install

### LINKED ISSUE #552: Make it easy to install a specific version of Kmesh

In the future, we may need a command line tool to install specific version of Kmesh with one click.

But since the project is still in its early stage, we may need some simpler methods quickly to implement this.

**Why is this needed**:

Until now, three versions of Kmesh has been released.

But I find there is no easy way to install specific version of Kmesh except directly modifying the image tag in the yaml.

thebigbone:
**What type of PR is this?**

/kind feature

adds install command for `kmeshctl` which will install all the resources for kmesh in a k8s cluster

Fixes #552 

```release-note
NONE
```


ctl/main.go
```diff
@@ -43,6 +44,7 @@ func main() {
 	rootCmd.AddCommand(waypoint.NewCmd())
 	rootCmd.AddCommand(version.NewCmd())
 	rootCmd.AddCommand(accesslog.NewCmd())
+	rootCmd.AddCommand(install.NewCmd())
```
YaoZengzeng, line 47: Also need to update [doc](https://github.com/kmesh-net/website/blob/main/content/en/docs/userguide/use_kmeshctl.md) when this PR get merged.

ctl/install/install.go
```diff
@@ -0,0 +1,167 @@
...
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "install kmesh with all the resources",
+		Run: func(cmd *cobra.Command, args []string) {
+			version, err := cmd.Flags().GetString("version")
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			kubeconfig := os.Getenv("HOME") + "/.kube/config"
```
YaoZengzeng, line 85: Use existing (tools)[https://github.com/kmesh-net/kmesh/blob/main/ctl/utils/utils.go#L31] to create k8s client direclty which has built-in method for applying yaml files.

ctl/install/install.go
```diff
@@ -0,0 +1,167 @@
+package install
```
YaoZengzeng, line 1: Add copyright

ctl/install/install.go
```diff
@@ -0,0 +1,167 @@
...
+
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+
+	"k8s.io/apimachinery/pkg/api/meta"
```
YaoZengzeng, line 20: Arrange the dependency order, put the standard lib together, put Kmesh lib together and put third-party dependencies together.

ctl/install/install.go
```diff
@@ -0,0 +1,167 @@
...
+		FieldManager: "sample-controller",
+	})
+
+	return err
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "install kmesh with all the resources",
```
YaoZengzeng, line 78: Add examples, refer to other subcommands.

ctl/install/install.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			kubeconfig := os.Getenv("HOME") + "/.kube/config"
+
+			config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+			if err != nil {
+				panic(err.Error())
+			}
+
+			resources := []Resource{
```
YaoZengzeng, line 92: Why not just apply all files in the `yaml` diretory instead of listing each files which may change in the future?

ctl/install/install.go
```diff
@@ -0,0 +1,167 @@
...
+				{Name: "clusterRole", URL: getURL(version, "clusterrole.yaml")},
+				{Name: "clusterRoleBinding", URL: getURL(version, "clusterrolebinding.yaml")},
+				{Name: "kmesh", URL: getURL(version, "kmesh.yaml")},
+				{Name: "l7EnvoyFilter", URL: getURL(version, "l7-envoyfilter.yaml")},
+				{Name: "serviceAccount", URL: getURL(version, "serviceaccount.yaml")},
+			}
+
+			fmt.Println("using version: ", version)
```
LiZhenCheng9527, line 100: nit：install kmesh version:

ctl/install/install.go
thebigbone: How would that work?

ctl/install/install.go
YaoZengzeng: I think you could.

ctl/install/install.go
YaoZengzeng: You could get all yaml contents and call [this function](https://github.com/istio/istio/blob/master/pkg/kube/client.go#L203)

YaoZengzeng: I don't know if there is an elegant way to direclty get the contents of all files under yaml directory or to know which files are in this directory and download them one by one.

thebigbone: > I don't know if there is an elegant way to direclty get the contents of all files under yaml directory or to know which files are in this directory and download them one by one.

 curl -s "https://api.github.com/repos/kmesh-net/kmesh/contents/deploy/yaml" | jq -r '.[] | .download_url'

thebigbone: The `ApplyYAMLContents` function works fine but it requires namespace argument. For now, I can separate the `kmesh-system` and `istio` but it can break in the future because there could be a new yaml file with a different namespace.

YaoZengzeng: I think you could specify namespace as "", so that it will direcly use the ns specified in yaml.

thebigbone: Oh, that's helpful. I was parsing the namespace field in every file.

thebigbone:
Applied all the requested changes.

ctl/install/install.go
```diff
@@ -0,0 +1,113 @@
...
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "install kmesh with all the resources",
+		Example: `# Install all kmesh resources (defaults to main):
+kmeshctl install
+
+# Install a specific version of kmesh
+kmeshctl install --version 0.5`,
+		Run: func(cmd *cobra.Command, args []string) {
+			version, err := cmd.Flags().GetString("version")
```
LiZhenCheng9527, line 48: We still have issue #949.
@YaoZengzeng  Has this problem been fixed?

YaoZengzeng: No, I hope this issue could be fixed with the solution of #892, but it seems that this problem is not being addressed now.  If no one has dealt it by then, I will fix it.

ctl/install/install.go
```diff
@@ -0,0 +1,113 @@
...
+		if filepath.Ext(file.Name) == ".yaml" {
+			resp, err := http.Get(file.DownloadURL)
+			if err != nil {
+				log.Fatal(err)
+			}
+			defer resp.Body.Close()
+
+			deployYaml, err := io.ReadAll(resp.Body)
+			if err != nil {
+				log.Fatalf("yamlFile.Get err   #%v ", err)
```
YaoZengzeng, line 105:
```suggestion
				log.Fatalf("yamlFile.Get err: %v ", err)
```

ctl/install/install.go
```diff
@@ -0,0 +1,113 @@
...
+
+type GitHubFile struct {
+	Name        string `json:"name"`
+	DownloadURL string `json:"download_url"`
+}
+
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "install kmesh with all the resources",
```
YaoZengzeng, line 41: Descriptive content use `Kmesh`, with the first letter capitalized.

YaoZengzeng:
The overall framework seems fine for me, @hzxuzhonghu PTAL.

ctl/install/install.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	cmd.Flags().String("version", "main", "Version of the resources to initialize")
+	return cmd
+}
+
+func getYAMLFile(version string) string {
+	var url string
+	if version != "main" {
+		url = fmt.Sprintf("https://api.github.com/repos/kmesh-net/kmesh/contents/deploy/yaml?ref=release-%s", version)
```
hzxuzhonghu, line 76: I donot think this is quite correct, because?:
1.  we didn't update the image tag 
2. should not use the github branch, but use the tag instead

So in short should we use helm chart https://github.com/kmesh-net/kmesh/releases/download/v0.5.0/kmesh-helm-v0.5.0.tgz 

or the OCI chart image https://github.com/orgs/kmesh-net/packages/container/package/kmesh-helm

ctl/install/install.go
```diff
@@ -0,0 +1,113 @@
...
+	} else {
+		url = "https://api.github.com/repos/kmesh-net/kmesh/contents/deploy/yaml?ref=main"
+	}
+
+	resp, err := http.Get(url)
+	if err != nil {
+		log.Fatal("error fetching files:", err)
```
hzxuzhonghu, line 83:
```suggestion
		log.Fatal("error fetching files: ", err)
```

ctl/install/install.go
thebigbone: It is using release tag? Using helm will introduce new dependencies which is totally unnecessary.

thebigbone:
Any update?

ctl/install/install.go
hzxuzhonghu: See the released version above

YaoZengzeng:
ping @hzxuzhonghu

hzxuzhonghu:
please see above comment, when you set version-=0.5, then you will install kmesh with latest image  https://github.com/kmesh-net/kmesh/blob/release-0.5/deploy/yaml/kmesh.yaml#L72

This is not expected, can you replace that image?

I would suggest we support v0.5.0 or 0.5.0 rather than 0.5, see our releases

thebigbone:
> please see above comment, when you set version-=0.5, then you will install kmesh with latest image https://github.com/kmesh-net/kmesh/blob/release-0.5/deploy/yaml/kmesh.yaml#L72
> 
> This is not expected, can you replace that image?
> 
> I would suggest we support v0.5.0 or 0.5.0 rather than 0.5, see our releases

Should I just replace the image and just get the yaml files from `main` then?

hzxuzhonghu:
>Should I just replace the image and get the yaml files from main than?

This probably not work, think of the case that a yaml is updated because of some new features. The old image may cannot work with the new daemonset yaml

thebigbone:
Removed the branch and instead using tags now.

thebigbone:
ping @hzxuzhonghu

ctl/install/install.go
```diff
@@ -0,0 +1,123 @@
...
+
+	var combinedYAML string
+
+	for _, file := range files {
+		if filepath.Ext(file.Name) == ".yaml" {
+			resp, err := http.Get(file.DownloadURL)
+			if err != nil {
+				log.Fatal(err)
+			}
+			defer resp.Body.Close()
```
hzxuzhonghu, line 111: nit: not use defer in a for loop

hzxuzhonghu (dismissed):
/lgtm

YOu can fix it later

thebigbone:
The kube package was changed so it's not able to find `ApplyYamlContents`

thebigbone:
ping @hzxuzhonghu

hzxuzhonghu:
Can you implement this method? Or use helm client https://github.com/kurator-dev/kurator/blob/ac68f35c8799773aea22643e39ca032a20e65fb7/pkg/client/client.go#L66

LiZhenCheng9527:
@thebigbone This PR can be merged after you pass CI.

thebigbone:
> Can you implement this method? Or use helm client https://github.com/kurator-dev/kurator/blob/ac68f35c8799773aea22643e39ca032a20e65fb7/pkg/client/client.go#L66

Should I use the istio's implementation directly?

thebigbone:
ping @hzxuzhonghu

thebigbone:
ping @LiZhenCheng9527

ctl/install/install.go
```diff
@@ -0,0 +1,123 @@
...
+
+			if version != "main" {
+				versionYaml := strings.Replace(combinedYAMLFile, ":latest", ":"+version, -1)
+
+				err = cli.ApplyYAMLContents("", versionYaml)
+				if err != nil {
+					log.Fatal(err)
+				}
+			} else {
+				err = cli.ApplyYAMLContents("", combinedYAMLFile)
```
LiZhenCheng9527, line 71: There is no ApplyYamlContents in kmesh's kube pkg.
After fixing this problem, this PR can be merged.

thebigbone: Yes, which is why I asked if I should use the istio's implementation or the previous doServerSideApply function which I used at the beginning.

thebigbone: I tried importing the `istio.io/istio/pkg/kube` and a wrapper around `ApplyYAMLContents` like this at `pkg/kube/client.go`:

```go
import istio "istio.io/istio/pkg/kube"

type CLIClient interface {
        /* existing code 
        Client

        // PodsForSelector finds pods matching selector.
        PodsForSelector(ctx context.Context, namespace string, labelSelectors ...string) (*v1.PodList, error)

        // NewPortForwarder creates a new PortForwarder configured for the given pod. If localPort=0, a port will be
        // dynamically selected. If localAddress is empty, "localhost" is used.
        NewPortForwarder(podName string, ns string, localAddress string, localPort int, podPort int) (PortForwarder, error)
        */

        ApplyYAMLContents(namespace string, yamls ...string) error
}

type client struct {
        /* existing code
        config *rest.Config

        kube          kubernetes.Interface
        gatewayapi    gatewayapiclient.Interface
        clientFactory *genericclioptions.ConfigFlags
        */
        
        // added istio client
        istioClient   istio.CLIClient
}

// wrapper around it
func (c *client) ApplyYAMLContents(namespace string, yamls ...string) error {
        return c.istioClient.ApplyYAMLContents(namespace, yamls...)
}

```

But I am getting panic error:

```sh
rogue@minikube:~/kmesh/ctl$ go run . install
install kmesh version:  main
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x276a440]

goroutine 1 [running]:
kmesh.net/kmesh/pkg/kube.(*client).ApplyYAMLContents(0x337978b?, {0x0?, 0xc000c79c80?}, {0xc000ca4610?, 0x2?, 0x0?})
        /home/rogue/kmesh/pkg/kube/client.go:81 +0x20
kmesh.net/kmesh/ctl/install.NewCmd.func1(0xc0007aa300?, {0x337975b?, 0x4?, 0x337975f?})
        /home/rogue/kmesh/ctl/install/install.go:71 +0x2e7
github.com/spf13/cobra.(*Command).execute(0xc000c62308, {0x5a8a4c0, 0x0, 0x0})
        /home/rogue/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:989 +0xa91
github.com/spf13/cobra.(*Command).ExecuteC(0xc000251208)
        /home/rogue/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(0xc0000061c0?)
        /home/rogue/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041 +0x13
main.main()
        /home/rogue/kmesh/ctl/main.go:27 +0x18
exit status 2
```

thebigbone:
I have implemented the method @LiZhenCheng9527

PR #982, opened 2024-10-25
add clusterid in grpc stream context

### LINKED ISSUE #981: Authentication failed in multi-cluster scenario
**What happened**:
My istiod control plane cluster differs from the data plane cluster, istiod calls the corresponding kube client for authentication when receive grpc stream. And the cluster id is extracted from grpc stream context to specify the data plane cluster, which it seems not passed in current version.

```go
// istio/security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go
kubeClient := a.getKubeClient(clusterID)
if kubeClient == nil {
return nil, fmt.Errorf("could not get cluster %s's kube client", clusterID)
}

id, err := tokenreview.ValidateK8sJwt(kubeClient, targetJWT, security.TokenAudiences)
if err != nil {
return nil, fmt.Errorf("failed to validate the JWT from cluster %q: %v", clusterID, err)
}
```
When cluster id is not set, control plane kube client will be used, and finally it leads to falsely calling of the control plane apiserver instead of data plane apiserver, and failes to authenticate kmesh controller.

```
2024-10-25T02:37:16.202931Z     error   security        Failed to authenticate client from 10.0.0.143:36586: Authenticator ClientCertAuthenticator: no verified chain is found; Authenticator KubeJWTAuthenticator: failed to validate the JWT from cluster "": the service account authentication returns an error: [invalid bearer token, square/go-jose: error in cryptographic primitive]
```

**Environment**:
- Kmesh version: *
- Others:

Kuromesi:
**What type of PR is this?**
/kind bug

Add clusterid in grpc stream context

Fixes #981 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/client.go
+++ b/pkg/controller/client.go
@@ -23,6 +23,7 @@ import (
 
 	discoveryv3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 
 	bpfwl "kmesh.net/kmesh/pkg/bpf/workload"
 	"kmesh.net/kmesh/pkg/constants"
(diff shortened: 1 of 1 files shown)


LiZhenCheng9527:
Thanks.
/lgtm
/approve

PR #1015, opened 2024-11-05
Use bpf map to pass node IP and pod sub gateway. Skip kubelet probe traffic management from ebpf

### LINKED ISSUE #961: Skip kubelet probe traffic management from ebpf

Add a way to tell ebpf the kubelete probe src ip address, so that in xdp and sockops, we donot need to process these connection/packet.

And then no need to even send metric info/authz req to kmesh-daemon.

I did some research the kubelet probe src ip commonly is node ip or pod sub net gateway address, diff because cni implement.

For calico env, the src ip is node ip 172.18.0.5

For kindnet, the src ip is 192.168.1.1

```
apiVersion: v1
kind: Node
metadata:
  annotations:
    
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/os: linux
    kubernetes.io/arch: amd64
    kubernetes.io/hostname: kind-worker
    kubernetes.io/os: linux
  name: kind-worker
  resourceVersion: "21264"
  uid: c0a23c49-52f2-4e8e-8480-79fa88208001
spec:
  podCIDR: 192.168.1.0/24
  podCIDRs:
  - 192.168.1.0/24
  providerID: kind://docker/kind/kind-worker
status:
  addresses:
  - address: 172.18.0.5
    type: InternalIP
  - address: kind-worker
    type: Hostname

````

### PR DESCRIPTION
**Pull request by LiZhenCheng9527**
**What type of PR is this?**
/kind enhancement

Fixes #961 

```release-note

```

### DIFF
--- a/bpf/include/common.h
+++ b/bpf/include/common.h
@@ -63,6 +63,12 @@ struct kmesh_context {
     bool via_waypoint;
 };
 
+struct kmesh_config {
+    __u32 bpf_log_level;
+    __u32 node_ip[4];
+    __u32 pod_gateway[4];
+};
+
 static inline void *kmesh_map_lookup_elem(void *map, const void *key)
 {
     return bpf_map_lookup_elem(map, key);
@@ -134,8 +140,8 @@ struct {
 struct {
     __uint(type, BPF_MAP_TYPE_ARRAY);
     __uint(max_entries, 1);
-    __uint(key_size, sizeof(__u32));
-    __uint(value_size, sizeof(__u32));
+    __type(key, int);
+    __type(value, struct kmesh_config);
 } kmesh_config_map SEC(".maps");
 
 #if KERNEL_VERSION_HIGHER_5_13_0
--- a/bpf/kmesh/workload/sockops.c
+++ b/bpf/kmesh/workload/sockops.c
@@ -56,12 +56,6 @@ static inline bool skip_specific_probe(struct bpf_sock_ops *skops)
         return false;
     }
 
-    BPF_LOG(ERR, SOCKOPS, "node ip is %u.%u", data->node_ip[0], data->node_ip[1]);
-    BPF_LOG(ERR, SOCKOPS, "node ip is %u.%u", data->node_ip[2], data->node_ip[3]);
-    BPF_LOG(ERR, SOCKOPS, "pod gateway is %u.%u", data->pod_gateway[0], data->pod_gateway[1]);
-    BPF_LOG(ERR, SOCKOPS, "pod gateway is %u.%u", data->pod_gateway[2], data->pod_gateway[3]);
-    BPF_LOG(ERR, SOCKOPS, "remote ip is %u", skops->remote_ip4);
-
     if (skops->family ==  AF_INET) {
         if (data->node_ip[3] == skops->remote_ip4) {
             return true;
@@ -71,7 +65,20 @@ static inline bool skip_specific_probe(struct bpf_sock_ops *skops)
         }
     }
 
-    if (skops->family == AF_INET6) {}
+    if (skops->family == AF_INET6) {
+        if (data->node_ip[0] == skops->remote_ip6[0] && 
+        data->node_ip[1] == skops->remote_ip6[1] && 
+        data->node_ip[2] == skops->remote_ip6[2] && 
+        data->node_ip[3] == skops->remote_ip6[3]) {
+            return true;
+        }
+        if (data->pod_gateway[0] == skops-> remote_ip6[0] &&
+        data->pod_gateway[1] == skops-> remote_ip6[1] &&
+        data->pod_gateway[2] == skops-> remote_ip6[2] &&
+        data->pod_gateway[3] == skops-> remote_ip6[3]) {
+            return true;
+        }
+    }
 
     return false;
 }
--- a/bpf/kmesh/workload/sockops.c
+++ b/bpf/kmesh/workload/sockops.c
@@ -46,6 +46,36 @@ static inline bool is_managed_by_kmesh(struct bpf_sock_ops *skops)
     return (*value == 0);
 }
 
+static inline bool skip_specific_probe(struct bpf_sock_ops *skops)
+{   
+    struct kmesh_config *data = {0};
+    int key_of_kmesh_config = 0;
+    data = kmesh_map_lookup_elem(&kmesh_config_map, &key_of_kmesh_config);
+    if (!data) {
+        BPF_LOG(ERR, SOCKOPS, "get kmesh congfig failed");
+        return false;
+    }
+
+    BPF_LOG(ERR, SOCKOPS, "node ip is %u.%u", data->node_ip[0], data->node_ip[1]);
+    BPF_LOG(ERR, SOCKOPS, "node ip is %u.%u", data->node_ip[2], data->node_ip[3]);
+    BPF_LOG(ERR, SOCKOPS, "pod gateway is %u.%u", data->pod_gateway[0], data->pod_gateway[1]);
+    BPF_LOG(ERR, SOCKOPS, "pod gateway is %u.%u", data->pod_gateway[2], data->pod_gateway[3]);
+    BPF_LOG(ERR, SOCKOPS, "remote ip is %u", skops->remote_ip4);
+
+    if (skops->family ==  AF_INET) {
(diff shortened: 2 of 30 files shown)


bpf/kmesh/workload/sockops.c
```diff
@@ -46,6 +46,20 @@ static inline bool is_managed_by_kmesh(struct bpf_sock_ops *skops)
     return (*value == 0);
 }
 
+static inline bool skip_kubelet_probe(struct bpf_sock_ops *skops)
```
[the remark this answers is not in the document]
LiZhenCheng9527: Done

bpf/include/common.h
```diff
@@ -133,11 +133,31 @@ struct {
 
 struct {
     __uint(type, BPF_MAP_TYPE_ARRAY);
-    __uint(max_entries, 1);
+    __uint(max_entries, 3);
```
hzxuzhonghu, line 136: I think we should not use each element to store one option. This may cause the map change frequently.

How about  increase the value size, we use value memory to store a options struct

pkg/bpf/bpf.go
```diff
@@ -264,6 +273,87 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
 	return versionMap
 }
 
+func (l *BpfLoader) skipSpecificProbe() {
```
hzxuzhonghu, line 276: we can define an struct to hold all the bpf options, then call this function `UpdateBpfProgOptions`

pkg/bpf/bpf.go
```diff
@@ -264,6 +273,87 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+	var nodeIPStr string
+	nodeAddresses := node.Status.Addresses
+	for _, address := range nodeAddresses {
+		if address.Type == corev1.NodeInternalIP {
+			nodeIPStr = address.Address
+		}
+	}
+
+	nodeIP := net.ParseIP(nodeIPStr)
+	return binaryToUint32(nodeIP[12:16])
```
hzxuzhonghu, line 323: why only make use the last 4 bytes

hzxuzhonghu: we have `ConvertIpToUint32` and `ConvertPortToBigEndian` select which suitable

pkg/bpf/bpf.go
```diff
@@ -264,6 +273,87 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+}
+
+func getNodePodSubGateway(nodeName string, clientSet kubernetes.Interface) *net.IPNet {
+	node, err := clientSet.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
hzxuzhonghu, line 327: donot duplicate calling this get

pkg/bpf/bpf.go
LiZhenCheng9527: > why only make use the last 4 bytes

Since the type of the map value is `unit32`, only the ipv4 case was considered first.

bpf/include/common.h
LiZhenCheng9527: But different options have differenet update frequencies. 
For example the `log_level` will change in use, but the `node IP` and `pod gateway` only need to be passed when the Kmesh-daemon is started.
So I think putting the configurations in one structure will transfer invalid information when updating, resulting in a waste of resources.

pkg/bpf/bpf.go
```diff
@@ -264,6 +272,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+func (l *BpfLoader) UpdateBpfProgOptions() {
+	nodeName := os.Getenv("NODE_NAME")
+	if nodeName == "" {
+		log.Errorf("skip kubelet probe failed: %s", "node name empty")
+		return
+	}
+
+	clientSet, err := utils.GetK8sclient()
+	if err != nil {
+		log.Errorf("get kubernetest client for getting node IP error: %v", err)
```
YaoZengzeng, line 284: return?

pkg/bpf/bpf.go
```diff
@@ -264,6 +272,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+		log.Errorf("failed to resolve ip from podCIDR: %v", err)
+		return nil
+	}
+
+	return podIPNet
+}
+
+func calculateGateway(ipnet *net.IPNet) uint32 {
+	if ipnet == nil {
+		log.Errorf("failed to calculate pod sub gateway: %s", "*ipNet is empty")
```
YaoZengzeng, line 339: Why not combine into one string?

pkg/bpf/bpf.go
```diff
@@ -264,6 +272,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+		log.Errorf("failed to get node: %v", err)
+		return
+	}
+
+	// pass node ip and pod gateway to skip processing of kubelet access traffic.
+	nodeIP := getNodeIPAddress(nodeName, node)
+	gateway := calculateGateway(getNodePodSubGateway(nodeName, node))
+
+	keyOfNodeIP := uint32(1)
+	keyOfPodSubGateway := uint32(2)
```
YaoZengzeng, line 298: Define as constants?

LiZhenCheng9527:
@hzxuzhonghu

pkg/bpf/bpf.go
```diff
@@ -264,6 +279,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
 }
 
+func (l *BpfLoader) setBpfProgOptions() {
+	nodeName := os.Getenv("NODE_NAME")
+	if nodeName == "" {
+		log.Error("skip kubelet probe failed: node name empty")
+		return
+	}
+
+	clientSet, err := utils.GetK8sclient()
```
hzxuzhonghu, line 289: this function is moved

pkg/bpf/bpf.go
```diff
@@ -264,6 +279,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+		NodeIP:      nodeIP,
+		PodGateway:  gateway,
+	}
+
+	if l.kmeshConfig == nil {
+		log.Errorf("skip kubelet probe failed: %v", "kmeshConfigMap is nil")
+		return
+	}
+	if err := l.kmeshConfig.Update(&keyOfKmeshBpfConfig, &ValueOfKmeshBpfConfig, ebpf.UpdateAny); err != nil {
+		log.Errorf("update kmeshConfigMap failed: %v", err)
```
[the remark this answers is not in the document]
LiZhenCheng9527: Done

pkg/bpf/bpf.go
```diff
@@ -264,6 +279,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+func getNodePodSubGateway(node *corev1.Node) [16]byte {
+	podCIDR := node.Spec.PodCIDR
+	_, subNet, err := net.ParseCIDR(podCIDR)
+	if err != nil {
+		log.Errorf("failed to resolve ip from podCIDR: %v", err)
+		return [16]byte{0}
+	}
+	podGateway := [16]byte{0}
+	nets.CopyIpByteFromSlice(&podGateway, subNet.IP.To16())
+	if err != nil {
```
hzxuzhonghu, line 350: remove,i think this err is handled before

pkg/bpf/bpf.go
```diff
@@ -264,6 +279,82 @@ func recoverVersionMap(pinPath string) *ebpf.Map {
...
+		BpfLogLevel: constants.BPF_LOG_INFO,
+		NodeIP:      nodeIP,
+		PodGateway:  gateway,
+	}
+
+	if l.kmeshConfig == nil {
+		log.Errorf("skip kubelet probe failed: %v", "kmeshConfigMap is nil")
+		return
+	}
+	if err := l.kmeshConfig.Update(&keyOfKmeshBpfConfig, &ValueOfKmeshBpfConfig, ebpf.UpdateAny); err != nil {
```
hzxuzhonghu, line 317: Suggest: if if l.kmeshConfig != nil {

}, rather than printing the error log in L314

hzxuzhonghu (approved):
Otherwise lgtm

hzxuzhonghu (approved):
/lgtm

PR #1047, opened 2024-11-18
Correct Workload handle order

### LINKED ISSUE #1042: Kmesh workload update bpf map order maybe not correct
**Please provide an in-depth description of the question you have**:

**What do you think about this question?**:

In the case a pod is created for a service, istio will push the workload to kmesh daemon.

- What does kmesh do?

In `handleWorkload`, it will first check workload related services. And then in  `handleWorkloadNewBoundServices`, it update

`endpoint map -> service map`

Then at last, it update the `backendmap -> frontend map`.

- Let's take a look how bpf prog works?

In cgroup/connect, it first check the frontend map -> service map-> endpoint map-> backend map.

As the bpf map update is not atomic, so if backend map has not been updated, but the bpf prog has selected this backendUID, we can get no backend. Then kmesh lb policy does not work now, it will fallback to k8s clusterIP direct access.

**Environment**:
- Kmesh version:
- Others:

hzxuzhonghu:
**What type of PR is this?**

No matter workload/service update, we should make sure bpf map is updated in the following orders:

backendmap-> endpoint map -> service map -> frontend map

In case workload/service delete, the order should be as below:

frontend map-> service map -> endpoint map -> backend map

Fixes #1042

And made many cleanups

```release-note

```

### DIFF
--- a/pkg/controller/workload/cache/endpoint_cache.go
+++ b/pkg/controller/workload/cache/endpoint_cache.go
@@ -30,7 +30,11 @@ type Endpoint struct {
 type EndpointCache interface {
 	List(uint32) map[uint32]Endpoint // Endpoint slice by ServiceId
 	AddEndpointToService(ep Endpoint, serviceID uint32)
+	// DeleteEndpoint delete a endpoint regardless of the priority
 	DeleteEndpoint(workloadID, serviceID uint32)
+	// DeleteEndpointWithPriority delete a endpoint with given priority
+	DeleteEndpointWithPriority(serviceID, workloadID, prio uint32)
+	// DeleteEndpointByServiceId delete all endpoints belong to a given service
 	DeleteEndpointByServiceId(uint32)
 }
 
@@ -60,6 +64,16 @@ func (s *endpointCache) DeleteEndpoint(serviceID, workloadID uint32) {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 	delete(s.endpointsByServiceId[serviceID], workloadID)
+}
+
+func (s *endpointCache) DeleteEndpointWithPriority(serviceID, workloadID, prio uint32) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	if s.endpointsByServiceId[serviceID] != nil {
+		if ep, ok := s.endpointsByServiceId[serviceID][workloadID]; ok && ep.Prio == prio {
+			delete(s.endpointsByServiceId[serviceID], workloadID)
+		}
+	}
 }
 
 func (s *endpointCache) DeleteEndpointByServiceId(serviceId uint32) {
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -580,10 +580,10 @@ func (p *Processor) updateEndpointOneByOne(serviceId uint32, epsUpdate []cache.E
 			log.Errorf("addWorkloadToService workload %d service %d failed: %v", ev.BackendUid, sKey.ServiceId, err)
 			return err
 		}
-		eksDeletes := []bpfcache.EndpointKey{ek}
+		epKeys := []bpfcache.EndpointKey{ek}
 		// delete ek
-		if err := p.deleteEndpointRecords(eksDeletes); err != nil {
-			log.Errorf("delete [%#v] from endpoint map failed: %s", eksDeletes, err)
+		if err := p.deleteEndpointRecords(epKeys); err != nil {
+			log.Errorf("delete [%#v] from endpoint map failed: %s", epKeys, err)
 		}
 	}
 	return nil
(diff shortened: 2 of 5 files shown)


pkg/controller/workload/workload_processor.go
```diff
@@ -413,14 +411,24 @@ func (p *Processor) updateWorkload(workload *workloadapi.Workload) error {
...
-				return err
-			}
+func (p *Processor) updateWorkloadInFrontendMap(workload *workloadapi.Workload) error {
+	// we should not store frontend data of hostname network mode pods
+	// please see https://github.com/kmesh-net/kmesh/issues/631
+	if workload.GetNetworkMode() == workloadapi.NetworkMode_HOST_NETWORK {
+		return nil
+	}
+
+	uid := p.hashName.Hash(workload.GetUid())
```
LiZhenCheng9527, line 425: nit: uid->backendUid

pkg/controller/workload/workload_processor.go
```diff
@@ -466,8 +481,8 @@ func (p *Processor) handleWorkload(workload *workloadapi.Workload) error {
 		return err
 	}
 
-	// update frontend and backend bpf map
-	if err := p.updateWorkload(workload); err != nil {
+	// 4. update workload in frontend map
+	if err := p.updateWorkloadInFrontendMap(workload); err != nil {
 		log.Errorf("updateWorkload %s failed: %v", workload.Uid, err)
```
YaoZengzeng, line 486:
```suggestion
		log.Errorf("updateWorkloadInFrontendMap %s failed: %v", workload.Uid, err)
```

pkg/controller/workload/workload_processor.go
```diff
@@ -632,49 +648,56 @@ func (p *Processor) storeServiceData(serviceName string, waypoint *workloadapi.G
...
+				if err = p.updateEndpoint(sk.ServiceId, false); err != nil { // this will change bpf map totally
+					log.Errorf("UpdateEndpoint failed, err:%s", err)
+				}
+				updateValue := bpf.ServiceValue{}
+				if err = p.bpf.ServiceLookup(&sk, &updateValue); err != nil {
+					log.Warnf("Lookup Service failed, err:%s", err)
+				}
+				updateValue.LbPolicy = newValue.LbPolicy
+				if err = p.bpf.ServiceUpdate(&sk, &updateValue); err != nil {
+					log.Errorf("Update Service failed, err:%s", err)
```
YaoZengzeng, line 671: Just log, don't return error?

pkg/controller/workload/cache/endpoint_cache.go
```diff
@@ -30,7 +30,11 @@ type Endpoint struct {
 type EndpointCache interface {
 	List(uint32) map[uint32]Endpoint // Endpoint slice by ServiceId
 	AddEndpointToService(ep Endpoint, serviceID uint32)
+	// DeleteEndpoint delete a endpoint regardless of the priority
 	DeleteEndpoint(workloadID, serviceID uint32)
```
nlgwcy, line 34: `DeleteEndpoint` maybe not used.

pkg/controller/workload/workload_processor.go
```diff
@@ -466,9 +481,9 @@ func (p *Processor) handleWorkload(workload *workloadapi.Workload) error {
 		return err
 	}
 
-	// update frontend and backend bpf map
-	if err := p.updateWorkload(workload); err != nil {
-		log.Errorf("updateWorkload %s failed: %v", workload.Uid, err)
+	// 4. update workload in frontend map
+	if err := p.updateWorkloadInFrontendMap(workload); err != nil {
+		log.Errorf("updateWorkloadInFrontendMap %s failed: %v", workload.Uid, err)
```
nlgwcy, line 486: It is recommended that failed resource recycling be considered. Otherwise, residual BPF maps exist.

pkg/controller/workload/cache/endpoint_cache.go
hzxuzhonghu: Good catch, this is only used in test now, will change that

pkg/controller/workload/workload_processor.go
hzxuzhonghu: what do you mean

pkg/controller/workload/workload_processor.go
```diff
@@ -918,30 +931,38 @@ func (p *Processor) deleteEndpointRecords(endpointKeys []bpf.EndpointKey) error
...
-
 			if err = p.bpf.EndpointLookup(&ek, &ev); err != nil {
 				log.Errorf("Lookup endpoint %#v failed: %s", ek, err)
 			}
-			p.EndpointCache.DeleteEndpoint(ek.ServiceId, ev.BackendUid)
 
+			// on delete, first update the service map, then update the endpoint map
 			sv.EndpointCount[ek.Prio] = sv.EndpointCount[ek.Prio] - 1
-			if err = p.bpf.ServiceUpdate(&sk, &sv); err != nil {
+			if err := p.bpf.ServiceUpdate(&sk, &sv); err != nil {
```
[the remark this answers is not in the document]
hzxuzhonghu: You meant we should first delete endpoint map and then update service map.

I would think should be service map first then endpoint map. At leat this way, we do not select non-exist endpoint.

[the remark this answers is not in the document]
hzxuzhonghu: done

PR #1049, opened 2024-11-18
Optimize service lb policy update

hzxuzhonghu:
**What type of PR is this?**

/kind enhancement

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -19,7 +19,6 @@ package workload
 import (
 	"fmt"
 	"os"
-	"slices"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


pkg/controller/workload/workload_processor.go
```diff
@@ -567,32 +574,22 @@ func (p *Processor) updateEndpointOneByOne(serviceId uint32, epsUpdate []cache.E
...
 		return p.updateEndpointOneByOne(serviceId, endpointSlice, toLLb)
 	} else {
-		filtered := []cache.Endpoint{}
-		for _, endpoint := range endpoints {
-			if endpoint.Prio > 0 {
-				filtered = append(filtered, endpoint)
-			}
-		}
+		filtered := slices.Filter(endpointSlice, func(e cache.Endpoint) bool {
+			return e.Prio > 0
```
LiZhenCheng9527, line 591: Can this return value be used as a function error return?

hzxuzhonghu: what do you mean

LiZhenCheng9527: My mistake. This one is right.

Issue #1033, opened 2024-11-12
Improve the compatibility of Kmesh observability with the Istio ecosystem


delavet:
**Task description**:
1. Currently, the names of kmesh's observability metrics begin with "kmesh," such as "kmesh_tcp_workload_sent_bytes_total". This could break some software in Istio Ecosystem such as kiali, which shows mesh topology graph based on Istio standard metrics like "istio_tcp_sent_bytes_total"

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

After a contributor has successfully completed 1-2 good first issue's,
they should be ready to move on to `help wanted` items, saving the remaining `good first issue` for other new contributors.

**How to join or take the task**:

Just reply on the issue with the message `/assign` in a separate line.

Then, the issue will be assigned to you.

**How to ask for help**:

If you need help or have questions, please feel free to ask on this issue.
The issue author or other members of the community will guide you through the contribution process.

Kuromesi:
Also, I have noticed that all metrics are stored in a single kmesh controller pod, in my experience, when there are about thousands destination services , memory usage can increase up to about 500 mb for a SINGLE envoy sidecar, so I think maybe we should find out a solution for the cost.

hzxuzhonghu:
cc @LiZhenCheng9527

LiZhenCheng9527:
> /assign

So are you responsible for adapting istio's standard telemetry API?

delavet:
> > /assign
> 
> So are you responsible for adapting istio's standard telemetry API?

I will currently focus on producing documentation that utilizes the Prometheus rename mechanism to make Kiali work on Kmesh. As for adapting the Telemetry API, based on previous discussions, the semantics of the Telemetry API may not be appropriate for the sidecar-less data plane models like Kmesh and ambient. Maybe we can first look at how the Istio community supports this for ambient. Currently, Ztunnel does not support it.

LiZhenCheng9527:
> 
> I will currently focus on producing documentation that utilizes the Prometheus rename mechanism to make Kiali work on Kmesh. As for adapting the Telemetry API, based on previous discussions, the semantics of the Telemetry API may not be appropriate for the sidecar-less data plane models like Kmesh and ambient. Maybe we can first look at how the Istio community supports this for ambient. Currently, Ztunnel does not support it.

Well Done. 
I will focus on improving the kmesh metrics deletion mechanism and scheduled erasing outdataed metrics data.

PR #1055, opened 2024-11-19
Improvement of metriclabels removal capability at workload granularity

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

**Which issue(s) this PR fixes**
a part of #1033 

```release-note

```

### DIFF
--- a/pkg/controller/telemetry/metric.go
+++ b/pkg/controller/telemetry/metric.go
@@ -575,16 +575,20 @@ func (m *MetricController) updatePrometheusMetric() {
 	}
 
 	// delete metrics
-	workloadDeleteLength := len(deleteWorkload)
-	serviceDeleteLength := len(deleteService)
-	for i := 0; i < workloadDeleteLength; i++ {
-		deleteWorkloadMetricInPrometheus(deleteWorkload[i])
+	deleteLock.Lock()
+	// Creating a copy reduces the amount of time spent adding locks in the programme
+	workloadReplica := deleteWorkload
+	deleteWorkload = nil
+	serviceReplica := deleteService
+	deleteService = nil
+	deleteLock.Unlock()
+
+	for i := 0; i < len(workloadReplica); i++ {
+		deleteWorkloadMetricInPrometheus(workloadReplica[i])
 	}
-	for i := 0; i < serviceDeleteLength; i++ {
-		deleteServiceMetricInPrometheus(deleteService[i])
+	for i := 0; i < len(serviceReplica); i++ {
+		deleteServiceMetricInPrometheus(serviceReplica[i])
 	}
-	deleteWorkload = deleteWorkload[workloadDeleteLength:]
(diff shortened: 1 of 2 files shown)


pkg/controller/telemetry/metric.go
```diff
@@ -577,14 +577,22 @@ func (m *MetricController) updatePrometheusMetric() {
 	// delete metrics
 	workloadDeleteLength := len(deleteWorkload)
 	serviceDeleteLength := len(deleteService)
+	// Creating a copy reduces the amount of time spent adding locks in the programme
+	workloadLock.Lock()
```
hzxuzhonghu, line 581: I donot think this is enough, `DeleteWorkloadMetric` function deleteWorkload is not locked.

In addition to that, can you share one lock for both servce and workload

pkg/controller/telemetry/metric.go
```diff
@@ -577,14 +577,22 @@ func (m *MetricController) updatePrometheusMetric() {
 	// delete metrics
 	workloadDeleteLength := len(deleteWorkload)
```
hzxuzhonghu, line 578: why not protect it

pkg/controller/telemetry/metric.go
```diff
@@ -577,14 +577,22 @@ func (m *MetricController) updatePrometheusMetric() {
...
 	serviceDeleteLength := len(deleteService)
+	// Creating a copy reduces the amount of time spent adding locks in the programme
+	workloadLock.Lock()
+	workloadCopy := deleteWorkload[:workloadDeleteLength]
```
YaoZengzeng, line 582: It's better to change the name, because in fact no copy occures, just a slice split.

pkg/controller/telemetry/metric.go
```diff
@@ -575,16 +575,22 @@ func (m *MetricController) updatePrometheusMetric() {
 	}
 
 	// delete metrics
+	deleteLock.Lock()
 	workloadDeleteLength := len(deleteWorkload)
 	serviceDeleteLength := len(deleteService)
+	// Creating a copy reduces the amount of time spent adding locks in the programme
+	workloadReplica := deleteWorkload[:workloadDeleteLength]
```
hzxuzhonghu, line 582:
```suggestion
	workloadReplica := deleteWorkload
```

pkg/controller/telemetry/metric.go
```diff
@@ -575,16 +575,22 @@ func (m *MetricController) updatePrometheusMetric() {
...
+	deleteWorkload = nil
+	serviceReplica := deleteService[:serviceDeleteLength]
```
hzxuzhonghu, line 584:
```suggestion
	serviceReplica := deleteService
```

hzxuzhonghu: And L579 L580 are not needed

Issue #1053, opened 2024-11-19
Encounter crash using the latest version of Kmesh


delavet:
**What happened**:
The Kmesh daemon is crashing on started when I user the latest Kmesh Image.
<img width="1219" alt="image" src="https://github.com/user-attachments/assets/fa1ddeec-6d92-41be-8db6-7f523940fe97">
The logs are as belows:

```
none on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --bpf-fs-path=\"/sys/fs/bpf\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --cgroup2-path=\"/mnt/kmesh_cgroup2\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --cni-etc-path=\"/etc/cni/net.d\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --conflist-name=\"\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --enable-accesslog=\"false\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --enable-bpf-log=\"true\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --enable-bypass=\"false\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --enable-mda=\"false\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --enable-secret-manager=\"false\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --help=\"false\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --mode=\"dual-engine\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --plugin-cni-chained=\"true\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="FLAG: --profiling=\"false\"" subsys=manager
time="2024-11-19T05:43:27Z" level=info msg="kmesh version map loadfailed: no such file or directory, start normally" subsys=bpf
time="2024-11-19T05:43:27Z" level=info msg="kmesh start with Normal" subsys=bpf
[...]

PR #1056, opened 2024-11-19
Add kmesh ut test on 5.15

hzxuzhonghu:
**What type of PR is this?**

This to make sure bpf prog can be loaded successfully in kernel 5.15,  since github action has a limited node image. So uubuntu 2004 is the smallest version that we can use.

We have encountered many cases that previous CI can run successfully, but in our local dev machine(kernel 5.15) it failed.

See https://github.com/kmesh-net/kmesh/issues/1053

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/main-5.15.yml
+++ b/.github/workflows/main-5.15.yml
@@ -1,0 +1,38 @@
+name: BPF Compatability Test on Kernel 5.15
+
+on:
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - '**.svg'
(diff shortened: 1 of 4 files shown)


.github/workflows/main-5.15.yml
```diff
@@ -0,0 +1,59 @@
+name: Kmesh BPF Test 5.15
```
LiZhenCheng9527, line 1:
```suggestion
name: Kmesh BPF Test 5.15 kernel version
```

Issue #1064, opened 2024-11-23
Kernel-native mode route function is unavailable.


lec-bit:
**What happened**:
Kernel-native mode route function is unavailable.

After optimizing the features supported by large-scale scenarios, some bugs have emerged:
The route function is unavailable on the enhanced kernel.
        The domains record cannot be queried properly, resulting in the route function being unavailable. Currently, the issue can be temporarily circumvented by modifying field_len=MAP_VAL_STR_SIZE. We look forward to a formal solution in the future.
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kmesh version:
- Others:

PR #1066, opened 2024-11-23
fix repeat string field proc

nlgwcy:
**What type of PR is this?**

/kind bug

Fixes https://github.com/kmesh-net/kmesh/issues/1064

```release-note

```

### DIFF
--- a/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
+++ b/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
@@ -270,8 +270,11 @@ static unsigned int bitmap_find_first_clear(unsigned char *bitmap, unsigned int 
 static unsigned int alloc_outer_key(struct op_context *ctx, int size)
 {
     unsigned int i, j;
-    if (size <= 0)
+    if (size <= 0) {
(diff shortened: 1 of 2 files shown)


bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -385,6 +386,25 @@ static int get_string_or_msg_field_size(struct op_context *ctx, const ProtobufCF
...
+}
+
+static int get_field_size(struct op_context *ctx, const ProtobufCFieldDescriptor *field)
+{
+    char *value = NULL;
+
+    if (field->type == PROTOBUF_C_TYPE_MESSAGE)
+        return get_msg_field_size(field);
+
+    if (field->type == PROTOBUF_C_TYPE_STRING) {
```
LiZhenCheng9527, line 401: The case `field->type == PROTOBUF_C_TYPE_STRING` is already excluded in line 519, so do we still need to do this case on this side?

hzxuzhonghu:
@bitcoffeeiux

LiZhenCheng9527:
@nlgwcy This PR needs to be updated

hzxuzhonghu (approved):

PR #1070, opened 2024-11-26
map_mertic and bpf_prog_metric test

skwwwwww:
**What type of PR is this?**
/kind enhancement

### DIFF
--- a/pkg/controller/telemetry/bpf_prog_metric_test.go
+++ b/pkg/controller/telemetry/bpf_prog_metric_test.go
@@ -1,0 +1,167 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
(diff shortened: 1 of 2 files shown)


pkg/controller/telemetry/bpf_prog_metric_test.go
```diff
@@ -0,0 +1,171 @@
...
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			bpfMetric := NewBpfProgMetric()
+			bpfMetric.updateOperationMetricCache(tt.data, tt.labels)
+
+			bpfMetric.mutex.RLock()
```
LiZhenCheng9527, line 89: Why do you need a lock here?

PR #1075, opened 2024-11-27
Remove enable-bpf-log from the startup parameters

### LINKED ISSUE #1071: Remove startup parameters: enable-bpf-log

**Why is this needed**:
enable-bpf-log does not prevent log uploads in the kernel. So it should be removed or updated to enable it to organize the kernel code to upload logs

LiZhenCheng9527:
**What type of PR is this?**
/kind cleanup

Fixes #1071

```release-note

```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -97,10 +97,8 @@ func (c *Controller) Start(stopCh <-chan struct{}) error {
 		return nil
 	}
 
-	if c.bpfConfig.EnableBpfLog {
-		if err := logger.StartRingBufReader(ctx, c.mode, c.bpfConfig.BpfFsPath); err != nil {
-			return fmt.Errorf("failed to start ringbuf reader: %v", err)
-		}
+	if err := logger.StartRingBufReader(ctx, c.mode, c.bpfConfig.BpfFsPath); err != nil {
+		return fmt.Errorf("fail to start ringbuf reader: %v", err)
 	}
 	// kmeshConfigMap.Monitoring initialized to uint32(1).
 	// If the startup parameter is false, update the kmeshConfigMap.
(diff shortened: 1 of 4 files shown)


pkg/controller/controller.go
```diff
@@ -97,10 +97,8 @@ func (c *Controller) Start(stopCh <-chan struct{}) error {
 		return nil
 	}
 
-	if c.bpfConfig.EnableBpfLog {
-		if err := logger.StartRingBufReader(ctx, c.mode, c.bpfConfig.BpfFsPath); err != nil {
-			return fmt.Errorf("failed to start ringbuf reader: %v", err)
-		}
+	if err := logger.StartRingBufReader(ctx, c.mode, c.bpfConfig.BpfFsPath); err != nil {
+		return fmt.Errorf("fail to start ringbuf reader: %v", err)
```
YaoZengzeng, line 101:
```suggestion
		return fmt.Errorf("failed to start ringbuf reader: %v", err)
```

So  we now use `kmeshctl log` to replace the function of `--enable-bpf-log` flag?

LiZhenCheng9527: No. 
This parameter was originally only used to create the ringbuf for `bpf_log`, and would actually cause an error if ringbuf was created. Not relevant for kmeshctl log.

LiZhenCheng9527: No. The two are not related.
Remove `enable-bpf-log` because if you don't create a ringbuf, it will cause a ringbuf blocking error.

hzxuzhonghu (approved):
/lgtm

We can also not start the reader by checking the kmesh_events map. This will reuqire us to make `kmesh_events` map protected with the macro `KERNEL_VERSION_HIGHER_5_13_0`

PR #1078, opened 2024-11-28
kmesh monitoring support for headless services

### LINKED ISSUE #816: Headless service support

We need to think about how to support headless service in kmesh. Need a proposal, of course.

As headless service has no cluster ip, but has associated endpoints. Dns resolve of a headless service can randomly return an endpoint ip. 

And there is another special headless service without selectors external name service: https://kubernetes.io/docs/concepts/services-networking/service/#without-selectors.  Istio has changed the behavior about externalname service. 

**Why is this needed**:

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Fixes #816 

```release-note

```

### DIFF
--- a/samples/httpbin/headless-httpbin.yaml
+++ b/samples/httpbin/headless-httpbin.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: httpbin-headless
+  labels:
+    app: httpbin
+spec:
+  ports:
+  - name: http
+    port: 8000
+    targetPort: 80
+  clusterIP: None
+  selector:
+    app: httpbin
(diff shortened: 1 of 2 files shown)


samples/httpbin/headless-httpbin.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
YaoZengzeng, line 1: It would be better to add a E2E test for Headless service.

LiZhenCheng9527: I will add it

hzxuzhonghu (approved):
/lgtm

PR #1102, opened 2024-12-09
Optimised handling when no podCIDR is available

LiZhenCheng9527:
**What type of PR is this?**
/kind bug

Not all clustered environments have `podCIDR`.
Therefore optimised for handling in this case
Fixes #

```release-note

```

### DIFF
--- a/pkg/bpf/bpf.go
+++ b/pkg/bpf/bpf.go
@@ -340,6 +340,10 @@ func getNodeIPAddress(node *corev1.Node) [16]byte {
 
 func getNodePodSubGateway(node *corev1.Node) [16]byte {
 	podCIDR := node.Spec.PodCIDR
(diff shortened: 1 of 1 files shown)


pkg/bpf/bpf.go
```diff
@@ -340,6 +340,10 @@ func getNodeIPAddress(node *corev1.Node) [16]byte {
 
 func getNodePodSubGateway(node *corev1.Node) [16]byte {
 	podCIDR := node.Spec.PodCIDR
+	if podCIDR == "" {
+		return [16]byte{0}
```
hzxuzhonghu, line 344: We should think about how to get the podCIDR in such env. Should file an issue

hzxuzhonghu (approved):
/lgtm

PR #1103, opened 2024-12-09
Expand the ringbuf size of tcp_info

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Extend the ringbuf size of tcp_info to avoid ringbuf reserve failure during high concurrency.
tcp_info reports 96bytes of data at a time. Combined with real usage, I think having 2000 simultaneous accesses to a link on a node is going to cover most usage scenarios. 
So choose to set the size of ringbuf to 256KB.
Fixes #

```release-note

```


hzxuzhonghu (approved):
Can you add some pr description why this value is used

LiZhenCheng9527:
@hzxuzhonghu PR's description has been updated

PR #1121, opened 2024-12-18
misc bug fix in manage controller

hzxuzhonghu:
**What type of PR is this?**

/kind cleanup

This is motivated by @bitcoffeeiux refereencing this controller. I found there are something not perfect here.

1.  We donot need shared informer factory here. 
2.  The controller Run() method cannot be stopped
3. We can reuse the handlePodUpdate function to process add event

Others are mostly clean code

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/manage/manage_controller.go
+++ b/pkg/controller/manage/manage_controller.go
@@ -310,7 +310,7 @@ func (c *KmeshManageController) syncPod(key QueueItem) error {
 	}
 	namespace, err := c.namespaceLister.Get(pod.Namespace)
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			return nil
 		}
 		return fmt.Errorf("failed to get pod namespace %s: %v", pod.Namespace, err)
--- a/pkg/controller/manage/manage_controller.go
+++ b/pkg/controller/manage/manage_controller.go
@@ -306,16 +306,14 @@ func (c *KmeshManageController) syncPod(key QueueItem) error {
 		if apierrors.IsNotFound(err) {
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


pkg/controller/manage/manage_controller.go
```diff
@@ -333,10 +295,39 @@ func (c *KmeshManageController) processItems() bool {
...
 	return true
 }
 
+func (c *KmeshManageController) syncPod(key QueueItem) error {
+	pod, err := c.podLister.Pods(key.podNs).Get(key.podName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		log.Errorf("failed to get pod %s/%s: %v", key.podNs, key.podName, err)
```
LiZhenCheng9527, line 309: This error message is presented on line L290. My suggestion is to delete the log here and put the message in the err return value.
```suggestion
		return fmt.Errorf("failed to get pod %s/%s: %v", key.podNs, key.podName, err)
```

pkg/controller/manage/manage_controller.go
```diff
@@ -333,10 +295,37 @@ func (c *KmeshManageController) processItems() bool {
...
+	pod, err := c.podLister.Pods(key.podNs).Get(key.podName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get pod %s/%s: %v", key.podNs, key.podName, err)
+	}
+	namespace, err := c.namespaceLister.Get(pod.Namespace)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
```
YaoZengzeng, line 313:
```suggestion
		if apierrors.IsNotFound(err) {
```
?

PR #1123, opened 2024-12-18
rename bpf map names to be more explicit

hzxuzhonghu:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/bpf/kmesh/workload/include/config.h
+++ b/bpf/kmesh/workload/include/config.h
@@ -19,11 +19,11 @@
 #define map_of_service       km_service
 #define map_of_endpoint      km_endpoint
 #define map_of_backend       km_backend
-#define map_of_auth          km_auth
-#define map_of_tuple         km_tuple
-#define map_of_tcp_info      km_tcpinfo
-#define map_of_authz         km_authz
-#define map_of_dst_info      km_dstinfo
+#define map_of_auth_result   km_auth_res
+#define map_of_auth_req      km_auth_req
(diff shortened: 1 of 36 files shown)


bpf/kmesh/workload/include/config.h
```diff
@@ -15,21 +15,21 @@
...
-#define map_of_service        km_service
-#define map_of_endpoint       km_endpoint
-#define map_of_backend        km_backend
-#define map_of_auth           km_auth
-#define map_of_tuple          km_tuple
-#define map_of_tcp_info       km_tcpinfo
-#define map_of_authz          km_authz
-#define map_of_dst_info       km_dstinfo
-#define map_of_tail_call_prog km_tailcallprog
-#define xdp_tailcall_map      km_xdptailcall
```
hzxuzhonghu, line 28: please take a look at this file only, other modifications are all to adapt to this

LiZhenCheng9527: I think this is a good modification. 
But it would also be nice to add comments to some of the maps with more detail. Such as `km_cgr_tailcall`.
After all, names are limited in length.

hzxuzhonghu: I do think comments at the place where map defined is better, because `xdp_tailcall_map` is used every where

LiZhenCheng9527:
@lec-bit Evaluate if there will be an impact on the Kernel-Native mode

PR #1148, opened 2024-12-27
Optimise the update logic of the frontend table to cope with duplicate service and workload addresses.

### LINKED ISSUE #1147: Optimising the frontend table for variable service addresses.
Optimise the update logic of the frontend table to cope with variable service addresses.

**Why is this needed**:

Previously, the design only considered cluster service. But after the introduction of `serviceEntry`, the existing processing logic will have loopholes, resulting in data residuals. Therefore, the processing logic needs to be optimised to meet the processing requirements of the out-of-cluster service.

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

The logic of the existing frontend map in handling service update does not consider the scenario of service address change. 
https://github.com/kmesh-net/kmesh/blob/7425245e40171db1759198cc9ced8d20cc0a5d66/pkg/controller/workload/workload_processor.go#L691-L695
When updating, the new service will be used directly in frontend map update, and will not process the service information that already exists in the map, which will result in the old service information being left. 

So, I add a function  can compare two slices and return the data added to a over b and the data missing from b over a.
And Optimising the frontend table for variable service addresses, to process the old data in the frontend map.

Fixes #1147 

```release-note

```


pkg/controller/workload/workload_processor.go
```diff
@@ -682,10 +702,24 @@ func (p *Processor) updateServiceMap(service *workloadapi.Service) error {
...
+		// Compare the addresses of the old and new maps to avoid residual.
+		// If the data can be found in the km_service map, it is also stored in the serviceCache.
+		newServiceAddress := utils.GetAddressesFromService(service)
+		oldServiceAddress := utils.GetAddressesFromService(oldService)
+		_, removeServiceAddress := utils.CompareIpByte(newServiceAddress, oldServiceAddress)
+
+		frontKey := bpf.FrontendKey{}
+		for _, address := range removeServiceAddress {
+			nets.CopyIpByteFromSlice(&frontKey.Ip, address)
+			if err := p.bpf.FrontendDelete(&frontKey); err != nil {
```
hzxuzhonghu, line 715: we can abstract a new function as i see it is used by workload handler too

pkg/utils/slices_handler.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	for _, item := range aSet {
+		aNew = append(aNew, item)
+	}
+
+	return aNew, bMissing
+}
+
+func GetAddressesFromService(service *workloadapi.Service) [][]byte {
```
hzxuzhonghu, line 70: This is not related to slice processing, we can move it to controller/workload pkg

LiZhenCheng9527: I will move it to pkg/nets

pkg/controller/workload/workload_processor.go
LiZhenCheng9527: Done

pkg/controller/workload/workload_processor.go
```diff
@@ -486,6 +487,22 @@ func (p *Processor) handleWorkload(workload *workloadapi.Workload) error {
 		log.Errorf("updateWorkloadInFrontendMap %s failed: %v", workload.Uid, err)
 		return err
 	}
+	if oldWorkload != nil {
+		// To be able to find a workload in the workloadCache,
+		// you need to determine whether the address of the workload has changed or not.
+		// And clean up the residue
+		newWorkloadAddresses := workload.GetAddresses()
+		oldWorkloadAddresses := oldWorkload.GetAddresses()
+		_, removeAddresses := utils.CompareIpByte(newWorkloadAddresses, oldWorkloadAddresses)
```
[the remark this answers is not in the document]
LiZhenCheng9527: Done

pkg/utils/slices_handler.go
hzxuzhonghu: isn't it  related to service object, or you can provide a method to it, like `ResourceName`

pkg/controller/workload/workload_processor.go
```diff
@@ -682,10 +702,19 @@ func (p *Processor) updateServiceMap(service *workloadapi.Service) error {
 				return nil
 			}
 		}
+
+		// Compare the addresses of the old and new maps to avoid residual.
+		// If the data can be found in the km_service map, it is also stored in the serviceCache.
+		newServiceAddress := getAddressesFromService(service)
+		oldServiceAddress := getAddressesFromService(oldService)
+		removeServiceAddress := nets.CompareIpByte(newServiceAddress, oldServiceAddress)
+		if err := p.deleteOutdataFrontend(removeServiceAddress); err != nil {
```
hzxuzhonghu, line 711: donot call outdated, by this you make the function limited to process outdated address only. But actually it provide general functionality

pkg/controller/workload/workload_processor.go
```diff
@@ -1022,3 +1052,30 @@ func (p *Processor) storeWorkloadPolicies(uid string, polices []string) {
 		log.Errorf("storeWorkloadPolicies failed, workload %s, err: %s", uid, err)
 	}
 }
+
+func (p *Processor) deleteOutdataFrontend(addresses [][]byte) error {
+	frontKey := bpf.FrontendKey{}
+	for _, address := range addresses {
+		nets.CopyIpByteFromSlice(&frontKey.Ip, address)
+		if err := p.bpf.FrontendDelete(&frontKey); err != nil {
```
hzxuzhonghu, line 1060: In case it delete a service frontend map i think we should do some protection. There are some other work we can do, maybe you should do that in another pr of you

pkg/controller/workload/workload_processor.go
LiZhenCheng9527: Done

pkg/nets/nets_test.go
```diff
@@ -67,3 +67,40 @@ func TestCopyIpByteFromSlice(t *testing.T) {
 		})
 	}
 }
+
+func TestCompareIpByte(t *testing.T) {
```
hzxuzhonghu, line 71: Recommend table driven test

hzxuzhonghu (approved):
/lgtm

PR #1161, opened 2025-01-04
fix ebpf map elem nums grow abnormal

lec-bit:
**What type of PR is this?**
/kind bug

When applying and deleting the same yaml configuration repeatedly, services or pods will be continuously created and deleted. At this time, the number of records in the ebpf-map will not be completely deleted, causing the memory usage of ebpf_map to continue to grow.
Fixes #

```release-note

```

### DIFF
--- a/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
+++ b/bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
@@ -1136,8 +1136,7 @@ static int repeat_field_del(struct op_context *ctx, const ProtobufCFieldDescript
             goto end;
 
         for (i = 0; i < *(size_t *)n; i++) {
-            outer_key = (unsigned int *)map_object + i;
-            indirect_field_del(ctx, *outer_key, field);
+            indirect_field_del(ctx, *((uintptr_t *)map_object + i), field);
         }
     default:
         break;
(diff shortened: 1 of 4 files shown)


bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
```diff
@@ -1473,4 +1473,4 @@ int deserial_init()
         return ret;
     }
     return 0;
-}
```
nlgwcy, line 1476: Why has this diff?

pkg/cache/v2/listener_test.go
```diff
@@ -407,3 +408,137 @@ func TestListenerFlushAndLookup(t *testing.T) {
 	listener.ApiStatus = core_v2.ApiStatus_UPDATE
 	assert.Equal(t, listener.String(), listener_val.String())
 }
+
+func checkMap64Count(adsObj *ads.BpfAds, t *testing.T) int {
```
nlgwcy, line 412:
```suggestion
func checkMapCount(map *ebpf.Map, expectCount int, t *testing.T) int {
```

nlgwcy: move `checkMapCount` to bpf_map.go

pkg/utils/test/bpf_map.go
```diff
@@ -81,3 +83,23 @@ func EqualIp(src [16]byte, dst []byte) bool {
 	}
 	return true
 }
+
+func GetMapCount(kmeshMap *ebpf.Map, t *testing.T) int {
+	var count int
```
nlgwcy, line 88:
```suggestion
	var count int = 0
```

Issue #1162, opened 2025-01-07
Should not store non local workloads' authorization 


hzxuzhonghu:
**What would you like to be added**:

Filter out non local authorizations processing

**Why is this needed**:

Kmesh only need the authorization policy that apply to local workloads. 

In addition, `map_of_authz_policy` size is MAP_SIZE_OF_AUTH_POLICY(512), which is not big enough to hold all the authz in large cluster

PR #1166, opened 2025-01-08
Filter unused authz

hzxuzhonghu:
**What type of PR is this?**

Fixes  https://github.com/kmesh-net/kmesh/issues/1162

Currently istiod does not support pushing only local authorizations. 

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -180,17 +180,16 @@ func (p *Processor) removeWorkload(uid string) error {
 	}
 	p.WorkloadCache.DeleteWorkload(uid)
 	telemetry.DeleteWorkloadMetric(wl)
-	return p.removeWorkloadFromBpfMap(uid)
+	return p.removeWorkloadFromBpfMap(wl)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/workload/workload_processor.go
```diff
@@ -212,16 +211,11 @@ func (p *Processor) removeWorkloadFromBpfMap(uid string) error {
...
 	// 4. delete auth policy of workload
-	wpkValue := bpf.WorkloadPolicyValue{}
-	wpkDelete.WorklodId = backendUid
-	if err = p.bpf.WorkloadPolicyLookup(&wpkDelete, &wpkValue); err == nil {
-		if err = p.bpf.WorkloadPolicyDelete(&wpkDelete); err != nil {
-			log.Errorf("WorkloadPolicyDelete failed: %s", err)
-			return err
-		}
+	if workload.Node == p.nodeName {
+		p.deleteWorkloadPolicies(backendUid)
```
LiZhenCheng9527, line 215: nit: Don't you have to check if there are any policies here?
Just like before

hzxuzhonghu: not needed

PR #1188, opened 2025-01-15
fix Update mode failed

lec-bit:
**What type of PR is this?**
/kind bug

kmesh currently does not support Update mode. When starting from Update mode, it should start in Normal mode. However, all residual data is not properly cleared, resulting in a startup error.
Fixes #

```release-note

```

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -73,6 +73,9 @@ const (
 	VersionPath         = "/bpf_kmesh/map/"
 	WorkloadVersionPath = "/bpf_kmesh_workload/map/"
 
+	KmKernelNativeBpfPath = "/bpf_kmesh"
+	KmDualEngineBpfPath   = "/bpf_kmesh_workload"
+
 	TailCallMap = "tail_call_map"
 	Prog_link   = "prog_link"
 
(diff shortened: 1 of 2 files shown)


LiZhenCheng9527:
When is the plan to support update?

hzxuzhonghu:
can you test it manually since we donot have e2e for this now

lec-bit:
image

pkg/constants/constants.go
```diff
@@ -73,6 +73,9 @@ const (
 	VersionPath         = "/bpf_kmesh/map/"
 	WorkloadVersionPath = "/bpf_kmesh_workload/map/"
 
+	kmKernelNativeBpfPath = "/bpf_kmesh"
```
LiZhenCheng9527, line 76:
```suggestion
	KmKernelNativeBpfPath = "/bpf_kmesh"
```

pkg/constants/constants.go
```diff
@@ -73,6 +73,9 @@ const (
...
+	kmDualEngineBpfPath   = "/bpf_kmesh_workload"
```
LiZhenCheng9527, line 77:
```suggestion
	KmDualEngineBpfPath   = "/bpf_kmesh_workload"
```

LiZhenCheng9527 (commented):
If your codes has some problem, how do you build successful?

PR #1225, opened 2025-02-06
Added changes to generated api file under same package

### LINKED ISSUE #1024: Make protobuf generated api file under same package with .proto

Currently we make the `https://github.com/kmesh-net/kmesh/tree/main/api/workloadapi` generated files into https://github.com/kmesh-net/kmesh/tree/main/api/v2/workloadapi. It does work well from fdunctionality, in my local dev IDE, goland/vscode, the code navigation does not work well, when i want to naviagate to https://github.com/kmesh-net/kmesh/blob/52d8f9ef7de5641e2718d1e25b641ba416e2c65c/api/v2/workloadapi/workload.pb.go#L707, but it jump to https://github.com/kmesh-net/kmesh/blob/52d8f9ef7de5641e2718d1e25b641ba416e2c65c/api/workloadapi/workload.proto#L138

7h3-3mp7y-m4n:
**What type of PR is this?**
### Added changes to `Makefile` to make protobuf generated api file under same package with .proto
/kind bug

This add nessacry change to `Makefile` to generate the file inside the same directory using `protoc --proto_path=. --go_out=. --go_opt=paths=source_relative workloadapi/workload.proto`
Fixes #1024 

**checking to CI because of arm64**

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -21,7 +21,7 @@ include ../mk/bpf.print.mk
 include ../mk/bpf.vars.mk
 
 C_OUTPUT_DIR := v2-c
-GO_OUTPUT_DIR := v2
+GO_OUTPUT_DIR := kmesh/api/workloadapi
 GO_BUILD_DIR := kmesh.net/kmesh/api
 
 PROTO_PATH := ../


hzxuzhonghu:
I can run make gen-proto successfully in my AMD64 env. 

But it is expected workload.pb.go to be in `api/workloadapi/workload.pb.go` instead of  `api/workloadapi/workloadapi/workload.pb.go`

hzxuzhonghu:
BTW, you can make other protobuf behave same

PR #1247, opened 2025-02-21
Refactor dns

LiZhenCheng9527:
**What type of PR is this?**
/kind enhancement

Refactored the DNS resolution module to decouple it from the CDS in Kernel-Native mode. This facilitates the reuse of the DNS resolution module in Dual-Engine mode.

Fixes #

```release-note

```


pkg/dns/ads_handler.go
hzxuzhonghu, line 1: I would suggest we do provide a dns pkg fundamental capability and make ads/workload  bussiness related part in their own pkg

pkg/dns/ads_handler.go
```diff
@@ -0,0 +1,218 @@
...
+		log.Errorf("failed to dns resolve: %v", err)
+		return false
+	}
+	adsResolver.adsDnsResolve(e, addresses)
+	adsResolver.adsCache.ClusterCache.Flush()
+	return true
+}
+
+func (adsResolver *AdsDnsResolver) refreshAdsWorker() {
+	for adsResolver.refreshAdsDns() {
```
hzxuzhonghu, line 101: I do believe such arch should be shared with workload

pkg/dns/dns.go
```diff
@@ -49,101 +40,28 @@ const (
...
-				}
-			}
-		}
-		if pos >= 0 {
-			e.LbEndpoints = slices.Replace(e.LbEndpoints, pos, pos+1, lbEndpoints...)
-		}
-	}
-
-	return ready
+type PendingResolveDomain struct {
```
hzxuzhonghu, line 56: How do you plean to support dual-engine then? 
If you are going to add another struct here, wouldn't that still coupling

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,239 @@
...
+	endpointv3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+	"k8s.io/client-go/util/workqueue"
+
+	core_v2 "kmesh.net/kmesh/api/v2/core"
+	"kmesh.net/kmesh/pkg/dns"
+)
+
+// adsDnsResolver is DNS resolver of Kernel Native
+type AdsDnsResolver struct {
```
hzxuzhonghu, line 35: Nit: 
```suggestion
type DnsResolver struct {
```

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,239 @@
...
+	Clusters        chan []*clusterv3.Cluster
+	adsCache        *AdsCache
+	dnsResolver     *dns.DNSResolver
+	dnsRefreshQueue workqueue.TypedDelayingInterface[any]
+}
+
+func NewAdsDnsResolver(adsCache *AdsCache) (*AdsDnsResolver, error) {
```
hzxuzhonghu, line 42: It is under ads pkg, so we can remove the redundant

```suggestion
func NewDnsResolver(adsCache *AdsCache) (*DnsResolver, error) {
```

pkg/dns/dns.go
```diff
@@ -49,106 +40,33 @@ const (
...
 type DNSResolver struct {
-	DnsResolverChan   chan []*clusterv3.Cluster
 	client            *dns.Client
-	resolvConfServers []string
-	cache             map[string]*domainCacheEntry
-	// adsCache is used for update bpf map
-	adsCache *ads.AdsCache
-	// dns refresh priority queue based on exp
-	dnsRefreshQueue workqueue.TypedDelayingInterface[any]
+	AdsDnsChan        chan string
```
hzxuzhonghu, line 44: rename to reliminate ads words here

pkg/dns/dns.go
```diff
@@ -49,106 +40,33 @@ const (
...
+	ResolvConfServers []string
+	Cache             map[string]*DomainCacheEntry
```
hzxuzhonghu, line 46: Donot just make it public, there may introduce race conditions

pkg/dns/dns.go
```diff
@@ -157,157 +75,100 @@ func NewDNSResolver(adsCache *ads.AdsCache) (*DNSResolver, error) {
...
-	domains := getPendingResolveDomain(clusters)
-
-	// Stow domain updates, need to remove unwatched domains first
-	r.removeUnwatchedDomain(domains)
-	for _, v := range domains {
-		r.Lock()
-		if r.cache[v.domainName] == nil {
-			r.cache[v.domainName] = &domainCacheEntry{}
+// TODO: Dns Cache clean up
+func (r *DNSResolver) GetAddressesFromDnsCache(domains map[string]struct{}) map[string]*DomainCacheEntry {
```
hzxuzhonghu, line 91: I donot think this is a good pattern, the caller cannot know exactly which domain is not returned unless compare one by one

pkg/dns/dns.go
```diff
@@ -157,157 +75,100 @@ func NewDNSResolver(adsCache *ads.AdsCache) (*DNSResolver, error) {
...
+	_, exist := r.Cache[e.Domain]
+	r.Unlock()
+	// if the domain is no longer watched, no need to refresh it
+	if !exist {
+		return true
 	}
+	_, ttl, err := r.resolve(e.Domain)
+	if err != nil {
+		log.Errorf("failed to dns resolve: %v", err)
+		return false
```
hzxuzhonghu, line 122: This is not right, returning false means quiting the resolve go routine

pkg/dns/dns.go
```diff
@@ -157,157 +75,100 @@ func NewDNSResolver(adsCache *ads.AdsCache) (*DNSResolver, error) {
...
+	}
+	if ttl > e.RefreshRate {
+		ttl = e.RefreshRate
+	}
+	if ttl == 0 {
+		ttl = DeRefreshInterval
+	}
+	r.RefreshQueue.AddAfter(e, ttl)
+	r.AdsDnsChan <- e.Domain
```
hzxuzhonghu, line 131: what does this mean?

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,241 @@
...
+		dnsResolver: resolver,
+	}, nil
+}
+
+func (adsResolver *AdsDnsResolver) StartAdsDnsResolver(stopCh <-chan struct{}) {
+	go adsResolver.startAdsResolver()
+	go adsResolver.dnsResolver.StartDnsResolver()
+	go func() {
+		<-stopCh
+		adsResolver.dnsResolver.RefreshQueue.ShutDown()
```
hzxuzhonghu, line 68: the propagation chain is too long, you can pass the stopCh to the `NewDNSResolver`

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,241 @@
...
+		Clusters: make(chan []*clusterv3.Cluster),
+		// dnsRefreshQueue: workqueue.NewTypedDelayingQueueWithConfig(workqueue.TypedDelayingQueueConfig[any]{Name: "dnsRefreshQueue"}),
+		adsCache:    adsCache,
+		dnsResolver: resolver,
+	}, nil
+}
+
+func (adsResolver *AdsDnsResolver) StartAdsDnsResolver(stopCh <-chan struct{}) {
+	go adsResolver.startAdsResolver()
+	go adsResolver.dnsResolver.StartDnsResolver()
```
hzxuzhonghu, line 65: I feel this is messy cand confusing, it is unclear what is AdsResolver and what is dns resolver

YaoZengzeng:
@LiZhenCheng9527 add a design doc first would be better for review.

pkg/dns/dns.go
LiZhenCheng9527: Notify `adsDnsResolver` that the `domain` has completed dns resolution

pkg/dns/dns.go
LiZhenCheng9527: Only getting addresses then removes the domain from the map to be resolved.

pkg/controller/ads/dns.go
LiZhenCheng9527: Done

pkg/dns/dns.go
LiZhenCheng9527: Done

pkg/dns/dns.go
LiZhenCheng9527: Done

pkg/dns/dns.go
LiZhenCheng9527: Done

pkg/dns/dns.go
```diff
@@ -164,150 +82,84 @@ func NewDNSResolver(adsCache *ads.AdsCache) (*DNSResolver, error) {
...
+		return
+	}
+	if ttl > e.RefreshRate {
+		ttl = e.RefreshRate
 	}
+	if ttl == 0 {
+		ttl = DeRefreshInterval
+	}
+	r.refreshQueue.AddAfter(e, ttl)
+	r.DnsChan <- e.Domain
```
hzxuzhonghu, line 124: why send the domain to DnsChan

pkg/dns/dns.go
```diff
@@ -164,150 +82,84 @@ func NewDNSResolver(adsCache *ads.AdsCache) (*DNSResolver, error) {
...
+	return addrs, ttl, nil
 }
 
-func (r *DNSResolver) GetAllCachedDomains() []string {
-	r.RLock()
-	defer r.RUnlock()
-	out := make([]string, 0, len(r.cache))
-	for domain := range r.cache {
-		out = append(out, domain)
+// resolveDomains takes a slice of cluster
```
hzxuzhonghu, line 150: outdate comment

pkg/dns/dns.go
```diff
@@ -164,150 +82,84 @@ func NewDNSResolver(adsCache *ads.AdsCache) (*DNSResolver, error) {
...
+func (r *DNSResolver) ResolveDomains(domainName string) {
```
hzxuzhonghu, line 151: The function name is not clear too

pkg/dns/dns.go
```diff
@@ -404,40 +256,51 @@ func getMinTTL(m *dns.Msg, refreshRate time.Duration) time.Duration {
...
+	r.RLock()
+	defer r.RUnlock()
+	out := make([]string, 0, len(r.cache))
+	for domain := range r.cache {
+		out = append(out, domain)
+	}
+	return out
+}
+
+func (r *DNSResolver) GetOneDomainFromCache(domain string) ([]string, bool) {
```
hzxuzhonghu, line 269: GetDomainAddress

pkg/dns/dns.go
```diff
@@ -404,40 +256,51 @@ func getMinTTL(m *dns.Msg, refreshRate time.Duration) time.Duration {
...
+	r.Lock()
```
hzxuzhonghu, line 270: Rlock

pkg/dns/dns.go
```diff
@@ -404,40 +256,51 @@ func getMinTTL(m *dns.Msg, refreshRate time.Duration) time.Duration {
...
+	addresses, ok := r.cache[domain]
+	r.Unlock()
+	return addresses.Addresses, ok
+}
+
+func (r *DNSResolver) GetAddressesFromCache(domains map[string]struct{}) map[string]*DomainCacheEntry {
+	r.Lock()
```
hzxuzhonghu, line 277: Rlock

pkg/dns/dns.go
```diff
@@ -404,40 +256,51 @@ func getMinTTL(m *dns.Msg, refreshRate time.Duration) time.Duration {
...
+	for domain := range r.cache {
+		if _, ok := domains[domain]; ok {
+			continue
 		}
+		delete(r.cache, domain)
 	}
+}
 
-	return domains
+func (r *DNSResolver) AddDomainIntoRefreshQueue(info *DomainInfo, time time.Duration) {
```
hzxuzhonghu, line 301: It seems this is called by upper level controller, so prefer renaming to, e.g `EnqueDnsDomain`

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,242 @@
...
+
+	for k, v := range domains {
+		r.dnsResolver.ResolveDomains(k)
+		domainInfo := &dns.DomainInfo{
+			Domain:      v.DomainName,
+			RefreshRate: v.RefreshRate,
+		}
+		r.dnsResolver.AddDomainIntoRefreshQueue(domainInfo, 0)
+	}
+	go r.refreshAdsWorker(domains)
```
hzxuzhonghu, line 112: when will this thread exit?

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,242 @@
...
+	go func() {
+		<-stopCh
+		close(r.Clusters)
+	}()
+}
+
+func (r *dnsController) startDnsController() {
+	rateLimiter := make(chan struct{}, dns.MaxConcurrency)
+	for clusters := range r.Clusters {
+		rateLimiter <- struct{}{}
```
hzxuzhonghu, line 75: I am not sure how does the ratelimiter effect, because in the dns pkg, it always run in order

pkg/dns/dns.go
LiZhenCheng9527: Notify the dns Controller that a domain name has resolved and you can update the clusters containing that domain name

pkg/dns/dns.go
LiZhenCheng9527: Already remove it

pkg/dns/dns.go
LiZhenCheng9527: Already renamed

pkg/dns/dns.go
LiZhenCheng9527: Done

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,334 @@
...
+	return &dnsController{
+		Clusters:           make(chan []*clusterv3.Cluster),
+		cache:              adsCache,
+		dnsResolver:        resolver,
+		clusterCache:       make(map[string]*pendingResolveDomain),
+		pendingClusterinfo: make(map[string]clusterInfo),
+	}, nil
+}
+
+func (r *dnsController) StartKernelNativeDnsController(stopCh <-chan struct{}) {
```
hzxuzhonghu, line 80:
```suggestion
func (r *dnsController) Run(stopCh <-chan struct{}) {
```

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,334 @@
...
+		}(clusters)
+	}
+}
+
+func (r *dnsController) resolveDomains(cds []*clusterv3.Cluster) {
+	domains := getPendingResolveDomain(cds)
+	hostNames := make(map[string]struct{})
+
+	for k := range domains {
+		hostNames[k] = struct{}{}
```
hzxuzhonghu, line 104: If you need convert after calling getPendingResolveDomain, why not directly return what you want

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	// store all pending hostnames of clusters in r.hostInfo
+	for _, cluster := range cds {
+		clusterName := cluster.GetName()
+		hostInfo := getHostInfo(cluster)
+		info := clusterInfo{
+			info: hostInfo,
```
hzxuzhonghu, line 112: can you explain why store  ip address?

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,334 @@
...
+	go func() {
+		<-stopCh
+		close(r.Clusters)
+	}()
+}
+
+func (r *dnsController) startDnsController() {
+	for clusters := range r.Clusters {
+		go func(clusters []*clusterv3.Cluster) {
+			r.resolveDomains(clusters)
```
hzxuzhonghu, line 94: I think this is not safe running async in separate routines,

think about this case:

t0:    a resolveDomains thread

t1:   another resolveDomains

If t1 is run before t0, even some step.The state machine would be broken

hzxuzhonghu (commented):
Please ADD SOME COMMENTS，it is very hard to review

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (r *dnsController) resolveDomains(cds []*clusterv3.Cluster) {
+	domains, hostNames := getPendingResolveDomain(cds)
+
+	// store all pending hostnames of clusters in r.hostInfo
+	for _, cluster := range cds {
+		clusterName := cluster.GetName()
+		info := getHostName(cluster)
+		r.pendingClusterInfo[clusterName] = info
```
hzxuzhonghu, line 98: From this line, i can understand pendingClusterInfo is not a good name.  please rename

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	// delete any scheduled re-resolve for domains we no longer care about
+	r.dnsResolver.RemoveUnwatchDomain(hostNames)
+
+	// Update clusters based on the data in the dns cache.
+	for k, v := range domains {
```
hzxuzhonghu, line 105: Seems you haven't send the domains to the real resolver yet

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+		addresses := r.dnsResolver.GetDNSAddresses(k)
+		// Already have record in dns cache
+		if addresses != nil {
+			r.updateClusters(v, addresses)
+			go r.cache.ClusterCache.Flush()
```
hzxuzhonghu, line 110: Is this right? You are getting one domain resolved address and then flush the cluster

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+func (r *dnsController) refreshAdsWorker(stop <-chan struct{}) {
+	for {
+		select {
+		case <-stop:
+			return
+		default:
+			domain := <-r.dnsResolver.DnsChan
+			pendingDomain := r.getClustersByDomain(domain)
+			addrs := r.dnsResolver.GetDNSAddresses(domain)
+			r.updateClusters(pendingDomain, addrs)
```
hzxuzhonghu, line 134: How many go routines aare there callling cluster flush? There is already one above in `resolveDomains`

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+	"kmesh.net/kmesh/pkg/dns"
+)
+
+// adsDnsResolver is DNS resolver of Kernel Native
+type dnsController struct {
+	Clusters    chan []*clusterv3.Cluster
+	cache       *AdsCache
+	dnsResolver *dns.DNSResolver
+	// Store the copy of pendingResolveDomain.
+	clusterCache map[string]*pendingResolveDomain
```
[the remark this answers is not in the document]
LiZhenCheng9527: I delete it now

pkg/controller/ads/ads_controller.go
```diff
@@ -44,8 +45,17 @@ type connection struct {
 }
 
 func NewController(bpfAds *bpfads.BpfAds) *Controller {
+	processor := newProcessor(bpfAds)
+	// create kernel-native mode ads resolver controller
+	dnsResolverController, err := NewDnsResolver(processor.Cache)
+	if err != nil {
+		log.Errorf("dns resolver of Kernel-Native mode create failed: %v", err)
+	}
+	processor.DnsResolverChan = dnsResolverController.Clusters
```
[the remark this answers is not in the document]
LiZhenCheng9527:
```
func (r *dnsController) startDnsController() {
	for clusters := range r.Clusters {
		r.resolveDomains(clusters)
	}
}
```
This function used `processor.DnsResolverChan`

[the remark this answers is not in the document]
LiZhenCheng9527:
```
func (r *dnsController) startDnsController() {
	for clusters := range r.Clusters {
		r.getDomains(clusters)
	}
}
```
This function uses `[]*cluster` in `processor.DnsResolverChan`

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+	go r.startDnsController()
+	go func() {
+		<-stopCh
+		close(r.Clusters)
+	}()
+}
+
+func (r *dnsController) startDnsController() {
+	for clusters := range r.Clusters {
+		r.resolveDomains(clusters)
```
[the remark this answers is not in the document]
LiZhenCheng9527: Yes. I will change function name.

pkg/controller/ads/dns.go
LiZhenCheng9527:
```
// Update clusters based on the data in the dns cache.
	for k, v := range domains {
		addresses := r.dnsResolver.GetDNSAddresses(k)
		// Already have record in dns cache
		if addresses != nil {
			r.updateClusters(v, k, addresses)
			go r.cache.ClusterCache.Flush()
		} else {
			// Initialize the newly added hostname
			// and add it to the dns queue to be resolved.
			r.dnsResolver.InitializeDomainInCache(k)
			domainInfo := &dns.DomainInfo{
				Domain:      k,
				RefreshRate: v.RefreshRate,
			}
			r.dnsResolver.ScheduleDomainRefresh(domainInfo, 0)
		}
	}
```
If addresses can be obtained, it means the domain is already in the queue. If addresses cannot be retrieved from the cache, initialize domainInfo and add it to the refresh queue.

pkg/controller/ads/dns.go
LiZhenCheng9527: I changed it to `pendingHostnamesByCluster`

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,317 @@
...
+	}, nil
+}
+
+func (r *dnsController) Run(stopCh <-chan struct{}) {
+	// Start dns resolver
+	go r.dnsResolver.StartDnsResolver(stopCh)
+	// Handle cds updates when a hostname completes resolution
+	go r.refreshAdsWorker(stopCh)
+	// Consumption of clusters to be resolved.
+	go r.startDnsController()
```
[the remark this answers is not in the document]
LiZhenCheng9527: Already change function name

pkg/controller/ads/ads_controller.go
```diff
@@ -44,8 +45,17 @@ type connection struct {
 }
 
 func NewController(bpfAds *bpfads.BpfAds) *Controller {
+	processor := newProcessor(bpfAds)
+	// create kernel-native mode ads resolver controller
+	dnsResolverController, err := NewDnsController(processor.Cache)
+	if err != nil {
+		log.Errorf("dns resolver of Kernel-Native mode create failed: %v", err)
```
hzxuzhonghu, line 52: Donot you need to exit? Otherwise it may panic

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	core_v2 "kmesh.net/kmesh/api/v2/core"
+	"kmesh.net/kmesh/pkg/dns"
+)
+
+// adsDnsResolver is DNS resolver of Kernel Native
+type dnsController struct {
+	Clusters    chan []*clusterv3.Cluster
```
hzxuzhonghu, line 39: nit： clustersChan

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	dnsResolver *dns.DNSResolver
+	// Store the copy of pendingResolveDomain.
+	clusterCache map[string]*pendingResolveDomain
+	// store all pending hostnames in the clusters
+	pendingHostnames map[string][]string
+	sync.RWMutex
+}
+
+// pending resolve domain info of Kennel-Native Mode,
+// domain name is used for dns resolution
```
hzxuzhonghu, line 50: remove

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	}, nil
+}
+
+func (r *dnsController) Run(stopCh <-chan struct{}) {
+	// Start dns resolver
+	go r.dnsResolver.StartDnsResolver(stopCh)
+	// Handle cds updates
+	go r.refreshWorker(stopCh)
+	// Consumption of clusters.
+	go r.processClusters()
```
hzxuzhonghu, line 77: IIUC， refreshWorker is to overwrite cluster and do bpf update
and processClusters is to recv clusters from the chan, and then get the domains from them. Then send them to the undelying dns resolver, right?

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (r *dnsController) processClusters() {
+	for clusters := range r.Clusters {
+		r.processDomains(clusters)
+	}
+}
+
+func (r *dnsController) processDomains(cds []*clusterv3.Cluster) {
+	domains, hostNames := getPendingResolveDomain(cds)
```
hzxuzhonghu, line 91: Are these two maps with same keys? If so can you pass domains to RemoveUnwatchDomain

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	// delete any scheduled re-resolve for domains we no longer care about
+	r.dnsResolver.RemoveUnwatchDomain(hostNames)
+
+	// Update clusters based on the data in the dns cache.
+	for k, v := range domains {
+		addresses := r.dnsResolver.GetDNSAddresses(k)
+		// Already have record in dns cache
+		if addresses != nil {
+			// k(domain) has been resolved, triggering refreshWorker.
+			r.dnsResolver.DnsChan <- k
```
hzxuzhonghu, line 109: I would suggest not write a same channel in two packages

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		} else {
+			// Initialize the newly added hostname
+			// and add it to the dns queue to be resolved.
+			r.dnsResolver.AddDomainInCache(k)
+			domainInfo := &dns.DomainInfo{
+				Domain:      k,
+				RefreshRate: v.RefreshRate,
+			}
+			r.dnsResolver.AddDomainInQueue(domainInfo, 0)
```
hzxuzhonghu, line 118: can you merge AddDomainInCache and AddDomainInQueue

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		return false
+	}
+	for _, cluster := range pendingDomain.Clusters {
+		ready, newCluster := r.overwriteDnsCluster(cluster, domain, addrs)
+		if ready {
+			if !r.cache.UpdateApiClusterIfExists(core_v2.ApiStatus_UPDATE, newCluster) {
+				log.Debugf("cluster: %s is deleted", cluster.Name)
+				return false
+			} else {
+				isClusterUpdate = true
```
hzxuzhonghu, line 153: what does the return value mean

hzxuzhonghu: You may want to return true directkly here

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	isClusterUpdate := false
+	if pendingDomain == nil || addrs == nil {
+		return false
+	}
+	for _, cluster := range pendingDomain.Clusters {
+		ready, newCluster := r.overwriteDnsCluster(cluster, domain, addrs)
+		if ready {
+			if !r.cache.UpdateApiClusterIfExists(core_v2.ApiStatus_UPDATE, newCluster) {
+				log.Debugf("cluster: %s is deleted", cluster.Name)
+				return false
```
hzxuzhonghu, line 151: If you return, then it has no chance to iterate other cluster, isn't this wrong?

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}
+
+	return domains, hostNames
+}
+
+func (r *dnsController) newClusterCache() {
+	if r.clusterCache != nil {
+		r.Lock()
+		defer r.Unlock()
```
hzxuzhonghu, line 299: I think you need to move before r.clusterCache

pkg/controller/ads/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		log.Debug("clean up dns clusters")
+		r.clusterCache = map[string]*pendingResolveDomain{}
+		return
+	}
+}
+
+func (r *dnsController) getClustersByDomain(domain string) *pendingResolveDomain {
+	if r.clusterCache != nil {
+		r.RLock()
+		defer r.RUnlock()
```
hzxuzhonghu, line 309: ditto

pkg/controller/ads/ads_controller.go
LiZhenCheng9527: Has been modified

pkg/controller/ads/dns.go
LiZhenCheng9527: Yes

pkg/controller/ads/dns.go
LiZhenCheng9527: No.
I need to traverse the entire pendingDomain.Clusters, and when a cluster can update, the cluster cache should be flushed. 
When a cluster can be updated, if the system directly returns and does not continue to check the subsequent clusters, After that, the cluster refresh can be triggered only after their hostnames are parsed again.

pkg/controller/ads/dns.go
LiZhenCheng9527: Done

pkg/controller/ads/dns.go
LiZhenCheng9527: Done

pkg/controller/ads/dns.go
LiZhenCheng9527: Done

pkg/controller/ads/dns.go
LiZhenCheng9527: Done

hzxuzhonghu (approved):
/lgtm

PR #1256, opened 2025-02-27
optimizie xdp auth

### LINKED ISSUE #1207: `kmeshctl authz` can't trigger offload authorization enable
**What happened**:

I tried to write a E2E test for offload authorization #641, but it would fail sometimes.

For the failed case, it seems that `kmeshctl authz` can never trigger the offload authorizaiton unless Kmesh is reinstalled.

`kmeshctl authz` should always work fine.

```bash
root@dev:~/kmesh# ./test/e2e/run_test.sh --skip-install-dep --skip-build -run "TestAuthorizationL4"
```

Based on PR #641 , continue to run the above command, which is used to reinstall Kmesh and run `TestAuthorizationL4`. When the test case failed, the environment can no longer use `kmeshctl authz` to trigger offload authorization.

**Environment**:

- Kmesh version:
- Istio version:
- Kernel version:
- Others:

weli-l:
**What type of PR is this?**

Fixes #1207 
```
Previously, since the xdp program only supports ip and port, policies configured with ns and principal types need to be
switched to user mode. However, the logic for determining whether to switch to user mode was wrong. The previous logic
was that when policy_index is greater than `MAX_MEMBER_NUM_PER_POLICY`, it will determine whether to switch to
user mode. However, `MAX_MEMBER_NUM_PER_POLICY` is a fixed value, and the number of deployed policies may be
less than this number.
The correct logic should be to traverse all policies in xdp. In each policy, if the policy does not match, the result of authz
will be stored in `match_ctx->auth_result`. If all policies are traversed and ns or principal rules are deployed, switch to
user mode. If it is not necessary to switch to user mode, return directly according to `match_ctx->auth_result`
```

###authz explanation and cases
#### B → C Traffic Authorization Policy Result

| Case  | Policy Scope             | Policy Rules          | Result  | Logic Description                   |
|-------|--------------------------|-----------------------|---------|--------------------------------------|
| 1     | Single B-targeted       | `allow B`             | PASS    | Direct match                        |
| 2     | Single B-targeted       | `deny B`              | DROP    | Direct match                        |
| 3     | Single non-B target     | `allow A`             | DROP    | Non-target defaults to deny         |
| 4     | Single non-B target     | `deny A`              | PASS    | Non-target defaults to allow        |
| 5     | Mixed (contains B)      | `allow B` + `deny A`  | PASS    | B-target priority                   |
| 6     | Mixed (contains B)      | `deny B` + `allow A`  | DROP    | B-target priority                   |
| 7     | Multi-B (ordered)       | `allow B` → `deny B`  | PASS    | First-match execution               |
| 8     | Multi-B (ordered)       | `deny B` → `allow B`  | DROP    | First-match execution               |

```release-note

```

### DIFF
--- a/bpf/kmesh/workload/include/authz.h
+++ b/bpf/kmesh/workload/include/authz.h
@@ -574,6 +574,9 @@ int policies_check(struct xdp_md *ctx)
     }
     policy = map_lookup_authz(policyId);
     if (!policy) {
+        // Currently, authz in xdp only support ip and port,
+        // if any principal or namespace type policy is configured,
+        // we need to tailcall to userspace.
         if (match_ctx->need_tailcall_to_userspace) {
             bpf_tail_call(ctx, &map_of_xdp_tailcall, TAIL_CALL_AUTH_IN_USER_SPACE);
             return XDP_PASS;
--- a/bpf/kmesh/workload/include/authz.h
+++ b/bpf/kmesh/workload/include/authz.h
@@ -35,6 +35,7 @@ struct match_context {
     __u8 policy_index;
     bool need_tailcall_to_userspace;
     __u8 n_rules;
+    int auth_result;
     wl_policies_v *policies;
     void *rulesPtr;
 };
@@ -573,7 +574,11 @@ int policies_check(struct xdp_md *ctx)
     }
     policy = map_lookup_authz(policyId);
     if (!policy) {
-        return XDP_PASS;
+        if (match_ctx->need_tailcall_to_userspace) {
+            bpf_tail_call(ctx, &map_of_xdp_tailcall, TAIL_CALL_AUTH_IN_USER_SPACE);
+            return XDP_PASS;
+        }
+        return match_ctx->auth_result;
     } else {
         rulesPtr = KMESH_GET_PTR_VAL(policy->rules, void *);
         if (!rulesPtr) {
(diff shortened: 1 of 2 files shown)


supercharge-xsy:
/lgtm
and we need to sort out the current authentication specifications, especially in the case of multiple policies.

hzxuzhonghu:
can you add some description about what this pr is doning

bpf/kmesh/workload/xdp.c
```diff
@@ -132,6 +132,7 @@ int xdp_authz(struct xdp_md *ctx)
         match_ctx.policies = policies;
         match_ctx.need_tailcall_to_userspace = false;
         match_ctx.policy_index = 0;
+        match_ctx.auth_result = XDP_PASS;
```
hzxuzhonghu, line 135: Wy need pass this? I kind of feel this is the default value. Not needed to pass

bpf/kmesh/workload/include/authz.h
```diff
@@ -672,15 +677,10 @@ int policy_check(struct xdp_md *ctx)
...
     }
-
-    match_ctx->policy_index++;
-    if (match_ctx->policy_index >= MAX_MEMBER_NUM_PER_POLICY) {
-        if (match_ctx->need_tailcall_to_userspace) {
-            bpf_tail_call(ctx, &map_of_xdp_tailcall, TAIL_CALL_AUTH_IN_USER_SPACE);
-            return XDP_PASS;
-        }
-        return XDP_PASS;
+    if (match_ctx->auth_result == XDP_PASS) {
```
hzxuzhonghu, line 680: Do you ever set it to XDP_DROP? Can you point it

weli-l: in Line 681,  if unmatched, authz result is stored temporarily
match_ctx->auth_result = match_ctx->action == ISTIO__SECURITY__ACTION__DENY ? XDP_PASS : XDP_DROP;

hzxuzhonghu (approved):
/lgtm

PR #1275, opened 2025-03-13
add channel for Rbac when Kmesh restart

### LINKED ISSUE #1265: Userspace auth incorrectly deny connection when Kmesh restart.
**What happened**:

When I try to fix #1192 , I find a lot of 503 issues: 

```bash
    restart_test.go:66: Minimum success threshold, 1.000000, was not met. 8590/8593 (0.999651) requests failed: 3 errors occurred:
                * request 5681: failed calling enrolled-to-kmesh (cluster=cluster-0)->'http://service-with-waypoint-at-service-granularity.echo-1-36558.svc.cluster.local:80': call failed from enrolled-to-kmesh (cluster=cluster-0) to http://service-with-waypoint-at-service-granularity.echo-1-36558.svc.cluster.local:80 (using http): response[0]: expected response code `200`, got "503". Response: RawContent:       [0] Url=http://service-with-waypoint-at-service-granularity.echo-1-36558.svc.cluster.local:80
        [0] SourceIP=fd00:10:244:1::6
        [0] Latency=228.355062ms
        [0] ActiveRequests=1
        [0] StatusCode=503
        [0] ResponseHeader=Content-Length:95
        [0] ResponseHeader=Content-Type:text/plain
        [0] ResponseHeader=Date:Wed, 05 Mar 2025 07:50:04 GMT
        [0] ResponseHeader=Server:envoy
        [0] ResponseHeader=X-Envoy-Decorator-Operation:service-with-waypoint-at-service-granularity.echo-1-36558.svc.cluster.local:80/*
        [0] ResponseHeader=X-Envoy-Upstream-Service-Time:209
        [0 body] upstream connect error or disconnect/reset before headers. reset reason: connection termination
        
        ID:               
        Method:           
        Protocol:         
        Alpn:             
        URL:              
        Version:          
        Port:             
        Code:             503
        Host:             
        Hostname:         
        Cluster:          
        IstioVersion:     
        IP:               fd00:10:244:1::6
        Request Headers:  map[]
        Response Headers: map[Content-Length:[95] Content-Type:[text/plain] Date:[Wed, 05 Mar 2025 07:50:04 GMT] Server:[envoy] X-Envoy-Decorator-Operation:[service-with-waypoint-at-service-granularity.echo-1-36558.svc.cluster.local:80/*] X-Envoy-Upstream-Service-Time:[209]]
```

By capturing the packet, I found that the waypoint received a TCP RST. After investigation, it was because Kmesh RBAC rule was triggered.

Although I didn't deploy any auth policy, XDP will still pass the packet to the userspace for checking. 

If Kmesh has just been restarted and has not yet sync with istio, it may cause rbac deny, ref: https://github.com/kmesh-net/kmesh/blob/main/pkg/auth/rbac.go#L184 

If no auth policy is configured, then xdp should not send the packet to user space for rbac check?

And we should also consider the situation that if Kmesh is restarted and has not yet sync completely with istio, but auth processing is triggered.

cc  @supercharge-xsy @weli-l 

**Environment**:

- Kmesh version:
- Istio version:
- Kernel version:
- Others:

weli-l:
**What type of PR is this?**

Fixes #1265 

When Kmesh is restarted, one thing to ensure is that the workload's xds configuration has been delivered before executing Rbac, otherwise, Rbac will reject the app request.

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_controller.go
+++ b/pkg/controller/workload/workload_controller.go
@@ -66,7 +66,20 @@ func NewController(bpfWorkload *bpfwl.BpfWorkload, enableMonitoring, enablePerfM
 }
 
 func (c *Controller) Run(ctx context.Context) {
-	go c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
+	go func() {
+		var addressReady, authzReady bool
+		for {
+			select {
+			case <-c.Processor.addressDone:
+				addressReady = true
+			case <-c.Processor.authzDone:
+				authzReady = true
+			}
+			if addressReady && authzReady {
+				go c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
+			}
+		}
+	}()
 	go c.MetricController.Run(ctx, c.bpfWorkloadObj.SockConn.KmTcpProbe)
 	if c.MapMetricController != nil {
 		go c.MapMetricController.Run(ctx)
--- a/pkg/controller/workload/workload_controller.go
+++ b/pkg/controller/workload/workload_controller.go
@@ -68,16 +68,16 @@ func NewController(bpfWorkload *bpfwl.BpfWorkload, enableMonitoring, enablePerfM
 func (c *Controller) Run(ctx context.Context) {
 	go func() {
 		var addressReady, authzReady bool
-		for {
-			select {
-			case <-c.Processor.addressDone:
-				addressReady = true
-			case <-c.Processor.authzDone:
-				authzReady = true
-			}
-			if addressReady && authzReady {
-				go c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
-			}
+		select {
+		case <-c.Processor.addressDone:
+			addressReady = true
+			close(c.Processor.addressDone)
+		case <-c.Processor.authzDone:
+			authzReady = true
+			close(c.Processor.authzDone)
+		}
+		if addressReady && authzReady {
+			go c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
 		}
 	}()
 	go c.MetricController.Run(ctx, c.bpfWorkloadObj.SockConn.KmTcpProbe)
--- a/pkg/controller/workload/workload_controller.go
+++ b/pkg/controller/workload/workload_controller.go
@@ -19,6 +19,7 @@ package workload
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	discoveryv3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/workload/workload_controller.go
```diff
@@ -66,7 +66,20 @@ func NewController(bpfWorkload *bpfwl.BpfWorkload, enableMonitoring, enablePerfM
 }
 
 func (c *Controller) Run(ctx context.Context) {
-	go c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
+	go func() {
+		var addressReady, authzReady bool
+		for {
```
hzxuzhonghu, line 71: I would suggest removing the for loop

<- c.Processor.addressDone

<- c.Processor.authzDone

BTW, as we discussed if no authpolicy set, will we receive the AuthorizationType response?

pkg/controller/workload/workload_controller.go
```diff
@@ -66,7 +66,20 @@ func NewController(bpfWorkload *bpfwl.BpfWorkload, enableMonitoring, enablePerfM
...
+			select {
+			case <-c.Processor.addressDone:
+				addressReady = true
+			case <-c.Processor.authzDone:
+				authzReady = true
+			}
+			if addressReady && authzReady {
+				go c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
+			}
```
YaoZengzeng, line 80: break loop once all ready?

pkg/controller/workload/workload_processor.go
```diff
@@ -115,8 +121,10 @@ func (p *Processor) processWorkloadResponse(rsp *service_discovery_v3.DeltaDisco
 	switch rsp.GetTypeUrl() {
 	case AddressType:
 		err = p.handleAddressTypeResponse(rsp)
+		p.addressDone <- struct{}{}
```
YaoZengzeng, line 124: Does getting an element mean that this type of resource has all been loaded?

pkg/controller/workload/workload_processor.go
```diff
@@ -115,8 +121,10 @@ func (p *Processor) processWorkloadResponse(rsp *service_discovery_v3.DeltaDisco
...
 	case AuthorizationType:
 		err = p.handleAuthorizationTypeResponse(rsp, rbac)
+		p.authzDone <- struct{}{}
```
YaoZengzeng, line 127: ditto

pkg/controller/workload/workload_controller.go
weli-l: I think if we no authpolicy is set, there is no need to start rbac module?

pkg/controller/workload/workload_processor.go
weli-l: > Does getting an element mean that this type of resource has all been loaded?

it means that this type of resource has been received at least once, but it does not mean that all resources have been loaded.

pkg/controller/workload/workload_controller.go
weli-l: > I would suggest removing the for loop
> 
> <- c.Processor.addressDone
> 
> <- c.Processor.authzDone
> 
> BTW, as we discussed if no authpolicy set, will we receive the AuthorizationType response?

If the for loop is removed, the signal can only be consumed once and subsequent signals will be blocked.

Kuromesi: Maybe if we write like this can remove the for loop?

```go
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		<-c.Processor.addressDone
		wg.Done()
	}()
	go func() {
		<-c.Processor.authzDone
		wg.Done()
	}()
	go func() {
		wg.Wait()
		c.Rbac.Run(ctx, c.bpfWorkloadObj.SockOps.KmAuthReq, c.bpfWorkloadObj.XdpAuth.KmAuthRes)
	}()
```

pkg/controller/workload/workload_processor.go
Kuromesi: If in that case, is it better to think that the xds has synced if no response received in a period of time?

pkg/controller/workload/workload_controller.go
hzxuzhonghu: @weli-l If you use the unbuffered channel, @Kuromesi gave a good guide,


>BTW, as we discussed if no authpolicy set, will we receive the AuthorizationType response?

How could you know there is not auth policy set beforehand? A policy can be created at ant time

hzxuzhonghu (approved):
/lgtm

Issue #1146, opened 2024-12-26
Add e2e for locality aware lb 


hzxuzhonghu:
Kmesh has supported locality lb FYI https://kmesh.net/en/docs/userguide/locality_loadbalance/

However it lacks e2e tests now

Murdock9803:
@hzxuzhonghu I would like to solve this issue. Please assign the issue to me 🙌

hzxuzhonghu:
Thank you. First you can try to learn about this feature, and then should learn how to write e2e case

/assign @Murdock9803

Murdock9803:
@hzxuzhonghu Thank you fir this... I will do as you said, and will try to come up with a good PR soon 🙌

7h3-3mp7y-m4n:
Hey @Murdock9803, are you working on this? If this is stale, I’d love to pick it up. Let me know. Thanks!

Murdock9803:
@7h3-3mp7y-m4n Yes you can work on this 🙌

7h3-3mp7y-m4n:
Thank you, @Murdock9803 and @hzxuzhonghu! I’ll go through the docs and raise the PR as soon as possible. In the meantime, @hzxuzhonghu, would it be possible to connect with you on Slack? That way, I can reach out for help if I run into any issues.

Murdock9803:
@7h3-3mp7y-m4n you can post your doubts on the group chat on Slack, so that everyone can get benefit fron this. You can find the slack invite link from kmesh.net site :raised_hands:

7h3-3mp7y-m4n:
thanks @Murdock9803!

harish2773:
Hi @hzxuzhonghu  @7h3-3mp7y-m4n  @Murdock9803 , I’d love to work on this issue if no one else is already handling it. Please let me know if that would be okay.
Thanks!

7h3-3mp7y-m4n:
for sure @harish2773  all the best!

PR #1277, opened 2025-03-14
Add E2E test for Locality Load Balancing (PreferClose mode)

ravjot07:
**Overview:**

This PR adds an end-to-end test for the Locality Load Balancing feature in Kmesh. The test is designed to validate that when the service’s traffic distribution is set to PreferClose, requests from a sleep client are routed to the local instance; and if the local instance is removed, traffic is automatically failed over to a remote instance.

**Test Steps:**

1.  **Namespace Creation:**  
    A dedicated namespace ("sample") is created for testing.
    
2.  **Service Deployment:**  
    A Service ("helloworld") is deployed with the `trafficDistribution` field set to `PreferClose`.
    
3.  **Deployments:**  
    Three Deployments of the helloworld application are deployed:
    
    -   **Local Instance:** On node `ambient-worker` with version `region.zone1.subzone1`.
    -   **Remote Instance 1:** On node `ambient-worker2` with version `region.zone1.subzone2`.
    -   **Remote Instance 2:** On node `ambient-worker3` with version `region.zone2.subzone3`.
4.  **Client Deployment:**  
    A sleep client is deployed on `ambient-worker`.
    
5.  **Verification:**
    
    -   Initially, the sleep client’s request (using curl) is expected to be served by the local instance.
    -   Then, the local deployment is deleted, and the test waits for failover traffic (from the remote instance).
6.  **Cleanup:**  
    All resources are cleaned up at the end of the test.

Contributes toward #1146

### DIFF
--- a/test/e2e/locality_lb_test.go
+++ b/test/e2e/locality_lb_test.go
@@ -1,0 +1,290 @@
+//go:build integ
+// +build integ
+
+/*
+Production-grade End-to-End Test for Kmesh Locality Load Balancing (PreferClose mode).
+
+This test performs the following steps:
+1. Create a dedicated namespace "sample".
+2. Deploy a Service ("helloworld") annotated with PreferClose load balancing.
+3. Deploy three helloworld Deployments on three different nodes:
+   - A local instance on node "ambient-worker" with version "region.zone1.subzone1".
+   - A remote instance on node "ambient-worker2" with version "region.zone1.subzone2".
+   - A remote instance on node "ambient-worker3" with version "region.zone2.subzone3".
+4. Deploy a sleep client on node "ambient-worker".
+5. Verify that:
+   - Initially, the sleep client’s request is served by the local instance.
+   - After deleting the local instance, traffic fails over to the remote instance.
+6. Cleanup all resources at the end of the test.
+*/
+
+package kmesh
+
+import (
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/shell"
+	"istio.io/istio/pkg/test/util/retry"
+)
+
+// applyManifest writes the provided manifest into a temporary file and applies it using kubectl.
+func applyManifest(ns, manifest string) error {
+	tmpFile, err := os.CreateTemp("", "manifest-*.yaml")
+	if err != nil {
+		return err
+	}
+	// Ensure the temporary file is removed after use.
+	defer os.Remove(tmpFile.Name())
+
+	// Write the manifest to the temporary file.
+	if _, err := tmpFile.WriteString(manifest); err != nil {
+		tmpFile.Close()
+		return err
+	}
+	tmpFile.Close()
+
+	cmd := "kubectl apply -n " + ns + " -f " + tmpFile.Name()
+	_, err = shell.Execute(true, cmd)
+	return err
+}
+
+func TestLocalityLoadBalancing(t *testing.T) {
+	// Declare err at the beginning so it can be reused throughout the test.
+	var err error
+
+	framework.NewTest(t).Run(func(t framework.TestContext) {
+		const ns = "sample"
+
+		// Create the test namespace "sample". Log but ignore errors if it already exists.
+		if _, err := shell.Execute(true, "kubectl create namespace "+ns); err != nil {
+			t.Logf("Namespace %s might already exist: %v", ns, err)
+		}
+
+		// Apply the Service manifest with PreferClose load balancing.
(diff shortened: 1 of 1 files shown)


test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,289 @@
...
+      - name: helloworld
+        env:
+        - name: SERVICE_VERSION
+          value: region.zone1.subzone1
+        image: docker.io/istio/examples-helloworld-v1
+        imagePullPolicy: IfNotPresent
+        ports:
+        - containerPort: 5000
+      nodeSelector:
+        kubernetes.io/hostname: ambient-worker
```
YaoZengzeng, line 118: We only have two nodes in E2E cluster: `kmesh-testing-control-plane` and `kmesh-testing-worker`.

hzxuzhonghu: Can we add two worker nodes, otherwise it may need to set pod tolerations to allow schedule on master node

ravjot07: > We only have two nodes in E2E cluster: `kmesh-testing-control-plane` and `kmesh-testing-worker`.

Yup, i have updated it accordingly. I missed it initially.......

test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,299 @@
...
+	"time"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/shell"
+	"istio.io/istio/pkg/test/util/retry"
+)
+
+// applyManifest writes the provided manifest into a temporary file and applies it using kubectl.
+func applyManifest(ns, manifest string) error {
+	tmpFile, err := os.CreateTemp("", "manifest-*.yaml")
```
hzxuzhonghu, line 34: can you change `manifest-*.yaml`

test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpFile.Name())
+
+	if _, err := tmpFile.WriteString(manifest); err != nil {
+		tmpFile.Close()
+		return err
+	}
+	tmpFile.Close()
```
hzxuzhonghu, line 44: move before L38, actually we can make both close and remove called within a defer function

test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	cmd := "kubectl apply -n " + ns + " -f " + tmpFile.Name()
```
hzxuzhonghu, line 46: @YaoZengzeng please point which pkg to use instead of via kubectl

test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,299 @@
...
+		t.Logf("Pods after deployment in namespace %s:\n%s", ns, pods)
+		endpoints, _ = shell.Execute(true, "kubectl get endpoints helloworld -n "+ns)
+		t.Logf("Endpoints for service helloworld after deployment:\n%s", endpoints)
+
+		// Check DNS resolution from the sleep pod.
+		sleepPod, err := shell.Execute(true, "kubectl get pod -n "+ns+" -l app=sleep -o jsonpath='{.items[0].metadata.name}'")
+		if err != nil || sleepPod == "" {
+			t.Fatalf("Failed to get sleep pod: %v", err)
+		}
+		nslookup, _ := shell.Execute(true, "kubectl exec -n "+ns+" "+sleepPod+" -- nslookup "+fqdn)
```
hzxuzhonghu, line 244: You donot need to nslookup the ip addr

test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,299 @@
...
+		if err := retry.Until(func() bool {
+			t.Logf("Attempting curl request at %s...", time.Now().Format(time.RFC3339))
+			out, execErr := shell.Execute(true,
+				"kubectl exec -n "+ns+" "+sleepPod+" -- curl -v -sSL --resolve "+fqdn+":5000:"+resolvedIP+" http://"+fqdn+":5000/hello")
+			if execErr != nil {
+				t.Logf("Curl error: %v", execErr)
+				return false
+			}
+			t.Logf("Curl output: %s", out)
+			if strings.Contains(out, "region.zone1.subzone1") {
```
hzxuzhonghu, line 264: This test is not right, here you return after checking only `region.zone1.subzone1` response.

YaoZengzeng:
@ravjot07 Try to fix the comments and get this PR merged.

Get code merged is what really matters.

PR #1278, opened 2025-03-15
add e2e test for locality aware load balancer

harish2773:
**What type of PR is this?**  
/kind feature  

This PR adds an end-to-end test for Locality-Aware Load Balancing. It verifies that service traffic is correctly routed to instances in the nearest locality based on region/zone/subzone topology.  

Contributes towards #1146 

**Test Steps:**  

1. **Namespace Setup:**  
   - A separate namespace (`sample`) is created to isolate the test environment.  

2. **Application Deployment:**  
   - Three instances of the `helloworld` application are deployed on different nodes to simulate locality-based traffic handling:  
     - **Local Instance:** Deployed on `ambient-worker` with the version `region.zone1.subzone1`.  
     - **Remote Instance 1:** Deployed on `ambient-worker2` with the version `region.zone1.subzone2`.  
     - **Remote Instance 2:** Deployed on `ambient-worker3` with the version `region.zone2.subzone3`.  

3. **Client Setup:**  
   - A `sleep` client is deployed on `ambient-worker` to generate test requests.  

4. **Validation:**  
   - The test verifies the following:  
     - Initially, the sleep client’s request (using `curl`) is handled by the local instance (`region.zone1.subzone1`).  
     - After terminating the local instance, the test ensures that traffic shifts to the nearest available instance based on locality (either remote instance 1 or 2).  

5. **Cleanup:**  
   - All test resources (namespace, deployments, and services) are deleted after the test completes.  
 

NONE

### DIFF
--- a/test/e2e/locality_aware_lb_test.go
+++ b/test/e2e/locality_aware_lb_test.go
@@ -1,0 +1,228 @@
+package e2e
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+)
+
+func runCommand(command string) (string, error) {
+	cmd := exec.Command("bash", "-c", command)
+	out, err := cmd.CombinedOutput()
+	return string(out), err
+}
+
+func TestLocalityAwareLoadBalancing(t *testing.T) {
+	t.Run("Setup Kind Cluster", func(t *testing.T) {
+		t.Log("Setting up Kind cluster...")
+		out, err := runCommand(`kind delete clusters --all || true &&
+kind create cluster --image=kindest/node:v1.31.0 --config=- <<EOF
+kind: Cluster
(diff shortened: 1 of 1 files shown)


test/e2e/locality_aware_lb_test.go
```diff
@@ -0,0 +1,249 @@
+//go:build integ
+// +build integ
+
+/*
+End-to-End Test for Kmesh Locality-Aware Load Balancing.
+
+This test verifies that Kmesh correctly routes traffic to the nearest available instance 
+based on locality (region, zone, and subzone). The test performs the following steps:
+1. Create a Kind cluster with three worker nodes.
```
YaoZengzeng, line 9: It's better to test in an existing E2E cluster ranther than creating a new Kind cluster.

Issue #1339, opened 2025-04-25
unexpected panic when handle workload in kmesh daemon


Kuromesi:
**What happened**:
I'm trying to find out the root cause of this panic, not sure why workload addresses are empty.
```
panic: runtime error: index out of range [0] with length 0

goroutine 48 [running]:
kmesh.net/kmesh/pkg/controller/workload.(*Processor).handleWorkload(0xc000e91550, 0xc000a6d680)
	kmesh.net/kmesh/pkg/controller/workload/workload_processor.go:507 +0x731
kmesh.net/kmesh/pkg/controller/workload.(*Processor).handleServicesAndWorkloads(0xc000e91550, {0xc000e8c600, 0x1f, 0x0?}, {0xc0005f1688?, 0xc000e0ed70?, 0x55992522aa80?})
	kmesh.net/kmesh/pkg/controller/workload/workload_processor.go:849 +0x490
kmesh.net/kmesh/pkg/controller/workload.(*Processor).handleAddressTypeResponse(0xc000e91550, 0xc001000d10)
	kmesh.net/kmesh/pkg/controller/workload/workload_processor.go:821 +0x415
kmesh.net/kmesh/pkg/controller/workload.(*Processor).processWorkloadResponse(0xc000e91550, 0xc001000d10, 0xc000f29820)
	kmesh.net/kmesh/pkg/controller/workload/workload_processor.go:117 +0x137
kmesh.net/kmesh/pkg/controller/workload.(*Controller).HandleWorkloadStream(0xc000ef38c0)
	kmesh.net/kmesh/pkg/controller/workload/workload_controller.go:130 +0xac
kmesh.net/kmesh/pkg/controller.(*XdsClient).handleUpstream(0xc000efc540, {0x5599252825a0, 0xc000f99320})
	kmesh.net/kmesh/pkg/controller/client.go:131 +0x111
created by kmesh.net/kmesh/pkg/controller.(*XdsClient).Run in goroutine 1
	kmesh.net/kmesh/pkg/controller/client.go:149 +0xd2
```

```go
+507 if !slices.Equal(newWorkloadAddresses[0], oldWorkloadAddresses[0]) {
			err := p.deleteFrontendByIp(oldWorkloadAddresses)
			if err != nil {
				return fmt.Errorf("frontend map delete failed: %v", err)
			}
		}
```
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

[...]

PR #1345, opened 2025-04-27
fix workload processer panic if newWorkload addresses is nil when add serviceEntry

LiZhenCheng9527:
… serviceEntry

**What type of PR is this?**
/kind bug

a part of #1339 

```release-note

```

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -864,6 +864,12 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service, 
 	}
 
 	for _, workload := range workloads {
+		// TODO: Kmesh supports ServiceEntry
+		if workload.GetAddresses() == nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/workload/workload_processor.go
```diff
@@ -520,6 +520,11 @@ func (p *Processor) handleWorkload(workload *workloadapi.Workload) error {
 		newWorkloadAddresses := workload.GetAddresses()
 		oldWorkloadAddresses := oldWorkload.GetAddresses()
 
+		if len(newWorkloadAddresses) == 0 {
+			log.Warnf("new workload address: %s/%s is nil", workload.Namespace, workload.Name)
```
hzxuzhonghu, line 524: I feel should filter workload without address in the entrance since we donot support dns service in dualengine now?

hzxuzhonghu (approved):
/lgtm

We can support it later

PR #1347, opened 2025-04-28
Fix kmesh manage, especially after kmesh restart, old xdp prog maybe …

### LINKED ISSUE #1346: old xdp linked after kmesh daemon restart
**What happened**:

After kmesh daemon restarted, i found the authz does not take effect

**Environment**:

- Kmesh version:
- Istio version:
- Kernel version:
- Others:

hzxuzhonghu:
…linked because of the event order

**What type of PR is this?**

/kind bug

Fix #1346

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/manage/manage_controller.go
+++ b/pkg/controller/manage/manage_controller.go
@@ -119,6 +119,9 @@ func NewKmeshManageController(client kubernetes.Interface, sm *kmeshsecurity.Sec
 	}
 
 	if _, err := namespaceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.handleNamespaceAdd(obj)
+		},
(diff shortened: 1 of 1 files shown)


pkg/controller/manage/manage_controller.go
```diff
@@ -138,6 +141,9 @@ func (c *KmeshManageController) handlePodAdd(obj interface{}) {
 
 	namespace, err := c.namespaceLister.Get(newPod.Namespace)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return
```
YaoZengzeng, line 145: Is there any change from before? Other than one less line of logging.

hzxuzhonghu: ?

hzxuzhonghu:
cc @Kuromesi

PR #1363, opened 2025-05-07
fix log cannot print in kernel 5.10 by bpf_trace_printk

lec-bit:
bpf_trace_printk cannot print two %s, need to fix

**What type of PR is this?**
/kind bug

bpf_trace_printk cannot print two %s, need to fix

Fixes #

```release-note

```

### DIFF
--- a/bpf/kmesh/ads/include/cluster.h
+++ b/bpf/kmesh/ads/include/cluster.h
@@ -311,13 +311,9 @@ static inline int cluster_handle_loadbalance(Cluster__Cluster *cluster, address_
         return -EAGAIN;
     }
 
-    BPF_LOG(
-        INFO,
-        CLUSTER,
-        "cluster=\"%s\", loadbalance to addr=[%s:%u]\n",
-        name,
-        ip2str(&sock_addr->ipv4, 1),
-        bpf_ntohs(sock_addr->port));
+    // Since bpf_trace_printk cannot print two %s, the name and ipv4 are printed separately.
+    BPF_LOG(INFO, CLUSTER, "cluster=\"%s\"\n", name);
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
@nlgwcy can you please give some suggestion

bpf/kmesh/ads/include/cluster.h
```diff
@@ -311,13 +311,9 @@ static inline int cluster_handle_loadbalance(Cluster__Cluster *cluster, address_
...
 
-    BPF_LOG(
-        INFO,
-        CLUSTER,
-        "cluster=\"%s\", loadbalance to addr=[%s:%u]\n",
-        name,
-        ip2str(&sock_addr->ipv4, 1),
-        bpf_ntohs(sock_addr->port));
+    // Since bpf_trace_printk cannot print two %s, the name and ipv4 are printed separately.
+    BPF_LOG(INFO, CLUSTER, "cluster=\"%s\"\n", name);
```
nlgwcy, line 315:
```suggestion
    BPF_LOG(INFO, CLUSTER, "cluster=\"%s\",", name);
```

hzxuzhonghu:
Ok, so in 5.10 kernel we must fix it. Is this the right way?

bpf/kmesh/ads/include/cluster.h
lec-bit: BPF logs are printed on a new line each time in bpf_trace_printk. Adding \n here will not affect whether the log is continuous.

hzxuzhonghu:
Can you also fix  bpf/kmesh/workload/xdp.c

hzxuzhonghu:
Seems `xdp_deny_packet` is not used anymore

lec-bit:
> Can you also fix bpf/kmesh/workload/xdp.c


I don't know if this place can be repaired simply by taking it apart.

hzxuzhonghu:
I think this function can be removed now

lec-bit:
> I think this function can be removed now

ok
I see that there is only empty definition and no usage.

PR #1376, opened 2025-05-08
rfac: outputAccesslog func to skip accesslog at conn establishment

### LINKED ISSUE #1375: Do not output accesslog at conn start
**What happened**:

**Environment**:

- Kmesh version:
- Istio version:
- Kernel version:
- Others:

yp969803:
**What type of PR is this?**
=/kind enhancement

outputAccesslog func to skip accesslog at conn establishment

Fixes #1375

```release-note

```

### DIFF
--- a/pkg/controller/telemetry/metric.go
+++ b/pkg/controller/telemetry/metric.go
@@ -38,7 +38,7 @@ import (
 
 const (
 	TCP_ESTABLISHED = uint32(1)
-	TCP_CLOSTED     = uint32(7)
+	TCP_CLOSED      = uint32(7)
 
 	connection_success = uint32(1)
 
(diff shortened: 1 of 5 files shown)


yp969803:
@hzxuzhonghu can u review, it is more better approach

yp969803:
also i haved fixed the type TCP_CLOSTED

pkg/controller/telemetry/metric.go
```diff
@@ -48,7 +48,7 @@ const (
 
 	DEFAULT_UNKNOWN = "-"
 
-	LONG_CONN_METRIC_THRESHOLD = uint64(5 * time.Second)
+	LONG_CONN_METRIC_THRESHOLD = uint64(30 * time.Second)
```
LiZhenCheng9527, line 51: Is there any basis for the threshold modification to 30s?

yp969803: The duration was initially set to 30 seconds, but @hzxuzhonghu reduced it to 5 seconds. In Prometheus, metrics are scraped at fixed intervals, with the maximum scrape interval being 15 seconds. For short-lived connections lasting less than 15 seconds, a connectionMetric might be created and deleted before it can be scraped. This results in unnecessary computation, as the metric is never captured by the user.

yp969803: i have changed it to 5

hzxuzhonghu (approved):
/lgtm

PR #1393, opened 2025-05-12
output kmesh log when e2e test failed

### LINKED ISSUE #1390: Print kmesh daemon logs when e2e tests fail

Print kmesh daemon logs when e2e tests fail, especially Kmesh daemon in the worker node

**Why is this needed**:

Because there are always some flaky cases that are very difficult to reproduce locally, but always appear in CI.

If we can print the log of Kmesh daemon, it will be very helpful for our debugging.

YaoZengzeng:
**What type of PR is this?**

/kind enhancement

Fixes #1390

```release-note

```

### DIFF
--- a/test/e2e/run_test.sh
+++ b/test/e2e/run_test.sh
@@ -137,6 +137,20 @@ function setup_kmesh() {
 		echo "Waiting for pods of Kmesh daemon to enter Running state..."
 		sleep 1
 	done
+
+	# Set log of each Kmesh pods.
+    PODS=$(kubectl get pods -n kmesh-system -l app=kmesh -o jsonpath='{.items[*].metadata.name}')
+
+    for POD in $PODS; do
+        echo $POD
+        kmeshctl log $POD --set bpf:debug
+		kmeshctl log $POD --set default:debug
+    done
+}
+
+function install_kmeshctl() {
+    # Install kmeshctl
+    cp kmeshctl $TMPBIN
 }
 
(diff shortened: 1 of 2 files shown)


YaoZengzeng:
@hzxuzhonghu ptal

test/e2e/run_test.sh
```diff
@@ -11,6 +11,8 @@ DEFAULT_KIND_IMAGE="kindest/node:v1.30.0@sha256:047357ac0cfea04663786a612ba1eaba
 
 ISTIO_VERSION=${ISTIO_VERSION:-"1.22.0"}
 
+LOGFILE="kmesh_daemon.log"
```
hzxuzhonghu, line 14: We cannot see it once CI job running failed

test/e2e/run_test.sh
```diff
@@ -283,9 +375,17 @@ if [[ -z ${SKIP_SETUP:-} ]]; then
...
+
+capture_pod_logs &
+
 cmd="go test -v -tags=integ $ROOT_DIR/test/e2e/... -istio.test.kube.loadbalancer=false ${PARAMS[*]}"
 
-bash -c "$cmd"
+bash -c "$cmd" || {
+	cat $LOGFILE
+	rm $LOGFILE
+	exit 1
```
hzxuzhonghu, line 387: why exit without cleaning up. I think failure does not mean not cleaning up, they are orthogonal

YaoZengzeng:
@hzxuzhonghu updated

PR #1214, opened 2025-01-29
Fix: `make build` should build kmeshctl

adwait-godbole:
Fixes #1113 

/kind bug
/kind enhancement
/kind documentation

```release-note
NONE
```

PR #1419, opened 2025-05-22
Fix: make build should build kmeshctl

### LINKED ISSUE #1113: `make build` should build kmeshctl

Now when running make build, it does not build kmeshctl. We have to run make kmeshctl. I would think make build should build all the components

**Why is this needed**:

Flying-Tom:
**What type of PR is this?**

/kind enhancement

Fixes #1113 

> Follow up of #1214

```release-note
NONE
```


oncn-mda/CMakeLists.txt
YaoZengzeng, line 1: Why are there so many other files modified?

Flying-Tom: maybe file permission problems, fix done.

Issue #1431, opened 2025-06-27
The output of "kmeshctl dump" is unordered


zrggw:
**What would you like to be added**:
 Sort the output of 'kmeshctl dump.
**Why is this needed**:
Because the output of "kmeshctl dump" is unordered, it is inconvenient to compare the results from two different uses of dump. It would be better if it were ordered.

PR #1432, opened 2025-06-27
Sort dump output by name

zrggw:
**What type of PR is this?**
/kind enhancement

#1431 
Because the output of "kmeshctl dump" is unordered, it is inconvenient to compare the results from two different uses of dump. So I sorted it by "name".

```release-note

```

### DIFF
--- a/pkg/controller/security/manager_test.go
+++ b/pkg/controller/security/manager_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/agiledragon/gomonkey/v2"
 	"github.com/stretchr/testify/assert"
 	"istio.io/istio/pkg/security"
+	"istio.io/istio/pkg/test/util/retry"
 
 	camock "kmesh.net/kmesh/pkg/controller/security/mock"
 )
@@ -165,25 +166,27 @@ func runTestretryFetchCert(t *testing.T) {
 	patches2.Reset()
 
 	secretManager.SendCertRequest(identity, RETRY)
-	time.Sleep(2000 * time.Millisecond)
-	for {
-		cert := secretManager.GetCert(identity)
-		if cert != nil {
-			secretManager.certsCache.mu.RLock()
-			hasCert := cert.cert != nil
-			secretManager.certsCache.mu.RUnlock()
-			if hasCert {
-				break
-			}
-		}
-		time.Sleep(100 * time.Millisecond)
-	}
 
-	cert := secretManager.GetCert(identity)
-	assert.NotNil(t, cert)
-	secretManager.certsCache.mu.RLock()
-	assert.NotNil(t, cert.cert)
-	secretManager.certsCache.mu.RUnlock()
+	err = retry.UntilSuccess(
+		func() error {
+			cert := secretManager.GetCert(identity)
+			if cert != nil {
+				secretManager.certsCache.mu.RLock()
+				hasCert := cert.cert != nil
+				secretManager.certsCache.mu.RUnlock()
+				if hasCert {
+					return nil
+				}
+			}
+			return fmt.Errorf("cert not found for identity %s", identity)
+		},
+		retry.Delay(100*time.Millisecond),
+		retry.Timeout(6*time.Second),
+	)
+
+	if err != nil {
+		t.Errorf("Failed to fetch cert after retry: %v", err)
+	}
 
 	close(stopCh)
 }


hzxuzhonghu:
LGTM
 
Can you add test cases to the existing test

hzxuzhonghu:
why do you change the bpf generated file for kernel native mode

pkg/status/status_server_test.go
```diff
@@ -225,20 +237,148 @@ func TestServer_configDumpWorkload(t *testing.T) {
...
+		svc := buildService(fmt.Sprintf("service-%d", i), fmt.Sprintf("hostname-%d", i))
+		services = append(services, svc)
+
+		fakeWorkloadCache.AddOrUpdateWorkload(w)
+		fakeServiceCache.AddOrUpdateService(svc)
 	}
 
-	util.RefreshGoldenFile(t, w.Body.Bytes(), "./testdata/workload_configdump.json")
+	// Create a new HTTP request and response
+	req2 := httptest.NewRequest(http.MethodGet, "/configDumpWorkload", nil)
```
hzxuzhonghu: I feel we donot need to send a second req, mixing the workloads and services at first, we can test getting only once.

zrggw:
> why do you change the bpf generated file for kernel native mode

The generated files under the kernelnative root directory were created incorrectly, and I will delete these files later.

hzxuzhonghu:
>The generated files under the kernelnative root directory were created incorrectly, and I will delete these files later.

@lec-bit Can you confirm?

zrggw:
I revise the code according to the code review results.
1. remove  second and third req.
2. remove rand.
3. remove generated files under the kernelnative root directory.

config/kmesh_marcos_def.h
```diff
@@ -7,22 +7,22 @@
  * loopback address link conflicts. Obtains the namespace cookie of the
  * current container based on the bpf_get_netns_cookie auxiliary function.
  */
-#define MDA_LOOPBACK_ADDR 1
```
LiZhenCheng9527: use `make clean` command clean up if you have build kmesh code locally before pushing your code to github.

pkg/status/status_server_test.go
```diff
@@ -225,20 +234,123 @@ func TestServer_configDumpWorkload(t *testing.T) {
...
+		w := buildWorkload(fmt.Sprintf("workload-%d-modified", i))
+		w.ClusterId = "cluster1" // Changed cluster
+		w.Uid = fmt.Sprintf("cluster1//Pod/ns/workload-%d-modified", i)
+		w.Status = workloadapi.WorkloadStatus_UNHEALTHY
+
+		workloads[i] = w
+		// Modify service properties
+		svc := buildService(fmt.Sprintf("service-%d-modified", i), fmt.Sprintf("hostname-%d-modified", i))
+		// Modify service ports
+		svc.Ports = []*workloadapi.Port{
```
hzxuzhonghu, line 309: why modify  ports, does it influence the order?

pkg/status/status_server_test.go
```diff
@@ -225,20 +234,123 @@ func TestServer_configDumpWorkload(t *testing.T) {
...
+	util.CompareContent(t, w1.Body.Bytes(), "./testdata/workload_configdump.json")
+
+	fakeWorkloadCache = cache.NewWorkloadCache()
+	fakeServiceCache = cache.NewServiceCache()
+
+	workloads := []*workloadapi.Workload{}
+	services := []*workloadapi.Service{}
+
+	for i := 0; i < 10; i++ {
+		w := buildWorkload(fmt.Sprintf("workload-%d", i))
```
hzxuzhonghu, line 259: Isn't this make the workloads alreadt sorted?

pkg/status/status_server_test.go
zrggw: Just to test that the sorting is not affected by other modifications, like port modifications.

pkg/status/status_server_test.go
zrggw: Later, when w is added to fakeWorkloadCache(line 326) and then a workDump is created(line 346 and status_servers.go, line 493), the wokload in workloadDump will be unordered due to the use of **map**.

hzxuzhonghu (approved):

hzxuzhonghu:
please fix the ci

hzxuzhonghu:
Looks this failure is related

pkg/controller/security/manager_test.go
```diff
@@ -160,13 +160,30 @@ func runTestretryFetchCert(t *testing.T) {
...
-	identity1 := "identity1"
 	secretManager.SendCertRequest(identity, ADD)
 	time.Sleep(100 * time.Millisecond)
 	patches2.Reset()
-	secretManager.SendCertRequest(identity1, RETRY)
+
+	secretManager.SendCertRequest(identity, RETRY)
 	time.Sleep(2000 * time.Millisecond)
-	assert.NotNil(t, secretManager.GetCert(identity).cert)
+	for {
```
hzxuzhonghu: nit: please use retry.UntilSuccess, which also have a timeout

And seems we can remove L168 now

zrggw: It seems we can also remove line 182-186?  Because we have already checked whether cert.cert is nil on line 173.

hzxuzhonghu: right

hzxuzhonghu (commented):
/approve

PR #1435, opened 2025-07-09
Fix: make format with too many containers

Flying-Tom:
**What type of PR is this?**

/kind enhancement

`make format` will create a `markdownlint-cli2` container with random name every time, which needs to be fix.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ tools/vagrant/.vagrant/
 .run/
 # Visual Studio Code
 .vscode/
+.local/
 # Bazel
 /bazel-*
 # vi swap files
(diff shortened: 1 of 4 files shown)


yp969803:
@Flying-Tom we can use the --rm flag when starting the container—this automatically removes the container once it exits. Please have a look at this approach

brushax:
> @Flying-Tom we can use the --rm flag when starting the container—this automatically removes the container once it exits. Please have a look at this approach

The previous implementation was mainly because I didn't want to create a new ephemeral container every time. However, using the `--rm` parameter introduces relatively low complexity.

PR #1438, opened 2025-07-10
Feat: Add Workload DnsController

Flying-Tom:
**What type of PR is this?**

/kind feature

Supprort Dns resolve for workload generated by `ServiceEntry` in Dual-Engine mode.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -148,6 +148,7 @@ func (c *Controller) Start(stopCh <-chan struct{}) error {
 	c.client = NewXdsClient(c.mode, c.bpfAdsObj, c.bpfWorkloadObj, c.bpfConfig.EnableMonitoring, c.bpfConfig.EnableProfiling)
 
 	if c.client.WorkloadController != nil {
+		c.client.WorkloadController.StartDnsController(stopCh)
 		c.client.WorkloadController.Run(ctx)
 	} else {
 		c.client.AdsController.StartDnsController(stopCh)
--- a/pkg/controller/workload/dns.go
+++ b/pkg/controller/workload/dns.go
@@ -1,0 +1,230 @@
+package workload
+
+import (
+	"net"
+	"net/netip"
+	"sync"
+	"time"
+
+	"google.golang.org/protobuf/proto"
+	"kmesh.net/kmesh/api/v2/workloadapi"
+	"kmesh.net/kmesh/pkg/controller/workload/cache"
+	"kmesh.net/kmesh/pkg/dns"
+)
+
+type dnsController struct {
+	workloadsChan chan []*workloadapi.Workload
+	cache         cache.WorkloadCache
+	dnsResolver   *dns.DNSResolver
+	// store the copy of pendingResolveWorkload.
+	workloadCache map[string]*pendingResolveDomain
+	// store all pending hostnames in the workloads
+	pendingHostnames map[string][]string
+	sync.RWMutex
+}
+
+// pending resolve domain info of Dual-Engine Mode,
+// workload is used for create the apiworkload
+type pendingResolveDomain struct {
+	Workloads   []*workloadapi.Workload
+	RefreshRate time.Duration
+}
+
+func NewDnsController(cache cache.WorkloadCache) (*dnsController, error) {
+	resolver, err := dns.NewDNSResolver()
+	if err != nil {
+		return nil, err
+	}
+	return &dnsController{
+		workloadsChan:    make(chan []*workloadapi.Workload),
+		cache:            cache,
+		dnsResolver:      resolver,
+		workloadCache:    make(map[string]*pendingResolveDomain),
+		pendingHostnames: make(map[string][]string),
+	}, nil
+}
+
+func (r *dnsController) Run(stopCh <-chan struct{}) {
+	go r.dnsResolver.StartDnsResolver(stopCh)
+	go r.refreshWorker(stopCh)
+	go r.processWorkloads()
+	go func() {
+		<-stopCh
+		close(r.workloadsChan)
+	}()
+}
+
+func (r *dnsController) processWorkloads() {
+	for workloads := range r.workloadsChan {
+		r.processDomains(workloads)
+	}
+}
+
+func (r *dnsController) processDomains(workload []*workloadapi.Workload) {
+	domains := getPendingResolveDomain(workload)
+
+	// store all pending hostnames of clusters in pendingHostnames
+	for _, workload := range workload {
+		workloadName := workload.GetName()
+		hostname := workload.GetHostname()
+		r.pendingHostnames[workloadName] = []string{hostname}
+		r.workloadCache[hostname] = &pendingResolveDomain{
+			Workloads:   []*workloadapi.Workload{workload},
+			RefreshRate: 15 * time.Second,
+		}
+	}
+
+	// delete any scheduled re-resolve for domains we no longer care about
+	r.dnsResolver.RemoveUnwatchDomain(domains)
+
+	// update workloadCache with pendingResolveWorkload
+	for k, v := range domains {
+		addresses := r.dnsResolver.GetDNSAddresses(k)
+		if addresses != nil {
+			go r.updateWorkloads(v.(*pendingResolveDomain), k, addresses)
+		} else {
+			// Initialize the newly added hostname
+			// and add it to the dns queue to be resolved.
+			domainInfo := &dns.DomainInfo{
+				Domain:      k,
+				RefreshRate: v.(*pendingResolveDomain).RefreshRate,
+			}
+			r.dnsResolver.AddDomainInQueue(domainInfo, 0)
+		}
+	}
+}
+
+func (r *dnsController) refreshWorker(stop <-chan struct{}) {
+	for {
+		select {
+		case <-stop:
+			return
+		case domain := <-r.dnsResolver.DnsChan:
+			pendingDomain := r.getWorkloadsByDomain(domain)
+			// log.Infof("dnsController refreshWorker: domain %s, pendingDomain %v", domain, pendingDomain)
+			addrs := r.dnsResolver.GetDNSAddresses(domain)
+			maxRetry := 3
+			for range maxRetry {
+				if len(addrs) > 0 {
+					r.updateWorkloads(pendingDomain, domain, addrs)
+				}
+				time.Sleep(1 * time.Second)
+			}
+		}
+	}
+}
+
+func (r *dnsController) updateWorkloads(pendingDomain *pendingResolveDomain, domain string, addrs []string) {
+	isWorkerUpdate := false
+	if pendingDomain == nil || addrs == nil{
+		return
+	}
+	log.Infof("dnsController updateWorkloads: pendingDomain %v, domain %s, addrs %v", pendingDomain, domain, addrs)
+
+	for _, workload := range pendingDomain.Workloads {
+		if ready, newWorkload := r.overwriteDnsWorkload(workload, domain, addrs); ready {
+			// log.Infof("dnsController update cache for workload %s with addresses %v", workload.ResourceName(), addrs)
+			if r.cache.GetWorkloadByUid(workload.GetUid()) != nil {
+				r.cache.AddOrUpdateWorkload(newWorkload)
+				delete(r.workloadCache, domain)
+				isWorkerUpdate = true
+			}
+		}
+	}
+
+	if isWorkerUpdate {
+		// w.cache.Flush()
+	}
+}
+
+func (r *dnsController) overwriteDnsWorkload(workload *workloadapi.Workload, domain string, addrs []string) (bool, *workloadapi.Workload) {
+	ready := true
+	hostNames := r.pendingHostnames[workload.GetName()]
+	addressesOfHostname := make(map[string][]string)
+
+	for _, hostName := range hostNames {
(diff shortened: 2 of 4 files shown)


pkg/controller/workload/dns.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+func (r *dnsController) overwriteDnsWorkload(workload *workloadapi.Workload, domain string, addrs []string) (bool, *workloadapi.Workload) {
+	ready := true
+	hostNames := r.pendingHostnames[workload.GetName()]
+	addressesOfHostname := make(map[string][]string)
+
+	for _, hostName := range hostNames {
+		// log.Infof("overwriteDnsWorkload: checking hostname %s for workload %s with domain %s", hostName, workload.ResourceName(), domain)
```
LiZhenCheng9527: remove

pkg/controller/workload/workload_processor.go
```diff
@@ -883,11 +884,41 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service,
...
+			// Non-ServiceEntry workload with nil addresses will be ignored.
+			uid := workload.GetUid()
+			if !strings.Contains(uid, "ServiceEntry") {
+				log.Warnf("workload: %s/%s addresses is nil", workload.Namespace, workload.Name)
+				continue
+			} else {
+				log.Warnf("workload: %s/%s addresses is nil, workload info: %+v", workload.Namespace, workload.Name, workload)
+				// workload from service entry need address resolving
+				if p.DnsResolverChan != nil {
+					p.DnsResolverChan <- workloads
```
Kuromesi: Does that mean all workloads will be resolved in every loop?

LiZhenCheng9527:
Can you provide more guidance?

pkg/controller/workload/workload_processor.go
LiZhenCheng9527: @Kuromesi Can you provide more guidance?

Kuromesi: I think we should pass workloads without addresses instead of all workloads to the resolverchan?

pkg/controller/workload/workload_processor.go
```diff
@@ -883,11 +884,41 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service,
...
+				}
+				go func() {
+					maxRetries := 50
+					var address [][]byte
```
Kuromesi: We should decide whether to retry based on the dns response code.

```go
// Message Response Codes, see https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml
RcodeSuccess        = 0  // NoError   - No Error                          [DNS]
RcodeFormatError    = 1  // FormErr   - Format Error                      [DNS]
RcodeServerFailure  = 2  // ServFail  - Server Failure                    [DNS]
RcodeNameError      = 3  // NXDomain  - Non-Existent Domain               [DNS]
```

pkg/controller/workload/workload_processor.go
```diff
@@ -883,11 +884,41 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service,
...
+						workload := p.WorkloadCache.GetWorkloadByUid(uid)
+						if address = workload.GetAddresses(); address != nil {
+							break
+						} else {
+							time.Sleep(WorkloadDnsRefreshRate)
+						}
+					}
+					if address != nil {
+						log.Infof("workload: %s/%s addresses resolved: %v", workload.Namespace, workload.Name, address)
+						if err := p.handleWorkload(workload); err != nil {
```
Kuromesi: Can we pass `handleWorkload` as a callback to the resolver, so we don't have to start a goroutine here?

pkg/controller/workload/workload_processor.go
brushax: > I think we should pass workloads without addresses instead of all workloads to the resolverchan?

I think so, workload passed to resolverchan should be limited.

brushax: But in this case, workloads passed to resolverchan have been filtered with condition 1) without address 2) from ServiceEntry.

pkg/controller/workload/workload_processor.go
brushax: I'm not sure whether this location needs to be blocked, and the existing `pkg/dns` module doesn't seem to expose any identifiable state information. @LiZhenCheng9527

pkg/controller/workload/workload_processor.go
brushax: Same reason as above.

pkg/controller/workload/workload_processor.go
Kuromesi: I'm not quite understand, if we send all workloads to the resolverchan in every loop, why we don't move this logic out the loop or send a single workload in every loop?

pkg/controller/controller.go
```diff
@@ -148,6 +148,7 @@ func (c *Controller) Start(stopCh <-chan struct{}) error {
 	c.client = NewXdsClient(c.mode, c.bpfAdsObj, c.bpfWorkloadObj, c.bpfConfig.EnableMonitoring, c.bpfConfig.EnableProfiling)
 
 	if c.client.WorkloadController != nil {
+		c.client.WorkloadController.StartDnsController(stopCh)
```
hzxuzhonghu: nit: can move StartDnsController into its Run method

pkg/controller/workload/workload_controller.go
```diff
@@ -48,9 +49,17 @@ type Controller struct {
 }
 
 func NewController(bpfWorkload *bpfwl.BpfWorkload, enableMonitoring, enablePerfMonitor bool) *Controller {
+	processor := NewProcessor(bpfWorkload.SockConn.KmeshCgroupSockWorkloadObjects.KmeshCgroupSockWorkloadMaps)
```
hzxuzhonghu, line 52: NewProcessor is too general, we can hardly get what it mean

pkg/controller/workload/workload_processor.go
```diff
@@ -883,11 +884,41 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service,
 
 	for _, workload := range workloads {
 		// TODO: Kmesh supports ServiceEntry
+		// log.Warnf("workload: %s/%s addresses: %v", workload.Namespace, workload.Name, workload.GetAddresses())
```
hzxuzhonghu: remove

pkg/controller/workload/workload_processor.go
```diff
@@ -883,11 +884,41 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service,
...
 		if workload.GetAddresses() == nil {
-			log.Warnf("workload: %s/%s addresses is nil", workload.Namespace, workload.Name)
-			continue
+			// Non-ServiceEntry workload with nil addresses will be ignored.
+			uid := workload.GetUid()
+			if !strings.Contains(uid, "ServiceEntry") {
```
hzxuzhonghu: instead of checking the Kind, can we check the Address please. Maybe one day they   can rename xEntry yEnty?

pkg/controller/workload/workload_processor.go
hzxuzhonghu: @Flying-Tom

brushax:
The improvements mentioned in the review will be fixed.

hzxuzhonghu:
@Flying-Tom any update

pkg/controller/workload/dns.go
```diff
@@ -0,0 +1,261 @@
...
+	domains := getPendingResolveDomain(workloads)
+
+	// store all pending hostnames of clusters in pendingHostnames
+	for _, workload := range workloads {
+		workloadName := workload.GetName()
+		hostname := workload.GetHostname()
+		if _, ok := r.pendingHostnames[workloadName]; !ok {
+			r.pendingHostnames[workloadName] = []string{}
+		}
+		r.pendingHostnames[workloadName] = append(r.pendingHostnames[workloadName], hostname)
```
hzxuzhonghu: not sure i understand, does a workload has multi hostnames?

hzxuzhonghu: another problem if this function is called multiple times with same workloads, this will store duplicate hostname

pkg/controller/workload/dns.go
```diff
@@ -0,0 +1,261 @@
...
+		if _, ok := r.workloadCache[hostname]; !ok {
+			// Initialize the newly added hostname
+			r.workloadCache[hostname] = &pendingResolveDomain{
+				Workloads:   make([]*workloadapi.Workload, 0),
+				RefreshRate: WorkloadDnsRefreshRate,
+			}
+		}
+		r.workloadCache[hostname].Workloads = append(
+			r.workloadCache[hostname].Workloads, workload,
```
hzxuzhonghu: same here, if processDomains called multitimes with same workloads, duplicate workload will be here

hzxuzhonghu:
Would like to see the ut

pkg/controller/workload/workload_controller.go
brushax: The name NewProcessor is originally like this (no changes have been made). When it is used externally, its function can be inferred from its package name "workload".

pkg/controller/controller.go
brushax: Done.

pkg/controller/workload/workload_processor.go
```diff
@@ -935,9 +937,29 @@ func (p *Processor) handleServicesAndWorkloads(services []*workloadapi.Service,
 		// TODO: Kmesh supports ServiceEntry
```
hzxuzhonghu: we can remove this comment now since SE is already supported

LiZhenCheng9527:
Can you display the results of the serviceEntry test in the comments?

brushax:
> Can you display the results of the serviceEntry test in the comments?

ServiceEntry yaml are as follows (with a sleep pod as client and 3 httpbin pod).
```yaml
apiVersion: networking.istio.io/v1
kind: ServiceEntry
metadata:
  name: external-svc-dns
spec:
  hosts:
    - foo.bar.com
  location: MESH_EXTERNAL
  ports:
    - number: 80
      name: http
      protocol: HTTP
  resolution: DNS
  endpoints:
    - address: httpbin1.default.svc.cluster.local
      ports:
        http: 8000
    - address: httpbin2.default.svc.cluster.local
      ports:
        http: 8000
    - address: httpbin3.default.svc.cluster.local
      ports:
        http: 8000
```

DNS resolution and `curl` results:

```bash
kubectl exec -n default sleep-5577c64d7c-fmh8f -c sleep -- curl -IsS "http://foo.bar.com/status/200"
HTTP/1.1 200 OK
server: envoy
date: Mon, 29 Sep 2025 10:09:11 GMT
content-type: text/html; charset=utf-8
access-control-allow-origin: *
access-control-allow-credentials: true
content-length: 0
x-envoy-upstream-service-time: 1
x-envoy-decorator-operation: foo.bar.com:80/*
```

```log
time="2025-09-29T10:01:07Z" level=warning msg="workload: default/external-svc-dns addresses is nil" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="wait for workload default/external-svc-dns/Kubernetes/networking.istio.io/ServiceEntry/default/external-svc-dns/httpbin1.default.svc.cluster.local address resolving" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="workload default/external-svc-dns/Kubernetes/networking.istio.io/ServiceEntry/default/external-svc-dns/httpbin1.default.svc.cluster.local addresses updated to [[10 96 242 173]]" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="workload: default/external-svc-dns addresses resolved: [[10 96 242 173]]" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=warning msg="workload: default/external-svc-dns addresses is nil" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="wait for workload default/external-svc-dns/Kubernetes/networking.istio.io/ServiceEntry/default/external-svc-dns/httpbin2.default.svc.cluster.local address resolving" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="workload default/external-svc-dns/Kubernetes/networking.istio.io/ServiceEntry/default/external-svc-dns/httpbin2.default.svc.cluster.local addresses updated to [[10 96 97 209]]" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="workload: default/external-svc-dns addresses resolved: [[10 96 97 209]]" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=warning msg="workload: default/external-svc-dns addresses is nil" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="wait for workload default/external-svc-dns/Kubernetes/networking.istio.io/ServiceEntry/default/external-svc-dns/httpbin3.default.svc.cluster.local address resolving" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="workload default/external-svc-dns/Kubernetes/networking.istio.io/ServiceEntry/default/external-svc-dns/httpbin3.default.svc.cluster.local addresses updated to [[10 96 222 245]]" subsys=workload_controller
time="2025-09-29T10:01:07Z" level=info msg="workload: default/external-svc-dns addresses resolved: [[10 96 222 245]]" subsys=workload_controller
```

hzxuzhonghu:
can you add the e2e test cases

PR #1452, opened 2025-07-24
ut_sendmsg

wxnzb:
**What type of PR is this?**

Fixes #

```release-note

```


bpf/kmesh/workload/sendmsg.c
```diff
@@ -7,7 +7,13 @@
...
 #include "bpf_log.h"
 #include "bpf_common.h"
-
+struct {
+    __uint(type, BPF_MAP_TYPE_SOCKHASH);
+    __type(key, struct bpf_sock_tuple);
+    __type(value, __u32);
+    __uint(max_entries, MAP_SIZE_OF_MANAGER);
+    __uint(map_flags, 0);
+} map_of_kmesh_socket SEC(".maps");
```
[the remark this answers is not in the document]
hzxuzhonghu: Yes, why define a map here, we already have one in bpf/kmesh/workload/sockops.c

test/bpf_ut/Makefile
```diff
@@ -47,6 +48,10 @@ tc_mark_encrypt_test.o: tc_mark_encrypt_test.c
 tc_mark_decrypt_test.o: tc_mark_decrypt_test.c
 	$(QUIET) $(CLANG) $(CLANG_FLAGS) $(TC_FLAGS) -c $< -o $@
 
+WORKLOAD_SENDMSG_FLAGS = -I$(ROOT_DIR)/bpf/kmesh/ -I$(ROOT_DIR)/bpf/kmesh/workload/include -I$(ROOT_DIR)/api/v2-c  
+workload_sendmsg_test.o: workload_sendmsg_test.c
```
hzxuzhonghu: I would suggest we moving these test building to a script and keep makefile simple

test/bpf_ut/bpftest/workload_test.go
```diff
@@ -767,3 +769,179 @@ func get_local_ipv4(t *testing.T) string {
...
+							var errno syscall.Errno
+							if errors.As(err, &errno) {
+								t.Fatalf("Attach failed: %v (errno=%d)", err, errno)
+							} else {
+								t.Fatalf("Attach failed: %v", err)
+							}
+						}
+						localIP := get_local_ipv4(t)
+						clientPort := 12345
+						serverPort := 54321
```
hzxuzhonghu, line 832: instead of using a fixed port, we can let os randomly choose which free port by calling net.Listen(:0) ?

test/bpf_ut/bpftest/workload_test.go
```diff
@@ -767,3 +769,179 @@ func get_local_ipv4(t *testing.T) string {
...
+							if !ip.Equal(expectedIP) {
+								t.Errorf("Unexpected TLV IP: got %v, want %v", ip, expectedIP)
+							}
+
+							port := binary.BigEndian.Uint16(buf[9:11])
+							if port != 53 {
+								t.Errorf("Unexpected TLV Port: got %d, want 53", port)
+							}
+
+							if buf[11] != 0xFE {
```
hzxuzhonghu: can you abstract a utility function to parse tlv protocol? So we can resue it in other tests  as well

test/bpf_ut/workload_sendmsg_test.c
```diff
@@ -0,0 +1,28 @@
...
+{
+    void *storage = NULL;
+    storage = bpf_sk_storage_get(map, sk, value, flags);
+    if (!storage && map == &map_of_sock_storage) {
+        storage = &mock_storage;
+    }
+    return storage;
+}
+
+#define bpf_sk_storage_get mock_bpf_sk_storage_get
```
hzxuzhonghu, line 69: can you elaborate a little bit how this function can be used

LiZhenCheng9527:
Run make gen beforn push your commit if you add or change bpf map.
And can you provide the result of sendmsg unit test?

test/bpf_ut/Makefile
wxnzb: > I would suggest we moving these test building to a script and keep makefile simple我建议我们将这些测试构建移动到脚本中并使 makefile 保持简单

Should all the previous include directives be combined into a single bash script?

test/bpf_ut/bpftest/workload_test.go
wxnzb: I need to verify the IP, port, and the correctness of the format specifically, so I feel it’s not really reusable. Do you think it’s necessary to abstract it?If abstracted, the message, IP, and port need to be passed as parameters. IPv4 and IPv6 should share the logic first when needed, right?

test/bpf_ut/workload_sendmsg_test.c
wxnzb: Okay, I will add some comments.

bpf/kmesh/workload/sendmsg.c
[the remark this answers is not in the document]
wxnzb: We need to redefine the map in sendmsg.c because each eBPF object file (.o) only contains its own maps. The Go test loads workload_sendmsg_test.o, so it can only access maps defined in that file. Even though the same map exists in sockops.c, it’s in a different ELF file, so the test can’t see it.

By redefining it as km_socket in sendmsg.c, the map exists in the test’s ELF file and the Go test can access it.

I think this solution works for the unit test. Do you think it’s reasonable, or do you have a better approach?

wxnzb:
> Run make gen beforn push your commit if you add or change bpf map.如果您添加或更改 bpf 映射，请在推送提交之前运行 make gen。 And can you provide the result of sendmsg unit test?您能提供 sendmsg 单元测试的结果吗？

截图 2025-08-24 18-53-01

test/bpf_ut/bpftest/workload_test.go
wxnzb: > can you abstract a utility function to parse tlv protocol? So we can resue it in other tests as well你能抽象出一个实用函数来解析 TLV 协议吗？这样我们就可以在其他测试中重复使用它了。



> can you abstract a utility function to parse tlv protocol? So we can resue it in other tests as well你能抽象出一个实用函数来解析 TLV 协议吗？这样我们就可以在其他测试中重复使用它了。

Yes, I’ve abstracted the TLV parsing logic into ParseTLVMessage and ValidateTLVMessage in a new file tlv_utils.go, so it can be reused in other tests.

bpf/kmesh/workload/sendmsg.c
```diff
@@ -7,7 +7,13 @@
...
 #include "bpf_log.h"
 #include "bpf_common.h"
-
+struct {
+    __uint(type, BPF_MAP_TYPE_SOCKHASH);
+    __type(key, struct bpf_sock_tuple);
+    __type(value, __u32);
+    __uint(max_entries, MAP_SIZE_OF_MANAGER);
+    __uint(map_flags, 0);
+} map_of_kmesh_sendmsg SEC(".maps");
```
hzxuzhonghu, line 17: add a comment before this struct to indicate this is used for test only

hzxuzhonghu (changes requested):
Wait ths map will pollute the dualengine, actually we only need this for test. Is there other way to only load the bpf for test

hzxuzhonghu:
please update

bpf/kmesh/workload/sendmsg.c
hzxuzhonghu: @wxnzb is there a way to move this out from this file,  i donot want to make it load and attached to kernel when we run kmesh.

wxnzb:
> Wait ths map will pollute the dualengine, actually we only need this for test. Is there other way to only load the bpf for test
> 等等，这个映射会污染双引擎，实际上我们只需要这个用于测试。还有其他方法可以只加载 bpf 进行测试吗？

Thanks for the feedback! You're absolutely right about the map pollution concern. 

I've addressed this by implementing conditional compilation to ensure the map is only compiled during testing:

**Changes made:**

1. **Added conditional compilation in `sendmsg.c`**:
   ```c
   #ifdef KMESH_UNIT_TEST
   struct {
       __uint(type, BPF_MAP_TYPE_SOCKHASH);
       __type(key, struct bpf_sock_tuple);
       __type(value, __u32);
       __uint(max_entries, MAP_SIZE_OF_MANAGER);
       __uint(map_flags, 0);
   } map_of_kmesh_sendmsg SEC(".maps");
   #endif
   ```

2. **Updated test Makefile** to define `KMESH_UNIT_TEST` only for test compilation:
   ```makefile
   WORKLOAD_SENDMSG_FLAGS = ... -DKMESH_UNIT_TEST
   ```

**Result:**
-  **Production builds**: The map is completely excluded from dualengine
-  **Test builds**: Include the necessary sockhash map for sk_msg program attachment
-  **Clean separation**: Zero pollution of production code

This approach ensures the BPF map is **only loaded during testing**, completely addressing the dualengine pollution concern while maintaining full test functionality.

hzxuzhonghu (approved):
Thank you

PR #1456, opened 2025-07-28
add ip namespace authz E2E test

Geaus:
**What type of PR is this?**

/kind security

The authentication policy lacks end-to-end testing


test/e2e/baseline_test.go
```diff
@@ -1307,3 +1307,258 @@ func getSupportedIPFamilies(t framework.TestContext) (v4 bool, v6 bool) {
 	}
 	return
 }
+
+func TestIPAuthorization(t *testing.T) {
```
YaoZengzeng: What's the difference between this case with the existing `TestAuthorizationL4`?

PR #1507, opened 2025-09-22
change Kmesh deploy yaml

LiZhenCheng9527:
**What type of PR is this?**

After Kmesh support dns proxy, the YAML file for Kmesh deployment lacks the corresponding environment variable configurations.
Fixes #

```release-note

```

### DIFF
--- a/deploy/yaml/kmesh.yaml
+++ b/deploy/yaml/kmesh.yaml
@@ -107,6 +107,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: KMESH_ENABLE_DNS_PROXY
+              value: "true"
           volumeMounts:
             - name: mnt
               mountPath: /mnt


deploy/yaml/kmesh.yaml
```diff
@@ -107,6 +107,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: KMESH_ENABLE_DNS_PROXY
+              value: "true"
```
hzxuzhonghu, line 111: It is intended to be false at beginning

LiZhenCheng9527: Then should we also change `feature.dnsProxy.enable` to false in the Helm values?

hzxuzhonghu: Ah, i am mistaken. It is now enabled by default in the helm

PR #1561, opened 2025-12-31
test: fix privilege checks and make persistPath testable 

yuvraj-kolkar17:
**What type of PR is this?**
This PR fixes test failures related to privilege requirements and makes the test suite more flexible. It also adds additional tests.

/kind bug
/kind enhancement

1. **Privilege-related test failures**: `TestHandleKmeshManage` required root privileges for namespace operations but failed without proper handling, blocking non-privileged test execution
2. **Path immutability**: `persistPath` was defined as a constant, preventing tests from using temporary directories and causing permission errors
3. **Insufficient test coverage**: Missing test cases for edge cases and error scenarios

# Changes
- **pkg/utils/hash_name.go**: Changed `persistPath` from `const` to `var` to allow testing with temporary directories
- **pkg/utils/hash_name_test.go**: Added test setup helpers (`setupTestPersistPath`, `restorePersistPath`) for proper test isolation
- **pkg/utils/enroll_test.go**: Added privilege check to `TestHandleKmeshManage` to skip when not running as root

```release-note
NONE
```

### DIFF
--- a/pkg/utils/hash_name.go
+++ b/pkg/utils/hash_name.go
@@ -26,7 +26,7 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-const (
+var (
 	persistPath = "/mnt/hash_name.yaml"
 )
 
--- a/pkg/utils/enroll_test.go
+++ b/pkg/utils/enroll_test.go
@@ -257,6 +257,11 @@ func TestShouldEnroll(t *testing.T) {
 }
 
 func TestHandleKmeshManage(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/utils/hash_name.go
```diff
@@ -26,7 +26,7 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-const (
+var (
```
LiZhenCheng9527, line 29: Revert!

PR #1585, opened 2026-02-09
Add a robust E2E testing suite for Kmesh Locality Load Balancing.

### LINKED ISSUE #1146: Add e2e for locality aware lb

sarafarajnasardi:
## Review

**What type of PR is this?**  
/kind enhancement

---

## What this PR does / why we need it

This PR introduces a robust **End-to-End (E2E) testing suite for Kmesh Locality Load Balancing**.

As Kmesh moves toward feature parity with Istio’s advanced traffic management, it is critical to ensure that traffic correctly respects **administrative locality boundaries** (region, zone, and subzone) to achieve optimal performance and high availability.

---

### 1. Dynamic Multi-Zone Topology Simulation
- Automatically detects available nodes in the cluster.
- Simulates a multi-zone environment by dynamically applying:
  - `topology.kubernetes.io/region`
  - `topology.kubernetes.io/zone`
  - `topology.kubernetes.io/subzone`
- Enables validation of locality-aware routing even on **single-cluster environments (e.g., Kind)**.

### 2. Advanced Pod Scheduling & Taint Handling
- Introduces a `patchDeployment` utility using `kubectl patch` to dynamically apply:
  - `nodeSelector`
  - `tolerations`
- Ensures client and server workloads are pinned to specific nodes to accurately verify **intra-zone and cross-zone traffic behavior**.
- Explicitly handles control-plane node taints, allowing the test to run on standard Kind clusters without manual intervention.

### 3. Verification Scenarios
- **Locality Priority**  
  Verifies that a client in `region1/zone1` sends traffic exclusively to the server subset in the same zone when `localityLbSetting` is enabled in the `DestinationRule`.

- **Failover Logic**  
  Simulates a zone-level outage by scaling the local server subset to zero and verifies that Kmesh seamlessly fails over to a remote subset in `zone2`.

### 4. Environmental Stability & Reliability
- Uses `t.Cleanup` to restore node labels to their original state after test execution.
- Employs optimized timeout and retry logic via the Istio integration framework’s `retry` package to handle asynchronous resource readiness.

---

## Which issue(s) this PR fixes

Fixes #1146

---

## Special notes for your reviewer
- The test suite leverages `istio.io/istio/pkg/test/framework` and `echo` components to maintain consistency with existing Kmesh integration tests.
- Service definitions were aligned to **port 5000** (standard echo instance port), while retaining **port 80** as the external entry point to ensure compatibility with Istio’s internal routing logic.

---

## Does this PR introduce a user-facing change?
NONE

---

## release-note

```release-note
Add refined End-to-End tests for Locality Load Balancing and Failover, including support for region/zone/subzone topology simulation.

### DIFF
--- a/test/e2e/locality_lb_test.go
+++ b/test/e2e/locality_lb_test.go
@@ -1,0 +1,347 @@
+//go:build integ
+// +build integ
+
+/*
+ * Copyright The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kmesh
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	echot "istio.io/istio/pkg/test/echo"
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/echo"
+	"istio.io/istio/pkg/test/framework/components/echo/check"
+	"istio.io/istio/pkg/test/framework/components/echo/common/ports"
+	"istio.io/istio/pkg/test/framework/components/echo/deployment"
+	"istio.io/istio/pkg/test/framework/components/echo/match"
+	"istio.io/istio/pkg/test/framework/components/namespace"
+	"istio.io/istio/pkg/test/shell"
+	"istio.io/istio/pkg/test/util/retry"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestLocalityLB(t *testing.T) {
+	framework.NewTest(t).Run(func(t framework.TestContext) {
+		// 1. Create a namespace
+		ns, err := namespace.New(t, namespace.Config{
+			Prefix: "locality",
+			Inject: false,
+		})
+		if err != nil {
+			t.Fatalf("failed to create namespace: %v", err)
+		}
+
+		enrollNamespaceOrFail(t, ns.Name())
+
+		// 2. Setup Topology (Label Nodes)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/e2e/locality_lb_test.go [hunk 1]
```diff
@@ -0,0 +1,262 @@
...
+  trafficPolicy:
+    loadBalancer:
+      localityLbSetting:
+        enabled: true
+      simple: ROUND_ROBIN
+`).ApplyOrFail(t)
+
+		// 5. Verification
+		t.NewSubTest("locality priority").Run(func(t framework.TestContext) {
+			// Traffic from client (zone1) should hit server-v1 (zone1)
```
LiZhenCheng9527, line 155: I have a question: Your load balance policy is ROUND_ROBIN. How to ensure that the client always hits server-v1?

sarafarajnasardi: Thanks for the review!
My understanding from diving into the locality implementation is that ``localityLbSetting`` acts as a prioritization layer before the load balancing policy is applied.
In this test setup:
Zone Priority: Kmesh groups endpoints by locality. Since the client and ``server-v1`` share the same ``zone (zone1)``, ``server-v1`` is placed in the highest priority pool (Priority 0).
Round Robin: The ``ROUND_ROBIN`` policy then selects from that Priority 0 pool. Since ``server-v1`` is the only healthy endpoint there, it receives 100% of the traffic.
Ideally, ``ROUND_ROBIN`` would only distribute traffic to ``server-v2 (in zone2)`` if there were multiple priority levels active or if the local pool was empty ...which matches what I observed in the ``failover`` subtest where traffic shifted only after scaling ``v1`` to zero.
Does this match the expected behavior for ``localityLbSetting`` in Kmesh?

hzxuzhonghu: yeah, that's right.

test/e2e/locality_lb_test.go
```diff
@@ -0,0 +1,262 @@
...
+				Subsets: []echo.SubsetConfig{
+					{
+						Replicas: 1,
+						Version:  "v1", // Local to client
+						Labels:   map[string]string{"app": "server", "version": "v1"},
+					},
+					{
+						Replicas: 1,
+						Version:  "v2", // Remote to client
+						Labels:   map[string]string{"app": "server", "version": "v2"},
```
hzxuzhonghu, line 119: I think you donot need to create subset here

hzxuzhonghu: IC, you created two deployment

hzxuzhonghu (approved):

Issue #1592, opened 2026-02-28
pkg/controller/bypass: Missing iptables cleanup during Kmesh daemon restart


shivansh-gohem:
<**What happened**:
While auditing the bypass controller logic in the `main` branch, I noticed a missing cleanup mechanism for iptables rules when the Kmesh daemon is unexpectedly restarted or terminated. 

There is a developer `TODO` explicitly acknowledging this technical debt in `pkg/controller/bypass/bypass_controller.go`:
`// TODO: add delete iptables in case we missed skip bypass during kmesh restart`

Because of this missing logic, if the Kmesh daemonset pod restarts, the `bypass_controller` fails to clean up the iptables rules it previously injected into the pod network namespaces. 

**What i expected **:
During the `bypass_controller` initialization phase, Kmesh should ideally perform a reconciliation/cleanup step. When the new Kmesh pod boots up, it should scan for existing Kmesh-specific bypass rules and flush them before establishing its new baseline to prevent stale or duplicated rules from accumulating on the node.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy a pod with the bypass label enabled so Kmesh injects the bypass iptables rules into the pod's network namespace.
2. Verify the rules are present inside the pod's network namespace using `nsenter` or a debug container.
3. Force a restart of the Kmesh control plane (`kubectl delete pods -n kmesh-system -l app=kmesh`).
4. Once the new Kmesh pod is running, check the application pod's network namespace again. 
5. The old iptables rules will still be present (and potentially duplicated by the new Kmesh instance) instead of being cleanly flushed.

I would love to take ownership of this issue and submit a PR to implement this cleanup logic, provided the maintainers agree with this approach! 

**Environment**:
- Kmesh version: `main` branch
- Kmesh mode(kmesh has `Kernel-Native Mode` and `Dual-Engine Mode`): `Dual-Engine Mode`
- Istio version: `1.26.0` (ambient)
- Kernel version: `6.8.0-57-generic` (Ubuntu)
- Others: Local `kind` cluster

shivansh-gohem:
hey @LiZhenCheng9527 @hzxuzhonghu could you please review this issue along with its pr !!!

PR #1593, opened 2026-03-01
fix(bypass): clean up stale iptables rules on kmesh restart

shivansh-gohem:
**What type of PR is this?**
/kind bug

This PR resolves a silent security/resource leak in the `bypass_controller` where stale `iptables` rules were left orphaned in pod network namespaces if Kmesh crashed or was restarted. 

If a pod had the `kmesh.net/bypass=enabled` label removed while the Kmesh daemon was down, the previous controller logic would ignore the pod upon restart (`!shouldBypass`), leaving the `PREROUTING/OUTPUT RETURN` rules permanently in the pod's netns. This resulted in a silent mTLS bypass.

fixes(#1592 )

**Changes:**
1. Replaced the `TODO` in `AddFunc`. When the pod informer syncs on startup, it now calls `deleteIptables` for any sidecar-injected pod that lacks the bypass label. Because `deleteIptables` is idempotent, this safely cleans up stale state without impacting non-bypassed pods.
2. Added an early return in `Run()` if `WaitForCacheSync` fails.
3. Added `TestBypassControllerKmeshRestart` using `gomonkey` to mock `ns.GetPodNSpath` and verify the reconciliation cleanup flows correctly.

**E2E Verification Performed:**
```bash
# Deployed pod, applied bypass label, then removed label and immediately killed Kmesh.
# Upon Kmesh restart, verified the target pod's netns to ensure rules were reconciled:
$ docker exec kmesh-bug-hunt-control-plane nsenter -t $PID -n iptables -t nat -L PREROUTING -n --line-numbers

Chain PREROUTING (policy ACCEPT)
num  target     prot opt source               destination
1    ISTIO_INBOUND  6    --  0.0.0.0/0            0.0.0.0/0
# SUCCESS: Kmesh RETURN rule cleanly flushed during AddFunc reconciliation.

### DIFF
--- a/pkg/controller/bypass/bypass_controller.go
+++ b/pkg/controller/bypass/bypass_controller.go
@@ -65,7 +65,21 @@ func NewByPassController(client kubernetes.Interface) *Controller {
 			}
 
 			if !shouldBypass(pod) {
-				// TODO: add delete iptables in case we missed skip bypass during kmesh restart
+				// On Kmesh restart the informer re-lists all existing pods and fires
+				// AddFunc for each of them. If a pod previously had the bypass label
+				// (and thus has PREROUTING/OUTPUT RETURN rules) but the label has since
+				// been removed, we must clean up those stale rules here.
+				nspath, err := ns.GetPodNSpath(pod)
+				if err != nil {
+					// Pod may still be initialising; this is not an error.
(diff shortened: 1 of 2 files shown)


shivansh-gohem:
@YaoZengzeng @nlgwcy please have a look at this !!

shivansh-gohem:
@LiZhenCheng9527 @hzxuzhonghu @YaoZengzeng could you please take a look at this !!!!!!

PR #1601, opened 2026-03-02
ctl/monitoring: fix getKmeshDaemonPod rejecting pod names with double hyphens

### LINKED ISSUE #1600: ctl/monitoring: getKmeshDaemonPod incorrectly rejects pod names containing double hyphens
## What happened

`getKmeshDaemonPod` in `ctl/monitoring/monitoring.go` uses `strings.Contains(arg, "--")` to detect flag-like arguments. This incorrectly rejects legitimate Kubernetes pod names that contain consecutive hyphens (e.g. `some--thing`), causing the command to operate on **all** daemons instead of the specified pod.

## Expected behavior

Only arguments starting with `--` should be treated as flags. Pod names containing `--` in the middle should be accepted.

## Suggested fix

Change `strings.Contains(arg, "--")` to `strings.HasPrefix(arg, "--")`:

```diff
-if strings.Contains(arg, "--") {
+if strings.HasPrefix(arg, "--") {
```

## How to reproduce

```
kmeshctl monitoring some--pod --accesslog enable
```

This will ignore `some--pod` and operate on all daemons instead.

Found during unit test review in #1599.

Eeshu-Yadav:
## What type of PR is this?
/kind bug

## What this PR does / why we need it:
Fixes `getKmeshDaemonPod` using `strings.Contains(arg, "--")` which incorrectly rejects legitimate Kubernetes pod names containing consecutive hyphens (e.g. `some--thing`). Changed to `strings.HasPrefix(arg, "--")` so only flag-style arguments starting with `--` are skipped.

```diff
-if strings.Contains(args[0], "--") {
+if strings.HasPrefix(args[0], "--") {
```

Includes a unit test verifying the corrected behavior.

## Which issue(s) this PR fixes:
Fixes #1600

### DIFF
--- a/ctl/monitoring/monitoring.go
+++ b/ctl/monitoring/monitoring.go
@@ -147,7 +147,7 @@ func getKmeshDaemonPod(args []string) (string, bool) {
 	if len(args) == 0 {
 		return "", false
 	}
-	if strings.Contains(args[0], "--") {
+	if strings.HasPrefix(args[0], "--") {
 		return "", false
 	}
 	return args[0], true
--- a/ctl/monitoring/monitoring_test.go
+++ b/ctl/monitoring/monitoring_test.go
@@ -1,0 +1,55 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


LiZhenCheng9527:
Please update your base branch.

PR #1625, opened 2026-03-22
bypass: do not insert iptables when sidecar not injected

### LINKED ISSUE #474: Donot insert iptables when sidecar not injected

```
			enableSidecar, _ := checkSidecar(client, pod)
			enableKmesh := isKmeshManaged(pod)
			if !enableSidecar && !enableKmesh {
				log.Info("do not need process, pod is not managed by sidecar or kmesh")
				return
			}
```

In bypass controller, we check whether sidecar or kmesh enabled for the pod, this is right. But the way it checks sidecar, it ir nor right

If a namespace with injection label, we donot check whether pod has a sidecar anymore. And similarly, we also check pod annotation, without checking sidecar.

This can be mistaken, when the pod is started earlier.

**Why is this needed**:

imshubham22apr-gif:
What this PR does / why we need it:
This PR improves the sidecar detection logic in the bypass controller. Previously, it only checked the pod annotation for sidecar status, which could be inaccurate if a pod was started before sidecar injection was enabled in its namespace. We've enhanced the detection to also check for the actual presence of the istio-proxy container and properly account for Kmesh management states.

Which issue(s) this PR fixes:
Fixes #474

Special notes for your reviewer:
NONE

Does this PR introduce a user-facing change?:

release-note
NONE

### DIFF
--- a/pkg/controller/bypass/bypass_controller.go
+++ b/pkg/controller/bypass/bypass_controller.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 
+	"kmesh.net/kmesh/pkg/constants"
 	ns "kmesh.net/kmesh/pkg/controller/netns"
 	"kmesh.net/kmesh/pkg/kube"
 	"kmesh.net/kmesh/pkg/logger"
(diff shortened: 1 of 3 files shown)


yashisrani:
@imshubham22apr-gif hey shubham, i saw your pr, please add DCO sign when push code ... i saw same issue in your all prs.

imshubham22apr-gif:
@yashisrani okI'll correct it

PR #1632, opened 2026-03-28
fix(ebpf): prevent file descriptor leak in resource retrieval

### LINKED ISSUE #1618: [BUG] File Descriptor Leak in BPF Resource Retrieval (GetProgramByName / GetMapByName)
## Description

The functions `GetProgramByName` and `GetMapByName` in `pkg/utils/ebpf.go` iterate over all system BPF objects to find one by name. During each iteration, they open a new file descriptor via `ebpf.NewProgramFromID` / `ebpf.NewMapFromID`. If the object's name does not match the target, the function continues to the next iteration without closing the previously opened file descriptor.

This results in a file descriptor leak on every non-matching iteration. In production environments with many BPF programs and maps loaded (e.g., a busy node with many pods), this can quickly exhaust the process's open file descriptor limit (ulimit -n), causing subsequent kernel calls to fail with EMFILE: too many open files.

## Affected Code

**File:** `pkg/utils/ebpf.go`

### `GetProgramByName` (lines 25-55):
```go
for {
    if progID, err = ebpf.ProgramGetNextID(progID); err != nil {
        return nil, err
    }

    if targetProg, err = ebpf.NewProgramFromID(progID); err != nil { // opens an FD
        return nil, err
    }

    if targetProgInfo, err = targetProg.Info(); err != nil {
        return nil, err
    }

    if strings.Compare(targetProgInfo.Name, name) == 0 {
        return targetProg, nil // only this path closes the FD (caller's responsibility)
    }
    // BUG: loop continues here without calling targetProg.Close()
}
```

The same pattern exists in `GetMapByName` (lines 57-87).

## Root Cause

When `targetProgInfo.Name != name`, neither `targetProg.Close()` nor `targetMap.Close()` is called before the next iteration, leaking the kernel file descriptor allocated by `NewProgramFromID` / `NewMapFromID`.

## Expected Fix

Call `.Close()` on the object before continuing to the next iteration when the name does not match:

```go
for {
    if progID, err = ebpf.ProgramGetNextID(progID); err != nil {
        return nil, err
    }

    targetProg, err = ebpf.NewProgramFromID(progID)
    if err != nil {
        return nil, err
    }

    targetProgInfo, err = targetProg.Info()
    if err != nil {
        targetProg.Close()
        return nil, err
    }

    if strings.Compare(targetProgInfo.Name, name) == 0 {
        return targetProg, nil
    }

    targetProg.Close() // Fix: close FD before next iteration
}
```

Apply the same fix to `GetMapByName`.

## Impact

- **Severity:** High
- - **Category:** Resource Leak
- - - **Affected Environments:** Any Kmesh deployment with a large number of BPF programs/maps (busy multi-pod nodes)
- - - - **Failure Mode:** EMFILE: too many open files on kernel BPF syscalls, causing daemon instability
## Steps to Reproduce

1. Deploy Kmesh on a node with many BPF programs already loaded
2. 2. Repeatedly call any code path that invokes `GetProgramByName` or `GetMapByName`
3. 3. Monitor open file descriptors via ls -la /proc/<pid>/fd | wc -l
4. 4. Observe FD count growing unboundedly on non-matching iterations

shivansh-gohem:
Fixes #1618.

**What this PR does :**
The `GetProgramByName` and `GetMapByName` functions in `pkg/utils/ebpf.go` were iterating over system BPF objects but failing to close the file descriptors generated by `ebpf.NewProgramFromID` and `ebpf.NewMapFromID` on non-matching iterations. In environments with many loaded BPF objects, this loop quickly exhausts the open file descriptor limit (`EMFILE`).

This PR fixes the issue by:
1. Adding explicit `Close()` calls on error paths and before continuing to the next iteration.
2. Avoiding `defer` inside the `for` loop to ensure FDs are released immediately per-iteration.
3. Adding a regression test (`TestGetProgramByName_FDLeak`) that aggressively searches for a non-existent program to verify FDs remain stable during full system iterations.

**Testing:**
Locally verified using the new regression test. Before the fix, the loop leaked over 120 FDs. After the fix, FDs remain perfectly stable:

```text
=== RUN   TestGetProgramByName_FDLeak
    ebpf_test.go:21: Initial open File Descriptors: 7
    ebpf_test.go:31: Final open File Descriptors: 7
    ebpf_test.go:37: No leak detected.
--- PASS: TestGetProgramByName_FDLeak (0.13s)

### DIFF
--- a/pkg/utils/ebpf.go
+++ b/pkg/utils/ebpf.go
@@ -34,23 +34,23 @@ func GetProgramByName(name string) (*ebpf.Program, error) {
 
 	for {
 		if progID, err = ebpf.ProgramGetNextID(progID); err != nil {
-			err = fmt.Errorf("failed to get system next program id, err is %v\n", err)
-			return nil, err
+			return nil, fmt.Errorf("failed to get system next program id: %w", err)
 		}
 
 		if targetProg, err = ebpf.NewProgramFromID(progID); err != nil {
-			err = fmt.Errorf("failed to get new program from id:%v, err is %v\n", progID, err)
-			return nil, err
+			return nil, fmt.Errorf("failed to get new program from id:%v: %w", progID, err)
 		}
 
 		if targetProgInfo, err = targetProg.Info(); err != nil {
-			err = fmt.Errorf("failed to get new program info from fd:%v, err is %v\n", targetProg, err)
-			return nil, err
+			targetProg.Close()
+			return nil, fmt.Errorf("failed to get program info for id:%v: %w", progID, err)
 		}
 
 		if strings.Compare(targetProgInfo.Name, name) == 0 {
 			return targetProg, nil
 		}
+
+		targetProg.Close()
 	}
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/utils/ebpf.go [hunk 1]
```diff
@@ -34,23 +36,26 @@ func GetProgramByName(name string) (*ebpf.Program, error) {
...
 		if targetProg, err = ebpf.NewProgramFromID(progID); err != nil {
-			err = fmt.Errorf("failed to get new program from id:%v, err is %v\n", progID, err)
-			return nil, err
+			return nil, fmt.Errorf("failed to get new program from id:%v: %w", progID, err)
 		}
 
 		if targetProgInfo, err = targetProg.Info(); err != nil {
-			err = fmt.Errorf("failed to get new program info from fd:%v, err is %v\n", targetProg, err)
-			return nil, err
+			_ = targetProg.Close()
```
LiZhenCheng9527, line 50: Is there no error when the shutdown is performed?

shivansh-gohem:
@LiZhenCheng9527 Good catch, thank you for the review! Yes, Close() does return an error. The _ = assignment is intentional — it explicitly acknowledges the error to satisfy the errcheck linter (which is enabled in .golangci.yaml) while doing best-effort cleanup.

On this error path, Info() has already failed, so we are returning that primary error. The underlying Close() is just syscall.Close(fd), which can only realistically fail with EBADF — not something we can recover from. This pattern is consistent with the rest of the codebase (e.g., client.go, rbac.go, ads_controller.go)!!!!

PR #1651, opened 2026-04-15
net: support port name mapping in ServiceEntry endpoints (#1088)

### LINKED ISSUE #1088: Kmesh should support port name mapping in ServiceEntry

```yaml 
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: test-se-v4
spec:
  hosts:
  - dummy-v4.example.com
  addresses:
  - 240.240.240.255
  ports:
  - number: 80
    name: http
    protocol: HTTP
  endpoints:
  - address: 172.18.0.3
    ports:
      http: 8080
```

According to the spec, in ServiceEntry, if the key of the ports in the inline WorkloadEntry matches the name of the port in ServiceEntry, the port is used as the port to access the backend, regardless of whether the target port is specified.

However, it doesn't work in Kmesh.

**Why is this needed**:

Official spec requirements

shubhtrek:
Fixes #1088

In ServiceEntry, if a WorkloadEntry defines a port mapping (e.g., mapping port name 'http' to 8080), Kmesh was previously ignoring this and defaulting to the ServicePort.

Changes:
- Updated handleService in workload_processor.go to iterate through service endpoints.
- If an endpoint provides a mapping for the specific port name, that TargetPort is prioritized.
- Maintains existing fallback logic for waypoints and unmapped ports.

Verified by mocking the resolution logic in a unit test.


FAUST-BENCHOU:
make sure ci happy first

PR #1678, opened 2026-05-06
optimize(workload): pass IPs directly to deletePodFrontendData

itvi-1234:
**What type of PR is this?**

/kind cleanup

Optimizing the workload cleanup process to make it faster and more reliable.

**The Improvement:**
Previously, when deleting a workload, the code performed an extra step to look up the IP address from a table. I have removed this extra step because we already have the IP address available in the workload data.

**Key Benefits:**
1. **Performance:** Skipping the extra lookup makes the deletion process faster.
2. **Better Cleanup:** The old logic sometimes missed deleting all IP addresses if a workload had more than one. My change ensures that every IP address for a workload is properly cleaned up.

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -17,6 +17,7 @@
 package workload
 
 import (
+	"errors"
 	"fmt"
 	"net/netip"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/controller/workload/workload_processor.go
```diff
@@ -199,8 +199,17 @@ func (p *Processor) processWorkloadResponse(rsp *service_discovery_v3.DeltaDisco
...
 }
 
-// TODO: optimize me by passing workload ip directly
-func (p *Processor) deletePodFrontendData(uid uint32) error {
+func (p *Processor) deletePodFrontendData(uid uint32, ips [][]byte) error {
+	if len(ips) > 0 {
+		for _, ip := range ips {
+			fk := bpf.FrontendKey{}
+			nets.CopyIpByteFromSlice(&fk.Ip, ip)
+			if err := p.bpf.FrontendDelete(&fk); err != nil {
```
LiZhenCheng9527: Have you previously used an IP as a key for storage?

itvi-1234: Yes! The FrontendKey in the KmFrontend BPF map uses the IP address ([16]byte) as the key .

PR #1680, opened 2026-05-06
fix(workload): prevent endpoint/service BPF map divergence on partial…

### LINKED ISSUE #1679: Potential BPF map consistency issue during partial endpoint update failures
### Description

While reviewing the workload/service endpoint update flow, I noticed a potential consistency issue in addWorkloadToService() where the in-memory service state can diverge from the actual BPF maps during partial failures.

The current flow mutates sv.EndpointCount before all BPF updates succeed:

sv.EndpointCount[priority]++

if err := p.bpf.EndpointUpdate(&ek, &ev); err != nil {
    return ek, err
}

if err := p.bpf.ServiceUpdate(sk, sv); err != nil {
    return ek, err
}

This creates multiple failure windows where:

EndpointCount is incremented in memory but the endpoint entry was never successfully inserted
endpoint entries exist in the BPF map but the corresponding service count update failed
userspace cache state and datapath-visible state can temporarily diverge
### Potential Impact

This may lead to inconsistent endpoint selection behavior in the datapath under partial BPF update failures such as:

map update failures
kernel memory pressure
transient map operation errors

### Possible symptoms:

newly added endpoints not receiving traffic
incorrect endpoint counts
stale/ghost endpoint entries
uneven traffic distribution

Because endpoint selection depends on EndpointCount, a failed ServiceUpdate() after a successful EndpointUpdate() may leave a valid endpoint unreachable by the datapath.

### Relevant Code Paths

Files:

pkg/controller/workload/workload_processor.go

Functions:

addWorkloadToService()
deleteEndpoint()
Suggested Direction

One possible approach could be making the update sequence more transactional:

Compute endpoint index without mutating service state
Perform endpoint map update first
Update service count only after endpoint insertion succeeds
Roll back endpoint insertion if service update fails

Additionally, the delete flow may benefit from clearer ordering guarantees to avoid transient inconsistent states during endpoint swaps/deletes.

### Reproduction Idea

This may be reproducible by:

injecting failures into EndpointUpdate() / ServiceUpdate()
simulating BPF map pressure
forcing partial failures during rapid pod churn
### Why This Matters

This affects consistency between:

userspace workload/service state
endpoint BPF maps
datapath-visible service topology

Since Kmesh relies heavily on BPF map correctness for traffic routing, preserving update invariants during failure scenarios seems important for long-term reliability.

OmAmbole009:
### Summary

This PR fixes a transactional consistency issue during endpoint registration/removal where partial failures could leave EndpointMap, ServiceMap, and EndpointCache in divergent states.

The fix ensures:

endpoint visibility is only exposed to the datapath after successful service count updates
failed updates correctly roll back intermediate state
cache state remains consistent with BPF-visible state
endpoint deletion ordering avoids transient duplicate backend visibility

Fixes #1679

### Root Cause
1. addWorkloadToService() mutated state before all writes succeeded

Previously:

sv.EndpointCount was incremented before EndpointUpdate
EndpointCache was updated before ServiceUpdate

This created two failure windows:

Failure window A

If EndpointUpdate() failed:

sv.EndpointCount was already incremented
in-memory state diverged from the BPF service map
Failure window B

If ServiceUpdate() failed:

endpoint entry already existed in the BPF endpoint map
service count was not incremented
endpoint became unreachable to the datapath
cache state incorrectly reflected a successful add

This could lead to silent traffic inconsistencies.

2. deleteEndpoint() exposed transient duplicate backend visibility

The original ordering was:

Swap -> decrement count -> delete last entry

Between swap and count decrement, the swapped backend could temporarily become visible at two indices simultaneously.

### Fix Strategy
addWorkloadToService()

New ordering:

EndpointUpdate -> increment count + ServiceUpdate -> cache update

Behavior:

backend index computed without mutating state
endpoint inserted first while still invisible to datapath
service count updated only after successful endpoint insertion
rollback performed if ServiceUpdate() fails
cache updated only after all BPF updates succeed
deleteEndpoint()

New ordering:

decrement count -> ServiceUpdate -> swap -> delete last entry

Behavior:

last slot becomes unreachable before swap
prevents transient duplicate backend selection
rollback attempted if swap fails
deleting the unreachable last entry becomes safe/non-critical
### Tests Added
addWorkloadToService
TestAddWorkloadToService_EndpointUpdateFailure
TestAddWorkloadToService_ServiceUpdateFailure
TestAddWorkloadToService_EndpointCountConsistency
TestAddWorkloadToService_RetryAfterFailureSucceeds
deleteEndpoint
TestDeleteEndpoint_CountDecrementedBeforeSwap
TestDeleteEndpoint_DeleteLastEntry
TestDeleteEndpoint_EndpointCountServiceMapConsistency

These tests validate:

rollback correctness
endpoint/service consistency
cache consistency
retry behavior
endpoint count invariants
map compactness after deletion
### Result

This change improves consistency guarantees between:

userspace workload state
EndpointMap
ServiceMap
datapath-visible topology

and prevents silent routing inconsistencies during partial BPF update failures.

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -399,27 +399,50 @@ func (p *Processor) removeServiceResourceFromBpfMap(svc *workloadapi.Service, na
 	return nil
 }
 
-// addWorkloadToService update service & endpoint bpf map when a workload has new bound services
+// addWorkloadToService updates the endpoint and service BPF maps when a workload is bound to a service.
+//
+// Ordering guarantee: the endpoint entry is written before the service EndpointCount is
+// incremented, so the eBPF datapath never observes a count that points to a non-existent
+// entry.  If the ServiceUpdate fails after the endpoint has already been written we roll
+// back the endpoint insertion and the in-memory EndpointCache so that all three views
+// (BPF endpoint map, BPF service map, EndpointCache) remain consistent.
 func (p *Processor) addWorkloadToService(sk *bpf.ServiceKey, sv *bpf.ServiceValue, workloadUid uint32, priority uint32) (bpf.EndpointKey, error) {
-	var (
-		ek = bpf.EndpointKey{}
-		ev = bpf.EndpointValue{}
-	)
+	// Compute the new backend index without mutating sv yet.  sv.EndpointCount is only
+	// incremented after both BPF map writes succeed so that a partial failure cannot
+	// leave the in-memory service value out of sync with the BPF service map.
+	newIndex := sv.EndpointCount[priority] + 1
+	ek := bpf.EndpointKey{
+		BackendIndex: newIndex,
+		ServiceId:    sk.ServiceId,
+		Prio:         priority,
+	}
+	ev := bpf.EndpointValue{BackendUid: workloadUid}
 
-	sv.EndpointCount[priority]++
-	ek.BackendIndex = sv.EndpointCount[priority]
-	ek.ServiceId = sk.ServiceId
-	ek.Prio = priority
-	ev.BackendUid = workloadUid
+	// Step 1: write the endpoint entry.  The service EndpointCount still points to the
+	// previous last entry, so the new entry is invisible to the eBPF program until step 2.
 	if err := p.bpf.EndpointUpdate(&ek, &ev); err != nil {
 		log.Errorf("Update endpoint map failed, err:%s", err)
 		return ek, err
 	}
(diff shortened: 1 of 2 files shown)


OmAmbole009:
@hzxuzhonghu @kwb0523

I’ve addressed the latest review feedback and pushed the follow-up fixes/tests. The PR should now be up to date with all discussed changes.

Could you please take a look when you get a chance? Thanks!

Issue #1693, opened 2026-05-10
Design discussion: minimum Kmesh API surface for an external Headlamp plugin (LFX 2026-T2 prep)


WasThatRudy:
Following up on #1658. As prep for the Integrating-Kmesh-into-Headlamp-UI mentorship, I mapped what a Headlamp plugin would actually need from the Kmesh side. Filing this so applicants and maintainers can align on the minimum API surface before code starts.

## What's there today

1. **One Kmesh-owned CRD:** `KmeshNodeInfo` (`kmesh.net/v1alpha1`, namespaced; `deploy/yaml/crd/kmesh.net_kmeshnodeinfos.yaml`). Spec covers IPsec wiring (`addresses`, `bootID`, `podCIDRs`, `spi`). `status` schema is currently empty (`type: object`, no properties).
2. **Waypoints are Istio Gateways, not a Kmesh CRD.** `ctl/waypoint/waypoint.go:130` filters `gateway.networking.k8s.io/v1.Gateway` by `gatewayClassName == constants.WaypointGatewayClassName` (upstream Istio constant, `"istio-waypoint"`). Enrollment is driven by labels `istio.io/dataplane-mode: Kmesh`, `istio.io/use-waypoint`, `istio.io/waypoint-for`.
3. **eBPF map state is HTTP, not a CRD.** The kmesh-daemon `status_server` (`pkg/status/status_server.go:47-103`) exposes on `:15200`:
   - `/debug/config_dump/bpf/kernel-native`, `/debug/config_dump/bpf/dual-engine` — eBPF map dumps
   - `/debug/config_dump/{kernel-native,dual-engine}` — full xDS-style dump
   - `/debug/ready`, `/version`, `/monitoring`, `/workload_metrics`, `/connection_metrics`, `/authz`

A Headlamp plugin reaches these via the Kubernetes API server's pod proxy (no port-forward needed in-browser), but only if those endpoints stay stable.

## Gaps that affect a clean Headlamp UX, in order of impact

1. **`KmeshNodeInfo.status` has no schema.** The mentorship deliverable "per-resource health/readiness indicators" needs `status.conditions` (or similar). Recommend populating `status` with a standard `conditions []metav1.Condition` and surfacing IPsec key/SPI rollover state. Without this, the plugin has nothing structured to render for health.
[...]

Issue #1658, opened 2026-04-28
[lfx-mentorship-2026-June-August] Integrating Kmesh into Headlamp UI


LiZhenCheng9527:
- Description: [Headlamp](https://headlamp.dev) is an open-source, extensible Kubernetes web UI offering easy cluster management, multi-cluster support, RBAC, and a plugin system for adding custom functionality. Users who work with Kmesh today have to switch back and forth between Headlamp (for general Kubernetes resource management) and CLI tools / `kubectl` (for Kmesh-specific inspection), which creates a fragmented workflow and a poor user experience. There is currently no simple visual way to view Kmesh resources, inspect waypoints and related components, understand overall mesh status, or troubleshoot issues quickly from within an interface users already use. This project proposes building a Headlamp plugin for Kmesh that brings Kmesh resources directly into the Headlamp UI, providing lightweight visibility of Kmesh resources alongside other Kubernetes resources. The full-featured Kmesh dashboard remains the place for advanced operations; the Headlamp plugin focuses on reducing context switching and improving ease of use for day-to-day workflows.
- Expected Outcome:
  - A Headlamp plugin (TypeScript/React) that registers Kmesh CRDs and surfaces them as first-class resources in the Headlamp UI.
  - List and detail views for core Kmesh resources (e.g., waypoints and eBPF map)
  - Visual indicators of mesh status: per-resource health, readiness, and recent Events; cluster-level summary of Kmesh components.
  - Inspection helpers: pretty-printed YAML, related-pod views, and quick links to associated workloads/services.
  - Documentation (README, screenshots, install guide) and a published plugin (Helm/manifest or Headlamp plugin registry entry).
  - Unit/component tests for the plugin and an end-to-end smoke test against a kind/minikube cluster running Kmesh.
- Recommended Skills:
  - Kubernetes (CRDs, RBAC, kubectl)
  - React / Next.js
  - TypeScript
  - Familiarity with Kmesh (or service mesh concepts in general)
  - Headlamp UI
- Mentor(s):
  - ZhenCheng Li(@LiZhenCheng9527, lizhencheng6@huawei.com),

iambahmanop:
@LiZhenCheng9527 happy to be a mentor here.

OmAmbole009:
Hi @LiZhenCheng9527, I’m interested in working on this LFX project.

From what I understand, the goal is to build a Headlamp plugin to surface Kmesh CRDs with list/detail views and basic status/debug info.

I’m planning to:

[...]

PR #1696, opened 2026-05-12
api: kmeshnodeinfo: Add status.conditions to KmeshNodeInfo CRD

WasThatRudy:
**What type of PR is this?**

/kind enhancement

`KmeshNodeInfoStatus` is currently an empty struct in `pkg/kube/apis/kmeshnodeinfo/v1alpha1/types.go`, and the generated CRD schema accordingly has `status: { type: object }` with no properties. This leaves external consumers, including the Headlamp plugin proposed in #1693, with no structured place to render per-node readiness or health state.

This PR adds the standard Kubernetes pattern for object conditions to `KmeshNodeInfoStatus`:

- A `Conditions []metav1.Condition` field with `+listType=map` / `+listMapKey=type` markers. These emit `x-kubernetes-list-type: map` and `x-kubernetes-list-map-keys: [type]` in the generated CRD schema, which is what tells the apiserver how to merge the list under Server-Side Apply.
- A `+kubebuilder:resource:path=kmeshnodeinfos,scope=Namespaced` marker on the root type so `controller-gen` keeps the existing plural form (without it, controller-gen pluralises the kind as `kmeshnodeinfoes`).
- A `+kubebuilder:subresource:status` marker on the root type, enabling the `/status` subresource. **This is a user-facing API change** (see "user-facing change" section below): writes to `.status` now go through the `/status` endpoint rather than the main update path. Standard Kubernetes pattern when a resource grows a meaningful status, recommended by reviewers on this PR.

Regenerated files:

- `deploy/yaml/crd/kmesh.net_kmeshnodeinfos.yaml` via `controller-gen crd paths=./... output:crd:dir=../../deploy/yaml/crd` from `pkg/kube`. Picks up the standard `Condition` sub-schema and the new `subresources.status` block.
- `deploy/charts/kmesh-helm/templates/kmesh.net_kmeshnodeinfos.yaml` copied verbatim from the canonical CRD, matching the existing duplication.
- `pkg/kube/apis/kmeshnodeinfo/v1alpha1/zz_generated.deepcopy.go` regenerated by `deepcopy-gen` using the existing boilerplate.

Drive-by fixes from review:

- Corrected the `informatioon` typo (two occurrences) in the `Spec.Addresses` Go comment.
- Fixed the root-type doc comment that read `// KmeshNode is the Schema for the kmeshnodes API` (and the companion `KmeshNodeInfoList` comment) to match the actual type and resource names.
- Trimmed the `Conditions` field's Go doc comment to the canonical one-liner; design context lives in the PR description and #1693 rather than the CRD's OpenAPI schema.
- Removed the inert `patchStrategy` / `patchMergeKey` / `protobuf` struct tags from the `Conditions` field. Those tags are read by the apiserver only for built-in types in `k8s.io/api`; for CRDs the apiserver only sees the OpenAPI schema, so the tags imply behaviour that isn't actually emitted. Merge semantics for the conditions list are carried by the `+listType=map` / `+listMapKey=type` markers (which emit `x-kubernetes-list-type: map` / `x-kubernetes-list-map-keys: [type]`).

No controller behaviour is changed by this PR; it is the schema scaffold. Actually populating the conditions would follow as a separate change tied to the LFX 2026 Term 2 mentorship work.

Design discussion: #1693. Related mentorship tracking: #1658.

- Tools used: `controller-gen v0.16.4` (matches the version pinned in the existing CRD's annotations) and `deepcopy-gen` from `k8s.io/code-generator v0.32.2`. Happy to re-run with different versions if the project pins them elsewhere.
- The `+kubebuilder:resource:path=kmeshnodeinfos,scope=Namespaced` marker is new on this type. Without it, regeneration would change the CRD's `metadata.name` to `kmeshnodeinfoes.kmesh.net`, which would be a breaking change. The marker preserves the existing plural.
- The `+kubebuilder:subresource:status` enablement was added in response to review feedback from @gemini-code-assist. Reviewers please confirm this scope is acceptable in the same PR rather than a follow-up.
- DCO sign-off applied (`Signed-off-by:` in the commit).

\`\`\`release-note
KmeshNodeInfo's status field now carries a standard \`conditions\` array (\`[]metav1.Condition\`) so external consumers can observe per-node readiness. The CRD also gains the \`/status\` subresource, so status writes must now use the status subresource endpoint and will no longer increment \`metadata.generation\`. No controller behaviour is changed by this PR.
\`\`\`


WasThatRudy:
Pushed `e6bb09e5` addressing the review feedback from @gemini-code-assist and @Copilot:

- Added `+kubebuilder:subresource:status` marker on `KmeshNodeInfo` per Gemini's suggestion. The regenerated CRD now exposes a `subresources.status` section, which is the standard Kubernetes pattern when introducing status conditions (independent updates, no `metadata.generation` bump on status-only writes, separate RBAC). No existing controller writes to `KmeshNodeInfo.status` today, so the change is purely additive.
- Fixed the `informatioon` → `information` typo (two occurrences) in the `Spec.Addresses` Go comment, then regenerated the CRDs so both the canonical YAML and the helm chart copy pick up the corrected description.

Let me know if the subresource enablement should be split into a separate PR, or if there's any other concern with it.

WasThatRudy:
Pushed `67326c01` addressing the latest review feedback from @Copilot:

- Trimmed the `Conditions` field doc comment in `types.go` to the canonical consumer-agnostic one-liner (`// Conditions represent the latest available observations of the resource's state.`). The design context and Headlamp-specific narrative now live in the PR description and #1693 rather than in the CRD's OpenAPI schema, so `kubectl explain kmeshnodeinfos.status.conditions` stays clean and won't go stale.
- Updated the PR description and release note to accurately reflect that enabling `+kubebuilder:subresource:status` is a user-facing API change (writes to `.status` now go through the `/status` endpoint, no `metadata.generation` bump on status-only writes). The "schema-only" framing was inaccurate; it's been replaced with an explicit subresource-enablement disclosure.

The two `Conditions` deepcopy entries and the subresource block in the regenerated CRD YAML are otherwise unchanged.

WasThatRudy:
Pushed `3d3c21e7` addressing the two latest @Copilot comments:

- Fixed the root-type doc comment that read `// KmeshNode is the Schema for the kmeshnodes API` (and the companion `// KmeshNodeLists contains a list of KmeshNode`) to match the actual type and resource names: `KmeshNodeInfo` / `kmeshnodeinfos`. Pre-existing typo, but worth correcting while the file is open.
- Trimmed the `Conditions` struct tag from `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"` to `json:"conditions,omitempty"`. The `patchStrategy` / `patchMergeKey` tags are inert for a CRD (the apiserver only sees the OpenAPI schema, not Go struct tags), and the merge semantics they imply are actually emitted by the `+listType=map` / `+listMapKey=type` markers on the field. `protobuf` is also vestigial since CRDs only serialize via JSON. The minimal tag avoids implying behaviour that isn't reflected in the published CRD.

Regenerated CRD picks up the corrected root description; everything else is unchanged.

WasThatRudy:
Updated the PR description in response to @Copilot's note: the `patchStrategy` / `patchMergeKey` mention has been removed since those struct tags were dropped in `3d3c21e7`. Description now matches what's actually in the code and the generated CRD schema. Also added a short note in the "Drive-by fixes from review" section explaining why the tags were removed (inert for CRDs; merge semantics are carried by `+listType=map` / `+listMapKey=type` instead). No code change in this round.

WasThatRudy:
Quick note on CI: the failing job is `E2E IPv6 Test (istio 1.27)`. The failing test is `TestMixNsAndServiceWaypoint/.../to_enrolled-to-kmesh/http` (a 60-second HTTP traffic timeout between two kmesh-enrolled pods via waypoint). Same-suite runs on istio 1.26 and 1.28 are green on this commit, so the failure looks like a flake on the istio-1.27 IPv6 path rather than something the diff here would touch (this PR is CRD schema only, no runtime / dataplane code). Happy to push an empty commit to retrigger, or leave it for a maintainer `/retest`.

PR #1697, opened 2026-05-13
optimize(nets): use netip.Addr as map key in CompareIpByte

itvi-1234:
## Description
This PR optimizes the `CompareIpByte` function in the `pkg/nets` package.

## SUmmary
The previous implementation converted `[][]byte` IP addresses into strings to use them as map keys. This approach was suboptimal due to:
1. **Unnecessary Allocations**: Every `ip.String()` call creates a new string allocation.
2. **Computational Overhead**: String formatting and hashing are more expensive than using the native `netip.Addr` value type.

By switching to `map[netip.Addr][]byte`, we leverage Go's efficient handling of the `netip.Addr` struct as a map key, which is designed for performance and zero-allocation comparisons.

## Changes
- Modified `CompareIpByte` in `pkg/nets/nets.go` to use `netip.Addr` as the map key.
- Added `continue` logic for cases where IP parsing fails to ensure robustness.
- Removed the resolved `TODO` comment: `// TODO: Optimising functions to be able to handle different data types`.

## Verification Results
Verified the fix by running existing unit tests:
```bash
go test -v pkg/nets/nets_test.go pkg/nets/nets.go
```
**Output:**

image

### DIFF
--- a/pkg/nets/nets.go
+++ b/pkg/nets/nets.go
@@ -116,7 +116,7 @@ func checkIPVersion() (ipv4, ipv6 bool) {
 	return ipv4, ipv6
 }
 
-// Compare two slices and return the data added to a over b and the data missing from b over a.
+// Compare two slices and return the data missing from b over a.
 //
 // Args:
 //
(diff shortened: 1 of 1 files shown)


itvi-1234:
Hi @LiZhenCheng9527 @yashisrani , Could you please review this PR whenever have some time

PR #1763, opened 2026-06-17
bpf/workload: prevent nil pointer panic when IPsec is disabled

### LINKED ISSUE #1706: bpf/workload: nil pointer panic in BpfWorkload.Load and Detach when IPsec is disabled
## Summary

`BpfWorkload.Load()` and `BpfWorkload.Detach()` unconditionally call methods on the `Tc` field, but `Tc` is only initialized when `cfg.EnableIPsec == true`. Any deployment that does not enable IPsec will panic on startup (during `Load`) or shutdown (during `Detach`).

## Affected Files

- `pkg/bpf/workload/loader.go`

## Root Cause

In `NewBpfWorkload` (lines 69–75), `workloadObj.Tc` is assigned only inside an `if cfg.EnableIPsec` guard:

```go
if cfg.EnableIPsec {
    var err error
    workloadObj.Tc, err = general.NewBpf(cfg)
    if err != nil {
        return nil, err
    }
}
```

When `EnableIPsec` is `false`, `Tc` remains its zero value (`nil`, a `*general.BpfTCGeneral`).

`Load()` then calls `w.Tc.LoadTC()` unconditionally at line 147:

```go
if err := w.Tc.LoadTC(); err != nil {   // PANIC: w.Tc is nil
    return err
}
```

`Detach()` calls `w.Tc.Close()` unconditionally at line 194:

```go
if err := w.Tc.Close(); err != nil {    // PANIC: w.Tc is nil
    return err
}
```

## Reproduction Steps

1. Deploy Kmesh in dual-engine mode with `EnableIPsec: false` (the default).
2. The daemon calls `BpfWorkload.Start()` → `Load()` → nil pointer dereference at `w.Tc.LoadTC()`.
3. Alternatively, send SIGTERM to a running Kmesh daemon: `Stop()` → `Detach()` → nil pointer dereference at `w.Tc.Close()`.

## Expected Behavior

`Load()` and `Detach()` skip TC operations when IPsec is disabled.

## Actual Behavior

Nil pointer dereference panic; the daemon crashes immediately on start or shutdown.

## Suggested Fix

Guard the TC calls with a nil check, mirroring the conditional in `NewBpfWorkload`:

```go
// In Load():
if w.Tc != nil {
    if err := w.Tc.LoadTC(); err != nil {
        return err
    }
}

// In Detach():
if w.Tc != nil {
    if err := w.Tc.Close(); err != nil {
        return err
    }
}
```

ravindra-RKB:
What type of PR is this?

/kind bug

What this PR does / why we need it

The `w.Tc` object is conditionally initialized when IPsec is enabled.

`Load()` and `Detach()` previously accessed `w.Tc` unconditionally, which could result in a nil pointer dereference when IPsec is disabled.

This change adds nil checks before invoking:

- `w.Tc.LoadTC()`
- `w.Tc.Close()`

to prevent runtime panics.

Which issue(s) this PR fixes

Fixes #1706

Special notes for your reviewer

This PR is intentionally scoped to the nil pointer fix only.

No test infrastructure, workflow, or E2E changes are included.

Does this PR introduce a user-facing change?

```release-note
NONE

### DIFF
--- a/pkg/bpf/workload/loader.go
+++ b/pkg/bpf/workload/loader.go
@@ -144,8 +144,10 @@ func (w *BpfWorkload) Load() error {
 		return err
 	}
 
-	if err := w.Tc.LoadTC(); err != nil {
-		return err
+	if w.Tc != nil {
+		if err := w.Tc.LoadTC(); err != nil {
+			return err
+		}
 	}
 	return nil
 }
@@ -191,8 +193,10 @@ func (w *BpfWorkload) Detach() error {
 		return err
 	}
 
-	if err := w.Tc.Close(); err != nil {
-		return err
+	if w.Tc != nil {
+		if err := w.Tc.Close(); err != nil {
+			return err
+		}
 	}
 
 	return nil


ravindra-RKB:
I investigated the E2E failure and could not establish a causal link to the change in `pkg/bpf/workload/loader.go`.

The current failure occurs during `kmeshctl log --set bpf:debug`, where the port-forwarded admin endpoint on port 15200 returns `connection refused`.

The branch has been cleaned and the PR now contains only the nil-guard fix for `w.Tc` in `Load()` and `Detach()`.

At this point I do not have sufficient evidence to modify either the E2E framework or the workload loader further. Additional daemon logs from the failing CI run would be required to determine whether this is a startup race, daemon initialization failure, or an unrelated infrastructure issue.

PR #1776, opened 2026-07-06
test: add regression test for RouteConfigCache.Flush with empty VirtualHosts

### LINKED ISSUE #1775: test: Add regression test for RouteConfigCache.Flush when VirtualHosts are empty
## Problem

`RouteConfigCache.Flush` contains logic to skip route configurations that have no `VirtualHosts` before calling `RouteConfigUpdate`.

Without this check, routes with empty `VirtualHosts` are passed to `RouteConfigUpdate`, even though they represent invalid route configurations.

Although this behavior was fixed in PR #1603 (`fix: skip empty routes in RouteConfigCache.Flush`), there is currently no regression test covering this code path.

As a result, future refactoring could accidentally remove or change this behavior without being detected by the test suite.

## Expected behavior

A route with:

- `ApiStatus == UPDATE`
- no `VirtualHosts`

should:

- skip `RouteConfigUpdate`
- reset its status to `ApiStatus_NONE`
- not trigger `RouteConfigDelete`

## Proposed solution

Add a regression test covering this scenario to ensure routes with empty `VirtualHosts` continue to be skipped during `RouteConfigCache.Flush`.

mahesh-09-12:
**What type of PR is this?**

/kind cleanup

* Adds a regression test for `RouteConfigCache.Flush` when a route has no `VirtualHosts`.
* Verifies that routes with empty `VirtualHosts` are skipped instead of being passed to `RouteConfigUpdate`.
* Protects the behavior introduced in #1603 from future regressions.

Fixes #1775

This test covers the regression scenario introduced by the fix in #1603 (`fix: skip empty routes in RouteConfigCache.Flush`). Without that fix, the new test fails because `RouteConfigUpdate` is invoked for a route with no `VirtualHosts`.

```release-note
NONE

### DIFF
--- a/pkg/cache/v2/route_test.go
+++ b/pkg/cache/v2/route_test.go
@@ -116,6 +116,44 @@ func TestRouteFlush(t *testing.T) {
 		assert.Equal(t, []string{}, deleteRouterName)
 	})
 
+	t.Run("route status is UPDATE with empty virtual hosts", func(t *testing.T) {
+		updateRouterName := []string{}
+		deleteRouterName := []string{}
+
+		patches := gomonkey.NewPatches()
+		defer patches.Reset()
+
+		patches.ApplyFunc(maps_v2.RouteConfigUpdate, func(key string, value *route_v2.RouteConfiguration) error {
+			updateRouterName = append(updateRouterName, key)
+			return nil
+		})
+
+		patches.ApplyFunc(maps_v2.RouteConfigDelete, func(key string) error {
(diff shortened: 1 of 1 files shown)


mahesh-09-12:
@LiZhenCheng9527 @YaoZengzeng the build workflow appears to have failed due to a github actions infrastructure/cache issue (`Cache service responded with 400` and the workflow reports that services aren't available). I don't have permission to rerun the workflow. Could you please rerun the failed job when you get a chance?

PR #1808, opened 2026-07-26
fix: separate IPv4 and IPv6 workload tail call maps

### LINKED ISSUE #1806: E2E jobs fail before tests because Kmesh admin port never starts
Title: E2E jobs fail before tests because Kmesh admin port never starts

**What happened**:

All six E2E jobs fail during setup. The Kmesh pod reaches `Running`, but `kmeshctl` cannot connect to the daemon on port 15200:

    dial tcp 127.0.0.1:15200: connect: connection refused
    dial tcp6 [::1]:15200: connect: connection refused
    Failed to set BPF debug log after 5 attempts

This happens across Istio 1.26, 1.27 and 1.28, with both IPv4 and IPv6. The tests never start.

The daemon admin endpoint should be available before setup changes the log level, and the E2E suite should run.

1. Open a PR that triggers the E2E workflows.
2. Wait for the Kmesh DaemonSet to be installed.
3. All six jobs fail at `kmeshctl log <pod> --set bpf:debug`.

This looks like a recurrence of #1637. The workaround in #1639 pinned CI to Ubuntu 22.04, but hosted Ubuntu 22.04 runners are using a 6.8 Azure kernel now.

The same failure is reproducible on unrelated PRs based on current `main`.

**Environment**:

- Kmesh version: main (`43d5f93`)
- Kmesh mode: Dual-Engine Mode
- Istio version: 1.26.4, 1.27.5 and 1.28.2
- Kernel version: 6.8 Azure kernel on GitHub-hosted Ubuntu 22.04
- Others: IPv4 and IPv6 E2E workflows

avinxshKD:
**What type of PR is this?**

/kind bug

Uses separate program-array maps for the IPv4 and IPv6 workload tail calls.

Newer kernels track the expected attach type for program arrays. Sharing one map causes the IPv6 program load to fail with `EINVAL`, which prevents Kmesh from starting.

Fixes #1806

The generated BPF bindings are updated to include the IPv6 tail-call map.

```release-note
NONE

### DIFF
--- a/bpf/kmesh/bpf2go/dualengine/kmeshcgroupsockworkload_bpfeb.go
+++ b/bpf/kmesh/bpf2go/dualengine/kmeshcgroupsockworkload_bpfeb.go
@@ -111,6 +111,7 @@ type KmeshCgroupSockWorkloadMapSpecs struct {
 	KmAuthReq     *ebpf.MapSpec `ebpf:"km_auth_req"`
 	KmAuthRes     *ebpf.MapSpec `ebpf:"km_auth_res"`
 	KmBackend     *ebpf.MapSpec `ebpf:"km_backend"`
+	KmCgrTail6    *ebpf.MapSpec `ebpf:"km_cgr_tail6"`
 	KmCgrTailcall *ebpf.MapSpec `ebpf:"km_cgr_tailcall"`
 	KmEndpoint    *ebpf.MapSpec `ebpf:"km_endpoint"`
 	KmFrontend    *ebpf.MapSpec `ebpf:"km_frontend"`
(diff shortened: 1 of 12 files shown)


avinxshKD:
@LiZhenCheng9527 the startup failure is gone. Both builds and five E2E jobs pass and the remaining IPv6/1.28 job starts Kmesh normally and fails later in TestCrossNamespace after namespace unenrollment (X-Request-Id is still present right after the label change).

The previous run failed in a different late test, so this looks unrelated to the tail-call map split. 
Looks good to me,  PTAL 

https://github.com/kmesh-net/kmesh/actions/runs/30246698404/job/89915254364

/cc @hzxuzhonghu @yashisrani

R4R3D1FF:
I tried running this locally  with kernel version 7.0.0-28-generic (since I had accidentally opened a duplicate issue for this).

The load for policy_check program is still failing, saying that it has too many instructions. Does that happen on your machine? I have attached the log.

[previous-run.log](https://github.com/user-attachments/files/30478401/previous-run.log)

PR #1810, opened 2026-07-27
wip(debug): instrument BPF workload loaders to diagnose verifier rejection

vanshika2720:
Draft/diagnostic PR - not for review yet.

E2E has been permanently broken since 2026-05-14 by a deterministic
eBPF verifier rejection during kmesh-daemon startup:

    Error: bpf Load failed: load program: invalid argument:
        processed 6750 insns (limit 1000000) max_states_per_insn 8 total_states 333 peak_states 306 mark_read 26

This is captured with the diagnostics added in #1807. This PR adds temporary instrumentation to BpfWorkload.Load() (progress logging per program) and enables full instruction-level verifier logging on all five workload program loaders, so a real CI run can reveal which program fails and the complete verifier log instead of the truncated
summary.

Will be replaced with the actual fix once the failing program and
verifier reason are identified from this run.

### DIFF
--- a/bpf/include/bpf_common.h
+++ b/bpf/include/bpf_common.h
@@ -100,7 +100,7 @@ struct {
  * From v5.4, bpf_get_netns_cookie can be called for bpf cgroup hooks, from v5.15, it can be called for bpf sockops
  * hook. Therefore, ensure that function is correctly used.
  */
-static inline void record_manager_netns_cookie(struct bpf_sock_addr *ctx)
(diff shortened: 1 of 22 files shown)


avinxshKD:
Thanks for looking into this. I’ve been already working on the same startup failure in #1808. The map split there has five of the six E2E jobs passing now and checking the remaining IPv6 1.28 and build failures separately. 

Sharing this so we don’t duplicate the investigation.

/cc @LiZhenCheng9527 @yashisrani

PR #1829, opened 2026-07-30
fix: handle IPsec tombstone delete events

### LINKED ISSUE #1828: IPsec cleanup is skipped for informer tombstone deletes
**What happened**:

`handleKNIDelete` only handles `*KmeshNodeInfo`.

Kubernetes informers may pass `cache.DeletedFinalStateUnknown` when a delete event is missed. In that case the handler returns without removing the node's XFRM state or Pod CIDRs from the KNI map.

The handler should extract `KmeshNodeInfo` from the tombstone and run the normal cleanup.

1. Create a `KmeshNodeInfo` with an address and Pod CIDR.
2. Wrap it in `cache.DeletedFinalStateUnknown`.
3. Pass it to `handleKNIDelete`.

The handler logs an unexpected type and skips cleanup.

This can leave stale IPsec and BPF state after an informer watch reconnect.

**Environment**:

- Kmesh version: main (`43d5f93e`)
- Kmesh mode: Dual-Engine Mode with IPsec enabled
- Istio version: N/A
- Kernel version: N/A
- Others: Kubernetes informer delete tombstone

avinxshKD:
**What type of PR is this?**

/kind bug

Handles informer tombstones in the KmeshNodeInfo delete handler so stale IPsec and CIDR state is cleaned up.

Fixes #1828

Adds coverage for the tombstone delete path.

```release-note
NONE

### DIFF
--- a/pkg/controller/encryption/ipsec/ipsec_controller.go
+++ b/pkg/controller/encryption/ipsec/ipsec_controller.go
@@ -292,8 +292,16 @@ func (c *Controller) handleKNIUpdate(oldObj, newObj interface{}) {
 func (c *Controller) handleKNIDelete(obj interface{}) {
 	node, ok := obj.(*v1alpha1.KmeshNodeInfo)
 	if !ok {
-		log.Errorf("expected *v1alpha1_core.KmeshNodeInfo but got %T in handle delete func", obj)
-		return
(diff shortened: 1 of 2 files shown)


avinxshKD:
@LiZhenCheng9527 @yashisrani checks are passing now :p

yashisrani:
@avinxshKD Looks good ..

PR #1832, opened 2026-07-30
fix(status): support GET in authzHandler to return authz offload status

devGPP23:
**Problem**

I was going through the `authz` CLI code and saw that `fetchAuthzStatus()` in `ctl/authz/authz.go` sends a GET to `/authz` to check the current state. But when I looked at the server side, `authzHandler` only accepts POST and rejects everything else with 405. So `kmeshctl authz` status was never actually working — it always gets a 405 back.
due to these lines of code in authzHandler,whenever someone tried GET endpoint , it return method not allowed
```
if r.Method != http.MethodPost {
		http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
		return
	}
```

**Solution**

I added GET support to `authzHandler` the same way `loggersHandler` already handles it — GET goes to a read function, POST goes to the write function.

The new `getAuthzStatus()` reads from `BpfLoader.GetAuthzOffload()` and returns the state as JSON. The POST logic is unchanged, just moved it to `setAuthzStatus()` to keep things clean.

There were no tests for this handler before so I added a couple — one to check the default state and one to verify it updates after enabling `authz`.

**Below is the terminal output from testing the /authz endpoint locally against the Kmesh status server:**
```
$ curl -s -v http://localhost:15020/authz
* Host localhost:15020 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:15020...
* Established connection to localhost (::1 port 15020) from ::1 port 34348 
* using HTTP/1.x
> GET /authz HTTP/1.1
> Host: localhost:15020
> User-Agent: curl/8.18.0
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Sat, 08 Aug 2026 05:21:38 GMT
< Content-Length: 24
< 
{
    "enabled": false
* Connection #0 to host localhost:15020 left intact
}
```

We got status 200 with Json telling if authz is enabled or not

### DIFF
--- a/pkg/status/status_server.go
+++ b/pkg/status/status_server.go
@@ -330,11 +330,32 @@ func (s *Server) connectionMetricHandler(w http.ResponseWriter, r *http.Request)
 }
 
 func (s *Server) authzHandler(w http.ResponseWriter, r *http.Request) {
-	if r.Method != http.MethodPost {
+	if r.Method == http.MethodGet {
+		s.getAuthzStatus(w)
+	} else if r.Method == http.MethodPost {
+		s.setAuthzStatus(w, r)
+	} else {
 		http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+	}
+}
+
+func (s *Server) getAuthzStatus(w http.ResponseWriter) {
+	authzOffload := s.loader.GetAuthzOffload()
+	enabled := authzOffload == constants.ENABLED
+
+	data, err := json.MarshalIndent(&struct {
+		Enabled bool `json:"enabled"`
+	}{Enabled: enabled}, "", "    ")
+	if err != nil {
+		log.Errorf("Failed to marshal authz status: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write(data)
+}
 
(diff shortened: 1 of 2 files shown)


devGPP23:
@yashisrani @LiZhenCheng9527 @hzxuzhonghu 

The failing E2E Test(istio 1.26) is unrelated to my changes
this PR only modifies the admin status server (pkg/status/status_server.go) and its unit test. 
The failing test (TestMixNsAndServiceWaypoint) tests waypoint proxy data-plane routing, which my code doesn't touch at all. The same test passed on Istio 1.27 and 1.28 in this same CI run. 

Could you please /retest when they get a chance?

devGPP23:
cc/ @yashisrani

pkg/status/status_server.go
```diff
@@ -330,11 +330,32 @@ func (s *Server) connectionMetricHandler(w http.ResponseWriter, r *http.Request)
...
+
+	data, err := json.MarshalIndent(&struct {
+		Enabled bool `json:"enabled"`
+	}{Enabled: enabled}, "", "    ")
+	if err != nil {
+		log.Errorf("Failed to marshal authz status: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	w.WriteHeader(http.StatusOK)
```
LiZhenCheng9527, line 355: `Content-Type: application/json` is not set.

PR #1836, opened 2026-07-31
fix(nets): prevent panic on IPv6 address in ConvertIpToUint32

### LINKED ISSUE #1835: bug: daemon panic (index out of range) when parsing IPv6 address in ConvertIpToUint32
### What happened:
When `nets.ConvertIpToUint32(ip string)` in [pkg/nets/nets.go](https://github.com/kmesh-net/kmesh/blob/main/pkg/nets/nets.go#L33-L46) processes an IPv6 address string (such as `"2001:db8::1"` or `"::1"`), `net.ParseIP` returns a 16-byte `net.IP` slice. 

Because `len(netIP)` equals `net.IPv6len` (16 bytes), the function enters the first conditional branch and calls `netIP.To4()`. Since the IP address is IPv6, `netIP.To4()` returns `nil`. `binary.LittleEndian.Uint32(nil)` is then evaluated, triggering an immediate runtime panic:

`panic: runtime error: index out of range [3] with length 0`

This crashes the entire Kmesh daemon process. In addition, the second conditional branch (`len(netIP) == net.IPv4len`) is dead code because Go's `net.ParseIP` always returns a 16-byte slice for any valid IP address representation.

### What you expected to happen:
`nets.ConvertIpToUint32` should safely handle IPv6 or invalid IP inputs without panicking, returning `0` when the input string is not a valid IPv4 address.

### How to reproduce it (as minimally and precisely as possible):
Add the following unit test case to `pkg/nets/nets_test.go` and run `go test ./pkg/nets`:

```go
func Test_ConvertIpToUint32_IPv6(t *testing.T) {
	// Should return 0 for IPv6 address without panicking
	val := ConvertIpToUint32("2001:db8::1")
	assert.Equal(t, uint32(0), val)
}
```

### Suggested Fix

Modify `ConvertIpToUint32` in [pkg/nets/nets.go](https://github.com/kmesh-net/kmesh/blob/main/pkg/nets/nets.go#L33-L46) to safely check `netIP.To4()`:

```go
// ConvertIpToUint32 converts an IPv4 address string to little-endian uint32 format.
// Returns 0 if the input is an invalid IP or IPv6 address.
func ConvertIpToUint32(ip string) uint32 {
	netIP := net.ParseIP(ip)
	if netIP == nil {
		return 0
	}
	ipv4 := netIP.To4()
	if ipv4 == nil {
		return 0
	}
	return binary.LittleEndian.Uint32(ipv4)
}
```

**Environment**:

- Kmesh version: `v0.5.0 / main branch (latest)`
- Kmesh mode(kmesh has `Kernel-Native Mode` and `Dual-Engine Mode`): `Kernel-Native Mode` & `Dual-Engine Mode`
- Istio version: `v1.24+`
- Kernel version: `Linux 5.10+ / Linux 6.x`
- Others: `Go 1.22+`

singhsrijan46:
**What type of PR is this?**

/kind bug

Fixes a panic in `nets.ConvertIpToUint32` when processing IPv6 address strings.

Previously, `len(netIP) == net.IPv6len` was evaluated on the 16-byte `net.IP` slice returned by `net.ParseIP`. For IPv6 addresses, calling `netIP.To4()` returned `nil`, and `binary.LittleEndian.Uint32(nil)` subsequently panicked with `index out of range [3]`.

This change safely checks `ipv4 := netIP.To4()` and returns `0` if `ipv4 == nil`.

Fixes #1835

Included unit test assertion in `pkg/nets/nets_test.go` to ensure IPv6 address strings return `0` without panicking.

```release-note
NONE

### DIFF
--- a/pkg/nets/nets.go
+++ b/pkg/nets/nets.go
@@ -35,14 +35,11 @@ func ConvertIpToUint32(ip string) uint32 {
 	if netIP == nil {
 		return 0
 	}
-	// TODO: is this right?
-	if len(netIP) == net.IPv6len {
-		return binary.LittleEndian.Uint32(netIP.To4())
+	ipv4 := netIP.To4()
+	if ipv4 == nil {
+		return 0
 	}
-	if len(netIP) == net.IPv4len {
-		return binary.LittleEndian.Uint32(netIP)
-	}
-	return 0
+	return binary.LittleEndian.Uint32(ipv4)
 }
 
 // ConvertPortToBigEndian convert uint32 to network order
--- a/pkg/nets/nets_test.go
+++ b/pkg/nets/nets_test.go
@@ -28,8 +28,11 @@ func Test_ConvertIpToUint32(t *testing.T) {
 	val := ConvertIpToUint32(ip)
 	assert.Equal(t, uint32(0x100a8c0), val)
 
-	// It can not panic even for invalid ip
+	// It can not panic even for invalid ip or ipv6
 	val = ConvertIpToUint32("a.b.c.d")
+	assert.Equal(t, uint32(0), val)
+
+	val = ConvertIpToUint32("2001:db8::1")
 	assert.Equal(t, uint32(0), val)
 }
 


yashisrani:
@singhsrijan46 To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

singhsrijan46:
> @singhsrijan46 To avoid having PRs blocked in the future, always include Signed-off-by: Author Name [authoremail@example.com](mailto:authoremail@example.com) in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Thanks! I've added the DCO sign-off and updated the commit.

PR #1839, opened 2026-08-01
fix: recalculate locality-LB endpoint priorities after LocalityInfo initialization

### LINKED ISSUE #1838: bug: Remote endpoints retain fallback priority after LocalityInfo initialization
**What happened**:

When a remote workload is processed before any local workload, `LocalityInfo` has not yet been initialized. As a result, the endpoint is added to the service using the fallback priority (`0`).

After the first local workload initializes `LocalityInfo`, the previously added remote endpoint continues to use the fallback priority instead of being updated to the priority derived from the locality load-balancing policy.

After `LocalityInfo` is initialized, endpoint priorities for locality-aware services should reflect the configured locality load-balancing policy, including endpoints that were added before locality information became available.

1. Create a service configured with locality-based load balancing.
2. Process a remote workload before any local workload is received.
3. Verify that the endpoint is inserted with fallback priority (`0`).
4. Process the first local workload to initialize `LocalityInfo`.
5. Observe that the remote endpoint remains at priority `0` instead of being updated according to the locality load-balancing policy.

The issue only affects workloads that are processed before `LocalityInfo` is initialized. Once locality information becomes available, existing endpoints continue using the fallback priority.

**Environment**:

- Kmesh version: `main`
- Kmesh mode (Kernel-Native Mode / Dual-Engine Mode): N/A
- Istio version: N/A
- Kernel version: N/A
- Others: Reproduced using the workload processor unit tests.

mahesh-09-12:
**What type of PR is this?**

/kind bug

When a remote workload is processed before any local workload, `LocalityInfo` has not yet been initialized. In this case, the endpoint is added to the service using the fallback priority (`0`).

Once the first local workload initializes `LocalityInfo`, existing endpoints belonging to locality-aware services are not re-evaluated, causing them to continue using the fallback priority instead of their locality-derived priority.

This PR recalculates endpoint priorities for locality-aware services immediately after `LocalityInfo` is initialized for the first time. It also adds a regression test covering this execution order to prevent future regressions.

Fixes #1838

The regression test reproduces the following sequence:

1. A locality-aware service is created.
2. A remote workload is processed before any local workload.
3. The endpoint is inserted with the fallback priority (`0`).
4. The first local workload initializes `LocalityInfo`.
5. The test verifies that the remote endpoint is moved to its locality-derived priority.

```release-note
Fix an issue where endpoints added before LocalityInfo initialization could retain the fallback priority instead of being updated according to locality-based load balancing.
```

### DIFF
--- a/pkg/controller/workload/workload_processor.go
+++ b/pkg/controller/workload/workload_processor.go
@@ -569,9 +569,31 @@ func (p *Processor) handleWorkload(workload *workloadapi.Workload) error {
 	}
 
 	// update kmesh localityCache
-	// TODO: recalculate endpoints priority once local locality is set
+	// The first local workload initializes LocalityInfo.
+	// Remote workloads processed before this were inserted with
+	// priority 0 because locality could not yet be calculated.
+	// Recalculate endpoint priorities for all locality-LB services.
 	if p.locality.LocalityInfo == nil && p.nodeName == workload.GetNode() {
-		p.locality.SetLocality(p.nodeName, workload.GetClusterId(), workload.GetNetwork(), workload.GetLocality())
+		p.locality.SetLocality(
+			p.nodeName,
+			workload.GetClusterId(),
+			workload.GetNetwork(),
+			workload.GetLocality(),
+		)
+
+		for _, svc := range p.ServiceCache.List() {
+			if lb := svc.GetLoadBalancing(); lb != nil &&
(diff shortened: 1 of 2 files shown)


mahesh-09-12:
Hey @kwb0523 @YaoZengzeng 
the build workflow appears to have failed due to a github actions infrastructure/cache issue (`Cache service responded with 400` and the workflow reports that services aren't available). I don't have permission to rerun the workflow, could you please rerun the failed job when you get a chance?

PR #1845, opened 2026-08-02
fix: retry cgroup2/bpf-fs path check to handle mount propagation delay

### LINKED ISSUE #1842: kmesh-daemon crash loops on kind because of a cgroup2 stat race
was setting up dual-engine mode on kind and mounted /mnt/kmesh_cgroup2 properly through extraMounts in the kind config, host /sys/fs/cgroup to node to pod through the existing /mnt hostpath. daemon still crashes every time on startup with "parse BpfConfig failed, stat /mnt/kmesh_cgroup2: no such file or directory"

went and checked daemon/options/bpf.go, ParseConfig() does a single os.Stat on the cgroup2 path with no retry or wait, just fails immediately if its not there yet

checked timestamps on the logs, pod started at 17:56:25 and crashed at 17:56:30, so about 5 seconds in. manually checked the same path a few seconds after the crash and it was there and fully populated, so the mount does land eventually just not fast enough for that one stat call

kind config i used:

kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
  - hostPath: /sys/fs/cgroup
    containerPath: /mnt/kmesh_cgroup2

Then I installed istio ambient profile and kmesh through deploy/charts/kmesh-helm and it just crash loops forever

seems like a small retry loop around the stat calls (both cgroup2path and bpfFsPath) would fix this instead of bailing on the first check. happy to try building it if someone can confirm thats the right direction

running wsl2 + docker desktop, kind v0.29.0, node image v1.33.1, kmesh built from main at c88ef30, dual-engine mode

bhumikadangayach:
Fixes #1842

The kmesh-daemon was crashing on startup with "stat /mnt/kmesh_cgroup2: no such file or directory" when running on kind. Traced it to ParseConfig() in daemon/options/bpf.go doing a single os.Stat call with no retry - if the mount isn't ready yet (which happens on kind due to mount propagation delay), it fails immediately instead of waiting.

This PR adds a waitForPath() helper that retries the stat call up to 10 times with 500ms between attempts (5 seconds total) before giving up. Applied to both the cgroup2 path and bpf-fs path checks.

Verified with make build - compiles clean, no errors.

### DIFF
--- a/daemon/options/bpf.go
+++ b/daemon/options/bpf.go
@@ -19,6 +19,7 @@ package options
 import (
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -53,18 +54,35 @@ func (c *BpfConfig) ParseConfig() error {
 	if c.Cgroup2Path, err = filepath.Abs(c.Cgroup2Path); err != nil {
 		return err
 	}
-	if _, err = os.Stat(c.Cgroup2Path); err != nil {
+	if err = waitForPath(c.Cgroup2Path); err != nil {
 		return err
 	}
 
 	if c.BpfFsPath, err = filepath.Abs(c.BpfFsPath); err != nil {
 		return err
 	}
-	if _, err = os.Stat(c.BpfFsPath); err != nil {
+	if err = waitForPath(c.BpfFsPath); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+// waitForPath retries os.Stat on path for a bounded time, to tolerate
+// mount propagation delays in containerized environments (e.g. kind).
+func waitForPath(path string) error {
+	const (
+		maxRetries = 10
+		retryDelay = 500 * time.Millisecond
+	)
+	var err error
+	for i := 0; i < maxRetries; i++ {
+		if _, err = os.Stat(path); err == nil {
+			return nil
+		}
+		time.Sleep(retryDelay)
+	}
+	return err
 }
 
 func (c *BpfConfig) KernelNativeEnabled() bool {


daemon/options/bpf.go
```diff
@@ -53,20 +54,37 @@ func (c *BpfConfig) ParseConfig() error {
...
 
+// waitForPath retries os.Stat on path for a bounded time, to tolerate
+// mount propagation delays in containerized environments (e.g. kind).
+func waitForPath(path string) error {
+	const (
+		maxRetries = 10
+		retryDelay = 500 * time.Millisecond
+	)
+	var err error
+	for i := 0; i < maxRetries; i++ {
```
LiZhenCheng9527:
```suggestion
	for i := 0; i <= maxRetries; i++ {
```
Ensure that a retry is attempted even after the last failure

bhumikadangayach: Fixed, thanks!

bhumikadangayach:
The failing E2E appears to be TestCrossNamespace, where the unenrolled workload unexpectedly gets L7 processing (X-Request-Id). TestKmeshRestart passes with 360/360 successful requests. This appears unrelated to the cgroup2/bpf-fs retry change.

bhumikadangayach:
Same failure on the IPv6 run too - TestCrossNamespace, unenrolled workload getting X-Request-Id set. Unrelated to this change, only touches daemon/options/bpf.go.

bhumikadangayach:
This run hit #1911 - tests all passed (TestCrossNamespace, TestKmeshRestart 350/350), suite still exits 1 on log-collection step. Unrelated to this change.

PR #1855, opened 2026-08-03
fix(ctl): match the daemon's monitoring-disabled message

### LINKED ISSUE #1854: kmeshctl monitoring never shows the "enable monitoring first" hint due to a message mismatch.
**What happened**:

When you try to enable  one of the sub-toggles with global monitoring disabled `kmeshctl` never prints the hint.
It prints:
```
Error: received status code 400
```
The daemon sends precise context in the response body and `kmeshctl` even reads that , but due to string mismatch  that is `status_server.go`  sends "disabled" as a part of the error string.

```go
http.Error(w, "Kmesh monitoring is disabled, cannot enable accesslog.", http.StatusBadRequest)
```
while `monitoring.go` matches this entire string with a typo i.e. "disable" which will always be false, thereby not printing the hint.

```go
bytes.Contains(bodyBytes, []byte(fmt.Sprintf("Kmesh monitoring is disable, cannot enable %s.", observablityType)))
```

Expected message:
```
Error: received status code 400
failed to enable accesslog: Kmesh monitoring is disabled, cannot enable accesslog.
Need to start Kmesh's Monitoring. Please run `kmeshctl monitoring -h` for more help.
```

```bash
kmeshctl monitoring <kmesh-daemon-pod> --all disable
kmeshctl monitoring <kmesh-daemon-pod> --accesslog enable
```

**Environment**:

- Kmesh version: `main` (https://github.com/kmesh-net/kmesh/commit/c88ef3007f342b7ebb3aeac12cabd1613e7f5fd1)
- Kmesh mode(kmesh has `Kernel-Native Mode` and `Dual-Engine Mode`): both — the affected code is mode-independent
- Istio version: n/a
- Kernel version: n/a
- Others: not environment-specific; it's a static string mismatch between `pkg/status` and `ctl/monitoring`

sathwikshetty33:
**What type of PR is this?**

/kind bug

`kmeshctl` searched the daemon's 400 response body for `"Kmesh monitoring is disable, cannot enable %s."`, but `pkg/status` sends `"Kmesh monitoring is disabled, cannot enable %s."` (`status_server.go`).

Because of the `disable` / `disabled` mismatch, the `bytes.Contains` check at `ctl/monitoring/monitoring.go` was always false for all three sub-toggles (`accesslog`, `workload metrics`, `connection metrics`), so the `log.Errorf` carrying the hint was unreachable. Enabling a sub-toggle while global monitoring was off printed only:

```
Error: received status code 400
```

even though the daemon had sent an explanation and `kmeshctl` had already read it into `bodyBytes`.

This PR corrects the string on the `kmeshctl` side so the daemon's explanation and the hint are surfaced.

Fixes #1854

This PR fixes the typo inline, but this could be easily reintroduced unintentionally; So I wondering could we make this error message a shared constant - which prevents such bugs in future.

Verification:

- `gofmt -l ./ctl/` clean; `golangci-lint run --config=common/config/.golangci.yaml ./ctl/...` exits 0; `codespell` clean.
- `go test ./ctl/monitoring/` passes.

No

### DIFF
--- a/ctl/monitoring/monitoring.go
+++ b/ctl/monitoring/monitoring.go
@@ -203,7 +203,7 @@ func SetObservabilityPerKmeshDaemon(cli kube.CLIClient, podName, info string, ob
 			return
 		}
 		bodyString := string(bodyBytes)
-		if resp.StatusCode == http.StatusBadRequest && bytes.Contains(bodyBytes, []byte(fmt.Sprintf("Kmesh monitoring is disable, cannot enable %s.", observablityType))) {
+		if resp.StatusCode == http.StatusBadRequest && bytes.Contains(bodyBytes, []byte(fmt.Sprintf("Kmesh monitoring is disabled, cannot enable %s.", observablityType))) {
 			log.Errorf("failed to enable %s: %v. Need to start Kmesh's Monitoring. Please run `kmeshctl monitoring -h` for more help.", observablityType, bodyString)
 		}
 	}


sathwikshetty33:
Can you please review this PR. Thanks in advance!
/cc @yashisrani @LiZhenCheng9527

PR #1862, opened 2026-08-04
Title: status: expose BPF operating mode in /version endpoint

devGPP23:
**Problem :** 
 While looking at the API, I saw that external tools (like `kmeshctl` or the new MCP server) have no direct way to know if Kmesh is running in kernel-native or dual-engine mode. Because of this, tools are forced to guess by blindly hitting mode-specific endpoints just to see if they fail.

**What I knew:** I knew that the Kmesh Server already stores this mode internally in `s.config.BpfConfig.Mode`. The simplest and cleanest way to fix this "guessing" problem is to just expose this mode in the standard /version endpoint.

**How I solved it :**

I updated the /version handler in `status_server.go`. Now, it returns the standard version info along with a new "mode" field in the JSON response.
I added a unit test (`TestServerVersion`) in `status_server_test.go` to make sure the API returns the correct mode based on the config.

**How it will be useful:** 
Now, any client can just make one simple call to /version on startup and immediately know which BPF maps and config endpoints they can use. Also, this makes it possible for the `kmeshctl` version command to print the operating mode in the future, which will be super helpful for users debugging their clusters.

### DIFF
--- a/pkg/status/status_server.go
+++ b/pkg/status/status_server.go
@@ -117,7 +117,13 @@ func NewServer(c *controller.XdsClient, configs *options.BootstrapConfigs, loade
 func (s *Server) version(w http.ResponseWriter, r *http.Request) {
 	v := version.Get()
 
-	data, err := json.MarshalIndent(&v, "", "  ")
+	data, err := json.MarshalIndent(&struct {
+		version.Info
+		Mode string `json:"mode"`
+	}{
+		Info: v,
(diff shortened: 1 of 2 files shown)


yashisrani:
@devGPP23 Kmesh CI workflow is failing, so please try to fix it ..

devGPP23:
cc/ @yashisrani I have fixed that , Please Review it whenever you are free

PR #1872, opened 2026-08-04
fix: kmeshctl version exits 0 even when pod version fetch fails

### LINKED ISSUE #1871: kmeshctl version exits 0 even when pod version fetch fails
Ran kmeshctl version <pod-name> against a nonexistent pod. It logs the failure correctly but exits with code 0:

kmeshctl version nonexistent-pod-xyz
echo $?
# time=... level=error msg="failed to start port forwarder for Kmesh daemon pod nonexistent-pod-xyz: complete failed: pods \"nonexistent-pod-xyz\" not found"
# exit code: 0

Traced it to ctl/version/version.go - getVersion() only returns version.Info, no error, so any failure inside it (port forward, HTTP request, JSON unmarshal) just logs and returns an empty struct. The caller checks v.GitVersion != "" to decide whether to print anything, but never checks for actual failure, so it falls through and exits normally.

Compare to the len(args) == 0 branch in the same file, which does call os.Exit(1) correctly on failure - so this looks like an inconsistency/oversight rather than intentional.

Have a fix ready (adding a proper error return to getVersion and checking it in runVersion) - opening a PR for it.

bhumikadangayach:
Fixes #1871

kmeshctl version <pod> was exiting 0 even when the version fetch failed. getVersion() only returned version.Info with no error, so callers had no way to detect failure - they just checked if v.GitVersion was empty, which is indistinguishable from a real failure.

Added an error return to getVersion(), set it at every failure point inside the function, and updated both call sites in runVersion() to check it: the single-pod case now exits 1 on failure (matching the existing len(args) == 0 branch's behavior), and the multi-pod loop skips a pod on failure instead of silently counting it as empty.

Verified: kmeshctl version <nonexistent-pod> now exits 1 instead of 0.


bhumikadangayach:
Hi @LiZhenCheng9527 — this has been open a few weeks, CI is green, no outstanding comments. Could you take a look when you get a chance?

ctl/version/version.go
```diff
@@ -95,45 +96,51 @@ func runVersion(cmd *cobra.Command, args []string) {
...
+	body, rerr := io.ReadAll(resp.Body)
+	if rerr != nil {
+		log.Errorf("failed to read HTTP response body: %v", rerr)
+		err = rerr
 		return
 	}
 
-	if err := json.Unmarshal(body, &version); err != nil {
-		log.Errorf("failed to unmarshal version info: %v", err)
+	if uerr := json.Unmarshal(body, &version); uerr != nil {
```
LiZhenCheng9527, line 141: Even if the JSON is parsed correctly, there is no guarantee that the version data is valid.

PR #1875, opened 2026-08-05
feat(status): add GET handlers for observability endpoints

devGPP23:
**Problem:**

 I noticed that external tools have no way to check if features like monitoring or access logs are currently turned on. 
 The API only allows `POST` requests to change the state, but we can't query the current state. Anyone trying to build an MCP server or a dashboard for Kmesh will face this same issue.

**How I solved it:**

I added `GET` handlers for all the observability endpoints (`/monitoring`, `/accesslog`, `/workload_metrics`, and `/connection_metrics`). 

Now, if a tool (like the MCP server) makes a GET request to these endpoints, it will simply receive a JSON response telling it whether the feature is enabled or not.
example for Monitoring 
```
$ curl -s -v http://localhost:15020/monitoring
* Host localhost:15020 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:15020...
* Established connection to localhost (::1 port 15020) from ::1 port 45244 
* using HTTP/1.x
> GET /monitoring HTTP/1.1
> Host: localhost:15020
> User-Agent: curl/8.18.0
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Sat, 08 Aug 2026 05:36:19 GMT
< Content-Length: 23
< 
{
    "enabled": true
* Connection #0 to host localhost:15020 left intact
}
```
**What this PR does :**

- Adds the `GET` logic in `status_server.go` for the 4 endpoints.
- Adds unit tests in `status_server_test.go` to make sure the GET handlers return the correct JSON.
- Fixes a small typo I found in the `GetWorkloadMetricTrigger` method name.

**User-facing change:**

Users and external integrations (like MCP servers, CLIs, and dashboards) can now make HTTP GET requests to Kmesh's observability status endpoints (/monitoring, /accesslog, /workload_metrics, and /connection_metrics) to query their current active state. The server now returns a standard JSON response (e.g., {"enabled": true}) instead of failing with a 405 Method Not Allowed error.

### DIFF
--- a/pkg/status/status_server.go
+++ b/pkg/status/status_server.go
@@ -218,7 +218,24 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func writeFeatureStatus(w http.ResponseWriter, enabled bool) {
+	data, err := json.MarshalIndent(&struct {
+		Enabled bool `json:"enabled"`
+	}{Enabled: enabled}, "", "    ")
+	if err != nil {
+		log.Errorf("Failed to marshal feature status: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write(data)
+}
+
 func (s *Server) accesslogHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodGet {
+		writeFeatureStatus(w, s.xdsClient.WorkloadController.GetAccesslogTrigger())
+		return
+	}
 	if r.Method != http.MethodPost {
 		http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
 		return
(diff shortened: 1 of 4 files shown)


devGPP23:
cc/ @yashisrani

pkg/status/status_server.go
```diff
@@ -218,7 +218,24 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
...
+func writeFeatureStatus(w http.ResponseWriter, enabled bool) {
+	data, err := json.MarshalIndent(&struct {
+		Enabled bool `json:"enabled"`
+	}{Enabled: enabled}, "", "    ")
+	if err != nil {
+		log.Errorf("Failed to marshal feature status: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
```
LiZhenCheng9527, line 230: need `Contebt-Type: application/json`

pkg/status/status_server.go
```diff
@@ -218,7 +218,24 @@ func (s *Server) loggersHandler(w http.ResponseWriter, r *http.Request) {
...
+	_, _ = w.Write(data)
+}
+
 func (s *Server) accesslogHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodGet {
+		writeFeatureStatus(w, s.xdsClient.WorkloadController.GetAccesslogTrigger())
```
LiZhenCheng9527, line 236: This must be restricted to use in dual-engine mode. This is because kernel-Native mode does not initialize the workloadController, which would cause a panic.

Issue #1879, opened 2026-08-06
pkg/kube: NewPortForwarder ignores provided namespace and localAddress


yash-pakhale-07:
**What happened**:

`pkg/kube.NewPortForwarder()` accepts `ns` and `localAddress` parameters, but `newPortForwarder()` initializes the Cobra `namespace` and `address` flags using hardcoded defaults (`kmesh-system` and `localhost`) instead of the values provided by the caller.

As a result, the port-forward command ignores the supplied namespace and bind address during initialization, even though those values are stored in the `portForwarder` struct.

**What you expected to happen**:

`newPortForwarder()` should initialize the Cobra `namespace` and `address` flags using the `ns` and `localAddress` arguments passed to `NewPortForwarder()`. This would preserve the current behavior for existing callers while making the implementation consistent with the API contract.

**How to reproduce it (as minimally and precisely as possible)**:

1. Call `NewPortForwarder()` with a custom namespace and/or a custom `localAddress`.
2. Inspect the initialized Cobra command flags.
3. Observe that:
   - The `namespace` flag is initialized to `kmesh-system` instead of the supplied namespace.
   - The `address` flag is initialized to `localhost` instead of the supplied bind address.

**Anything else we need to know?**:

The issue appears to be in `pkg/kube/client.go`, where the Cobra flags are initialized with hardcoded values:

```go
cmd.Flags().String("namespace", KmeshNamespace, "Specify the namespace to use")
cmd.Flags().StringSlice("address", []string{DefaultLocalAddress}, "Specify the addresses to bind")

Samyra312007:
I have submitted the PR which fixes #1879 by initializing the Cobra `namespace` and `address` flags from the `ns` and `localAddress` arguments passed to `NewPortForwarder()` instead of the hardcoded defaults.

Key changes:
- `pkg/kube/client.go`: `newPortForwarder()` now sets the `namespace` flag from `ns` and the `address` flag from `localAddress`.
- `pkg/kube/portforwarder_test.go`: new table-driven unit tests covering custom ns/address, localhost fallback, dynamic port selection, and struct field population.

Existing callers are unaffected (they pass `""`/`kmesh-system`, producing the same flag values). Verified with `go build`, `go vet`, and `go test ./pkg/kube/...`.

PR #1883, opened 2026-08-06
fix(ctl): support custom Kmesh namespace

### LINKED ISSUE #1882: kmeshctl assumes Kmesh is installed in kmesh-system
**What happened**:

The Helm chart installs Kmesh into `.Release.Namespace`, but `kmeshctl` still uses `kmesh-system` for daemon discovery, port-forwarding and Kmesh secrets.

Commands such as `version`, `dump`, `log`, `authz`, `monitoring` and `secret` therefore fail or return empty results when Kmesh is installed in another namespace.

`kmeshctl` should accept the Kmesh installation namespace, probably through a global `--kmesh-namespace` flag, while keeping `kmesh-system` as the default.

```sh
helm install kmesh ./deploy/charts/kmesh-helm \
  --namespace kmesh-test \
  --create-namespace
```

kubectl get pods -n kmesh-test
kmeshctl version

The daemon is running in kmesh-test, but kmeshctl looks for it in kmesh-system. Commands that port-forward to a named daemon pod fail for the same reason.

**Anything else we need to know?:**

This also matters for the MCP work in #1800. Its daemon discovery and diagnostic tools accept a namespace, so the shared Kmesh client path should not assume kmesh-system.

Related to #1879, but that only covers NewPortForwarder ignoring a namespace already passed by its caller. The CLI currently has no way to provide that namespace, and pod/secret operations remain hardcoded.

**Environment:**
Kmesh version: current main
Kmesh mode: both
Istio version: not relevant
Kernel version: not relevant

Others: Helm install using a namespace other than kmesh-system

avinxshKD:
**What type of PR is this?**

/kind bug

Adds a global `--kmesh-namespace` flag to `kmeshctl`, with `kmesh-system` kept as the default.

The selected namespace is used for daemon discovery, secret operations and daemon port-forwarding. This allows `kmeshctl` to work with Helm installations deployed outside `kmesh-system`.

Fixes #1882

The existing waypoint `--namespace` flag is unchanged. It selects the workload namespace, while `--kmesh-namespace` selects the namespace where Kmesh itself is installed.

This includes the namespace part of #1879. The `localAddress` issue remains separate.

```release-note
kmeshctl now supports a global --kmesh-namespace flag for Kmesh installations outside kmesh-system.

### DIFF
--- a/ctl/common/common.go
+++ b/ctl/common/common.go
@@ -24,6 +24,7 @@ import (
 	logcmd "kmesh.net/kmesh/ctl/log"
 	"kmesh.net/kmesh/ctl/monitoring"
 	"kmesh.net/kmesh/ctl/secret"
+	"kmesh.net/kmesh/ctl/utils"
 	"kmesh.net/kmesh/ctl/version"
 	"kmesh.net/kmesh/ctl/waypoint"
 )
@@ -37,6 +38,7 @@ func GetRootCommand() *cobra.Command {
 			DisableDefaultCmd: true,
 		},
 	}
+	rootCmd.PersistentFlags().StringVar(&utils.KmeshNamespace, "kmesh-namespace", utils.KmeshNamespace, "Namespace where Kmesh is installed")
 
 	rootCmd.AddCommand(logcmd.NewCmd())
 	rootCmd.AddCommand(dump.NewCmd())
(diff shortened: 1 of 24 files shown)


avinxshKD:
Hi @LiZhenCheng9527 @yashisrani PTAL when you have time.

yashisrani:
@avinxshKD E2E test is failing .. so can you fix it ?

avinxshKD:
@yashisrani done fixed :p

cc @LiZhenCheng9527 @YaoZengzeng

avinxshKD:
@LiZhenCheng9527 ptal when get chance

PR #1885, opened 2026-08-06
fix(authz): reduce verifier state explosion in policy_check

Jashk120:
**What type of PR is this?**
/kind bug

The BPF verifier hits state explosion when loading `policy_check` with the authorization matching call chain fully inlined:

```
processed 1000001 insns (limit 1000000)
total_states 65547
peak_states 5524
```

Marking all authz-matching functions `__noinline` avoids the instruction limit but exceeds the verifier's call-stack depth limit:
```
the call stack of 9 frames is too deep
```

This change selectively marks only the larger verifier-heavy functions as `__noinline`, keeping smaller functions inlined. This stays within both limits:
```
processed 1615 insns (limit 1000000)
total_states 149
peak_states 149
```

Fixes: N/A

- Verified that `policy_check` loads successfully after this change.
- Verified the Kmesh pod starts successfully without BPF verifier errors.
- Ran `make gen-check` successfully.
- AI assistance was used during investigation and implementation. All generated changes were reviewed and validated before submission.

```release-note
NONE
```

### DIFF
--- a/bpf/kmesh/workload/include/authz.h
+++ b/bpf/kmesh/workload/include/authz.h
@@ -144,7 +144,8 @@ static int construct_tuple_key(struct xdp_md *ctx, struct bpf_sock_tuple *tuple_
     return PARSER_SUCC;
 }
 
-static int match_dst_ports(Istio__Security__Match *match, struct xdp_info *info, struct bpf_sock_tuple *tuple_info)
+static __noinline int
+match_dst_ports(Istio__Security__Match *match, struct xdp_info *info, struct bpf_sock_tuple *tuple_info)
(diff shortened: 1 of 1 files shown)


Jashk120:
make gen-check passes locally, but the GitHub Actions gen-check job rewrites three header files with whitespace-only formatting changes and then fails because the working tree is no longer clean.

The only changes are formatting like:
```
-((n) & 0xff)
+((n)&0xff)
```
My local clang-format version is:
```
clang-format version 22.1.6
```
Is the project expected to use a specific clang-format version or formatting configuration for gen-check?

PR #1141, opened 2024-12-24
Fix test maglev test suite

hzxuzhonghu:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes https://github.com/kmesh-net/kmesh/issues/1137

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

Issue #1137, opened 2024-12-23
TestMaglevTestSuite influence other test


hzxuzhonghu:
**What happened**:

Seen from https://github.com/kmesh-net/kmesh/pull/1132

If TestMaglevTestSuite  enabled, then `TestListenerLookup`  and `TestMaglevTestSuite ` both failed. 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Kmesh version:
- Kmesh mode(kmesh has `Kernel-Native Mode` and `Duel-Engine Mode`):
- Istio version:
- Kernel version:
- Others:

hzxuzhonghu:
Seems this is related to pin map in the userspace. But not cleanup gracefully

PR #1887, opened 2026-08-06
fix(consistenthash): honor requested tableSize in maglev getLookupTable

Atishyy27:
getLookupTable(cluster, tableSize) sized its output slice with the tableSize argument but computed every offset/skip/permutation with the package-level maglevTableSize global instead. When the global is unset (0) this panics with an integer divide-by-zero in getOffsetAndSkip; when it merely differs from the requested size it panics with an index-out-of-range write in the lookup loop. Pass tableSize through to getOffsetAndSkip and getPermutation so the hash space always matches the table being built.

**What type of PR is this?**

/kind bug

`getLookupTable(cluster, tableSize)` sizes its output slice with the `tableSize` argument but computed every offset/skip/permutation with the package-level `maglevTableSize` global instead. When the global is unset (0) this panics with an integer divide-by-zero in `getOffsetAndSkip`; when it merely differs from the requested size it panics with an index-out-of-range write in the lookup loop. This threads `tableSize` through `getOffsetAndSkip` and `getPermutation` so the hash space always matches the table being built.

Added `TestGetLookupTable_RespectsRequestedTableSize` covering the pre-init / divergent-size case; the existing `TestGetLookupTable` still passes with an unchanged backend distribution, so the normal path is unaffected.

NONE (no tracking issue filed — happy to open one if preferred)

#1141 refactors this same file for a different issue (#1137, test isolation) and incidentally touches this inconsistency, but it is a stale unmerged rewrite. This is a minimal standalone fix, not a duplicate of that scope.

release-note
NONE

### DIFF
--- a/pkg/consistenthash/maglev/maglev.go
+++ b/pkg/consistenthash/maglev/maglev.go
@@ -155,8 +155,8 @@ func getOffsetAndSkip(address string, m uint64) (uint64, uint64) {
 	return offset, skip
 }
 
-func getPermutation(b Backend) uint64 {
-	return (b.offset + (b.skip * b.next)) % maglevTableSize
+func getPermutation(b Backend, tableSize uint64) uint64 {
+	return (b.offset + (b.skip * b.next)) % tableSize
 }
 
 func getLookupTable(cluster *cluster_v2.Cluster, tableSize uint64) ([]int, error) {
@@ -178,7 +178,7 @@ func getLookupTable(cluster *cluster_v2.Cluster, tableSize uint64) ([]int, error
 	backends := make([]Backend, 0, len(flatEps))
 
 	for i, ep := range flatEps {
-		epOffset, epSkip := getOffsetAndSkip(ep.GetAddress().String(), maglevTableSize)
+		epOffset, epSkip := getOffsetAndSkip(ep.GetAddress().String(), tableSize)
 		b := Backend{
 			ep:     ep,
 			index:  i,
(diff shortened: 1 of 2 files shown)


pkg/consistenthash/maglev/maglev.go
```diff
@@ -178,7 +178,7 @@ func getLookupTable(cluster *cluster_v2.Cluster, tableSize uint64) ([]int, error
 	backends := make([]Backend, 0, len(flatEps))
 
 	for i, ep := range flatEps {
-		epOffset, epSkip := getOffsetAndSkip(ep.GetAddress().String(), maglevTableSize)
+		epOffset, epSkip := getOffsetAndSkip(ep.GetAddress().String(), tableSize)
```
LiZhenCheng9527, line 185: If the `tableSize` is not validated, a panic caused by a division by zero may occur in the `getOffsetAndSkip` function.

LiZhenCheng9527 (commented):
General LGTM

pkg/consistenthash/maglev/maglev.go
Atishyy27: good catch, thanks. you're right
threading `tableSize` through moved the divisor but didn't actually guard 0. added an early check in `getLookupTable` that returns an error for a zero table size before it reaches `getOffsetAndSkip`, plus a test covering it. pushed.

LiZhenCheng9527 (approved):
/lgtm

PR #1902, opened 2026-08-08
fix: default ENHANCED_KERNEL to avoid empty -tags in go build

### LINKED ISSUE #1884: ENHANCED_KERNEL variable is empty by default, causing `go build -tags` to misfire
**What happened**:
When running `make all-binary` (or `make build`) to build `kmesh-daemon`,
the build fails with:

    named files must be .go files: kmesh-daemon
    make: *** [Makefile:110: all-binary] Error 1

This happens because the `ENHANCED_KERNEL` variable used in the Makefile:

    $(GO) build -ldflags $(LDFLAGS) -tags $(ENHANCED_KERNEL) -o $(APPS1) $(GOFLAGS) ./daemon/main.go

resolves to an **empty string** in my environment. When `ENHANCED_KERNEL` is
empty, the command effectively becomes:

    go build -ldflags "..." -tags -o kmesh-daemon -a ./daemon/main.go

Since `-tags` requires a value, it consumes the next token (`-o`) as its
argument instead, and `kmesh-daemon` is then treated as a positional file
argument, which Go rejects with the error above.

Per the docs (https://kmesh.net/docs/setup/use-enhanced-kernel/), Kmesh
determines whether the enhanced kernel is used via:

    grep -q "FN(parse_header_msg)" /usr/include/linux/bpf.h && echo "enhanced" || echo "unenhanced"

I'd expect `ENHANCED_KERNEL` to always resolve to either `enhanced` or
`unenhanced` (never empty), regardless of whether the host kernel actually
supports the enhanced features.

**How to reproduce it**:
1. On a standard (non-openEuler / non-enhanced-kernel) Linux host, e.g.
   Ubuntu, without the enhanced kernel headers installed.
2. Run `make all-binary` or `make build`.
3. Build fails with `named files must be .go files: kmesh-daemon`.

**Workaround**:
Manually setting the variable avoids the issue:

    ENHANCED_KERNEL=unenhanced make all-binary

This looks like the variable is expected to be set by some environment
setup step (possibly in `kmesh_compile_env_pre.sh` or a similar prepare
script), which either wasn't run in my case or doesn't set a fallback
default. A safe default (e.g. `ENHANCED_KERNEL ?= unenhanced` in
`mk/bpf.vars.mk`) combined with the actual detection logic would prevent
this from silently breaking `go build`.

**Environment**:
- OS: Ubuntu 24.04.4 LTS
- Kernel: 7.0.0-28-generic
- Go version: go1.24.2 linux/amd64

mohammd-khalil:
**What type of PR is this?**
/kind bug

When `ENHANCED_KERNEL` hasn't been set (e.g. `set_enhanced_kernel_env` in
`kmesh_compile_env_pre.sh` wasn't sourced, such as when running
`make all-binary` directly outside the official `make build` / Docker
flow), the variable resolves to an empty string.

This breaks the build:

    go build -ldflags $(LDFLAGS) -tags $(ENHANCED_KERNEL) -o $(APPS1) ...

becomes effectively:

    go build -ldflags "..." -tags -o kmesh-daemon -a ./daemon/main.go

Since `-tags` requires a value, it consumes the next token (`-o`) as its
argument, and `kmesh-daemon` becomes a stray positional argument, causing:

    named files must be .go files: kmesh-daemon

This PR adds a safe fallback default in `mk/bpf.vars.mk`:

    ENHANCED_KERNEL ?= normal

This matches the same default `set_enhanced_kernel_env` already uses when
enhanced kernel headers aren't detected, so it doesn't change behavior for
the official build flow — it only prevents the empty-string case from
silently breaking `go build`.

Fixes #1884

Verified locally:
- Before the fix: `make -n all-binary` shows `-tags -o kmesh-daemon` (empty tag value)
- After the fix: `make -n all-binary` shows `-tags normal` as expected
- Explicitly setting `ENHANCED_KERNEL=enhanced` still overrides correctly
  (the `?=` default doesn't clobber an already-set value)

NONE

```release-note
NONE
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,7 +117,7 @@ git commit -s -m "your commit message"
 
 This automatically appends:
 
-```
+```text
 Signed-off-by: Your Name <your-email@example.com>
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,8 @@ Welcome to Kmesh!
   - [AI Guidance](#ai-guidance)
   - [Contributor Workflow](#contributor-workflow)
     - [Creating Pull Requests](#creating-pull-requests)
+    - [Developer Certificate of Origin (DCO)]
+    - [Fixing a missing sign-off] 
     - [Code Review](#code-review)
   - [Membership](#membership)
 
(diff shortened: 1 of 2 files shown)


LiZhenCheng9527 (approved):
/lgtm

PR #1905, opened 2026-08-08
fix(utils): make HashName ring probe reach the final slot

### LINKED ISSUE #1904: HashName.Hash never reaches the final ring slot (math.MaxUint32); wraparound is dead code
while reading pkg/utils/hash_name.go i noticed the linear-probing loop in Hash() can never reach its last slot.

    for num = h.hash.Sum32(); num < math.MaxUint32; num++ {
        if _, exists := h.numToStr[num]; !exists { ...register...; break }
        // It's a ring
        if num == math.MaxUint32 { num = 0 }
    }

the condition `num < math.MaxUint32` is checked before the body every iteration, so the body never runs for num == math.MaxUint32. two consequences:

1. the `if num == math.MaxUint32 { num = 0 }` ring-wraparound is unreachable dead code (to enter the body num must already be < MaxUint32).
2. if a probe lands on slot math.MaxUint32 while it's free, Hash returns math.MaxUint32 but never writes numToStr/strToNum for it — so NumToStr(MaxUint32) returns "" and StrToNum(that string) returns 0, i.e. the round trip silently breaks.

Hash() derives the BPF-map keys for workload/service/policy ids, so an unregistered id there is a real (if rare, ~1-in-2^32 per name) correctness gap. num is a uint32, so num++ already wraps MaxUint32->0 — dropping the bound makes the probe a true ring and removes the dead check. happy to send a small PR (drop the bound + remove the dead wraparound + a regression test that forces the probe onto the top slot).

Atishyy27:
**What type of PR is this?**

/kind bug

HashName.Hash used `for num = h.hash.Sum32(); num < math.MaxUint32; num++`, so the loop exited before ever testing slot math.MaxUint32 and the `if num == math.MaxUint32 { num = 0 }` wraparound was unreachable dead code. When linear probing lands on the top slot while it's free, Hash returned math.MaxUint32 without recording it in numToStr/strToNum, so NumToStr and StrToNum disagree for that id. Since Hash derives the BPF-map keys for workload/service/policy ids, that's a real (if rare) correctness gap.

num is a uint32, so num++ already wraps math.MaxUint32 back to 0, so dropping the bound turns the probe sequence into a true ring and removes the now-dead wraparound. Added a regression test that fakes the hash to force the probe onto the top slot and asserts the id round-trips.

Fixes #1904

```release-note
NONE
```

### DIFF
--- a/pkg/utils/hash_name.go
+++ b/pkg/utils/hash_name.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"hash"
 	"hash/fnv"
-	"math"
 	"os"
 
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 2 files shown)


LiZhenCheng9527 (approved):
/lgtm

PR #1907, opened 2026-08-08
fix(waypoint): treat an empty-valued namespace label as present

### LINKED ISSUE #1906: kmeshctl waypoint apply: an empty-valued use-waypoint label is treated as absent
while reading ctl/waypoint i noticed namespaceHasLabel checks the label's value, not its presence.

in namespaceHasLabel (ctl/waypoint/waypoint.go:597-606):

    return nsObj.Labels[label] != "", nil

an empty-valued label is still a present label in kubernetes (e.g. istio.io/use-waypoint: "" is valid), but this returns false for it. that result feeds the overwrite guard in `kmeshctl waypoint apply --enroll-namespace` (waypoint.go:223):

    if !overwrite && hasWaypoint {
        // warn and return, don't overwrite an existing enrollment
    }

so a namespace already carrying use-waypoint="" comes back as hasWaypoint=false, the guard is skipped, and the existing enrollment is silently overwritten without the intended warning.

the sibling namespaceHasLabelWithValue is fine (it compares a specific value). only the presence check is wrong. happy to send a small PR that checks for the key instead of a non-empty value.

Atishyy27:
**What type of PR is this?**

/kind bug

namespaceHasLabel (ctl/waypoint/waypoint.go) checked Labels[label] against a non-empty value, so a namespace carrying the label with an empty value (istio.io/use-waypoint set to the empty string, a valid kubernetes label) was reported as not having the label. in `kmeshctl waypoint apply --enroll-namespace` this makes the `!overwrite && hasWaypoint` guard fall through and silently overwrite an existing enrollment without the intended warning. this checks for the key's presence instead of a non-empty value. added a regression test.

Fixes #1906

**Special notes for your reviewer:**

the test uses a minimal fake CLIClient (only Kube() is wired, which is the only method namespaceHasLabel calls) with a fake clientset seeding a namespace labeled use-waypoint="".

```release-note
NONE
```

### DIFF
--- a/ctl/waypoint/waypoint.go
+++ b/ctl/waypoint/waypoint.go
@@ -605,7 +605,10 @@ func namespaceHasLabel(kubeClient kube.CLIClient, ns string, label string) (bool
 	if nsObj.Labels == nil {
 		return false, nil
 	}
-	return nsObj.Labels[label] != "", nil
+	// A label present with an empty value is still present; check for the key,
+	// not a non-empty value (istio.io/use-waypoint: "" is a valid, present label).
+	_, ok := nsObj.Labels[label]
+	return ok, nil
 }
 
 func namespaceHasLabelWithValue(kubeClient kube.CLIClient, ns string, label, labelValue string) (bool, error) {
--- a/ctl/waypoint/waypoint_label_test.go
+++ b/ctl/waypoint/waypoint_label_test.go
@@ -1,0 +1,67 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


yashisrani:
@Atishyy27 E2E tests are failing .. can you fix it ?

Atishyy27:
@yashisrani looked into it. the failure is `TestMixNsAndServiceWaypoint` in the ipv6 lane, and i don't think this PR can reach it.

namespaceHasLabel, the only function this PR changes, has one caller: the --enroll-namespace branch of `kmeshctl waypoint apply` (ctl/waypoint/waypoint.go:218). the e2e suite never goes through it. `TestMixNsAndServiceWaypoint` sets the label with a direct API patch in SetWaypoint, and enrollNamespaceOrFail in manage_test.go patches the namespace label directly too, neither one shells out to kmeshctl with that flag. newWaypointProxyOrFail does call kmeshctl waypoint apply but without --enroll-namespace, so it stops short of the changed line.

the same commit passed 2 of the 3 ipv6 runs and every regular e2e run, and the three failing subtests all timed out at exactly 60s waiting on traffic, which reads like the waypoint pod not being ready rather than a label check.

happy to be wrong here. could you retrigger the job? if it fails again on the same test i'll dig further.

PR #1909, opened 2026-08-09
fix: CompareIpByte skips invalid IP bytes instead of using zero-value IP

### LINKED ISSUE #1908: CompareIpByte silently continues after invalid IP bytes, causing wrong service address diffs
CompareIpByte in pkg/nets/nets.go logs an error when netip.AddrFromSlice fails but doesn't stop:

go
ip, ok := netip.AddrFromSlice(item)
if !ok {
    log.Error("cannot compared IP: Unsupported data types")
}
aSet[ip.String()] = item

On failure ip is zero-value, ip.String() returns a fixed "invalid IP" string regardless of input. Multiple malformed items in a collide under the same key in aSet - only the last survives. Same bug in the second loop over b.

Used in workload_processor.go:824 to diff service addresses on update. A wrong diff here means a stale address never gets removed from the BPF map, or a valid one gets incorrectly removed.

Fix: continue after logging, in both loops, instead of falling through with a zero-value IP.

bhumikadangayach:
Fixes #1908

CompareIpByte logged an error on invalid IP bytes but didn't stop - the zero-value IP's String() ("invalid IP") was used as a map key anyway, so multiple malformed entries would collide under the same key, and a malformed entry in one side could incorrectly match against one left over from the other side.

Added continue after the error log in both loops, so malformed entries are skipped entirely instead of participating in the comparison.

Added a test case covering malformed entries mixed with valid ones on both sides, verifying the diff result only reflects genuine IP comparisons.

### DIFF
--- a/pkg/nets/nets.go
+++ b/pkg/nets/nets.go
@@ -135,16 +135,16 @@ func CompareIpByte(a, b [][]byte) [][]byte {
 		ip, ok := netip.AddrFromSlice(item)
 		if !ok {
 			log.Error("cannot compared IP: Unsupported data types")
+			continue
 		}
-
 		aSet[ip.String()] = item
 	}
-
 	var bMissing [][]byte
 	for _, item := range b {
 		ip, ok := netip.AddrFromSlice(item)
 		if !ok {
 			log.Error("cannot compared IP: Unsupported data types")
+			continue
 		}
 		if _, ok := aSet[ip.String()]; !ok {
 			bMissing = append(bMissing, item)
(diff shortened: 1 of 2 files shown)


yashisrani:
There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.

To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Here is how to fix the problem so that this code can be merge

bhumikadangayach:
E2E istio 1.26 failed on the log-collection step after all tests passed - PASS on TestKmeshRestart etc, but exits 1 because of cat: kmesh_daemon.log: No such file or directory afterward. Same pattern on #1845 and #1881, both resolved on re-run. Filed separately as #1911. Could you trigger a re-run?

PR #1910, opened 2026-08-09
Enhancement/add duplicate prog test

### LINKED ISSUE #799: Add restart test

Currently restart is an alpha feature, and we have met many issues like bpf map elements duplicate and bpf prog duplicated.

So I would suggest we add some tests to cover this.

In ut test, we can check no same named prog and map duplicated.   cc @lec-bit 

In e2e test, we can also check that after kmesh restart.  cc @YaoZengzeng

R4R3D1FF:
**What type of PR is this?**

/kind enhancement

Adds tests for kmesh restart ensuring no duplicate program is present.

Fixes https://github.com/kmesh-net/kmesh/issues/799

```release-note
NONE
```


yashisrani:
@R4R3D1FF ci are failing, so please fix it ..

R4R3D1FF:
> @R4R3D1FF ci are failing, so please fix it ..

Should work now @yashisrani

pkg/bpf/utils/bpf_helper_test.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+		License: "GPL",
+	})
+
+	if err != nil {
+		t.Fatalf("BpfProgUpdate failed: %v", err)
+	}
+
+	cgopt := link.CgroupOptions{
+		Path:    "/sys/fs/cgroup",
```
LiZhenCheng9527: Shouldn't it be cgroup2?

R4R3D1FF: I don't think it matters for the focus of this test since it's only supposed to test whether BpfProgUpdate properly replaces the program so that it is removed after no link points to it.

Should I still replace it anyway? (even though it has no actual affect on the functioning of the test)

PR #1912, opened 2026-08-11
fix(test): resolve log collection failure in E2E tests (#1911)

### LINKED ISSUE #1911: E2E test job fails on log collection step even when all tests pass
E2E test jobs intermittently fail with exit code 1 even though every actual test in the run passes:

--- PASS: TestKmeshRestart (20.09s)
FAIL
...
FAIL	kmesh.net/kmesh/test/e2e	433.008s
FAIL
cat: kmesh_daemon.log: No such file or directory
make: *** [Makefile:216: e2e] Error 1
Error: Process completed with exit code 2.

The cat kmesh_daemon.log step after the test suite runs seems to expect a log file at a path that doesn't reliably exist by the time it runs, causing a false failure. Seen this exact pattern across multiple unrelated PRs (#1845, #1881, #1909), always on one Istio version while others pass on identical code, and it self-resolves on a CI re-run.

Likely in the e2e Makefile target (Makefile:216) or the workflow step collecting logs - the log path/timing assumption seems off, causing a spurious failure unrelated to test correctness.

dwan-ith:
**What type of PR is this?**

This is an E2E test to fix issue #1911
/kind bug

The actual issue was that the run_test.sh script (invoked via Makefile) was performing cat kmesh_daemon.log after test suites, causing a spurious exit failure on jobs where logs weren't being exported yet or tests exited before outputting into it successfully

In the current contents of test/e2e/run_test.sh, logarithmic checks are now put behind a conditional, eliminating the FAIL cat: kmesh_daemon.log: No such file or directory error that caused CI failures earlier

Fixes #1911

```release-note

NONE

### DIFF
--- a/test/e2e/run_test.sh
+++ b/test/e2e/run_test.sh
@@ -440,11 +440,19 @@ bash -c "$cmd"
 EXIT_CODE=$?
 set -e
 
-if [ $EXIT_CODE -ne 0 ]; then
-	cat $LOGFILE
+# Log collection is diagnostic and must not replace the test exit status.
+if [ "$EXIT_CODE" -ne 0 ]; then
+	echo "E2E tests failed with exit code $EXIT_CODE."
+	if [[ -r $LOGFILE ]]; then
+		cat "$LOGFILE" || echo "Failed to read Kmesh daemon log: $LOGFILE"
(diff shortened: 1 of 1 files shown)


LiZhenCheng9527 (approved):
/lgtm

PR #1913, opened 2026-08-11
refactor(ctl): replace os.Exit with proper error handling via RunE

devGPP23:
**Problem**

I noticed any network error triggers a hard `os.Exit(1)` deep inside the Kmesh codebase. This abruptly crashes the entire MCP server process instead of gracefully returning the error to the caller.
Before this fix, an error fetching pods would print a raw logger message and forcefully terminate the CLI process without passing the error back to the Cobra framework:

```
$ ./kmeshctl authz status
client version: v0.0.0-master
time="2026-08-11T12:42:41Z" level=error msg="failed to get kmesh podList: Get \"https://127.0.0.1:34255/api/v1/namespaces/kmesh-system/pods?labelSelector=app%3Dkmesh\": dial tcp 127.0.0.1:34255: connect: connection refused"
```
**Solution**

To ensure `kmeshctl` is completely safe to embed in long-running processes, I audited the remaining subcommands (`authz`, `version`, `secret`, `monitoring`) and performed a full migration.
Specifically, I did the following:
1)Migrated the `Run` field to `RunE` in all subcommands so Cobra can natively handle returned errors.
2)Systematically removed all `os.Exit(1)` calls.
3)Refactored helper functions to return standard Go error objects rather than logging and crashing the process.
4)Replaced the hard dependency on standard `stdout/stderr` by using `io.Writer` and `cmd.OutOrStdout()` where applicable.

**How it Improves**
After my changes, errors are gracefully propagated up through `fmt.Errorf` rather than forcefully killing the application.we can safely import these CLI functions without the risk of an unexpected fatal exit.

In the CLI itself, Cobra now successfully catches the returned error and formats it cleanly with an Error: prefix instead of abruptly terminating:
Now Terminal Output is like below with Error: prefic
```
$ ./kmeshctl authz status
time="2026-08-11T12:42:41Z" level=warning msg="failed to create log directory: mkdir /var/run/kmesh/: permission denied, consider running with root user"
Error: failed to get kmesh podList: Get "https://127.0.0.1:34255/api/v1/namespaces/kmesh-system/pods?labelSelector=app%3Dkmesh": dial tcp 127.0.0.1:34255: connect: connection refused
```
This makes `kmeshctl` much safer, more idiomatic, and fully embeddable!

### DIFF
--- a/ctl/dump/dump.go
+++ b/ctl/dump/dump.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"os"
 	"strings"
 	"text/tabwriter"
 
@@ -57,8 +56,8 @@ kmeshctl dump <kmesh-daemon-pod> dual-engine
 # Output as raw JSON:
 kmeshctl dump <kmesh-daemon-pod> kernel-native -o json`,
 		Args: cobra.ExactArgs(2),
-		Run: func(cmd *cobra.Command, args []string) {
-			_ = RunDump(cmd, args, outputFormat)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return RunDump(cmd, args, outputFormat)
 		},
 	}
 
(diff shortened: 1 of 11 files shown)


ctl/dump/dump.go
```diff
@@ -70,65 +69,61 @@ func RunDump(cmd *cobra.Command, args []string, outputFormat string) error {
...
 	}
 
 	cli, err := utils.CreateKubeClient()
 	if err != nil {
-		log.Errorf("failed to create cli client: %v", err)
-		os.Exit(1)
+		return fmt.Errorf("failed to create cli client: %v", err)
 	}
 
 	fw, err := utils.CreateKmeshPortForwarder(cli, podName)
```
LiZhenCheng9527, line 80: Missing fw.Close()

devGPP23:
Hi @LiZhenCheng9527
 I've addressed the feedback:

1 ) Added defer fw.Close() across all ctl/ port-forwarder helpers to prevent any connection/goroutine leaks.
2) Refactored remaining os.Exit(1) calls to bubble errors through RunE.
3)All tests and CI checks are green. 
please take a look whenever u are free

PR #1928, opened 2026-08-15
refactor(constants): share the daemon admin API paths

sathwikshetty33:
While exploring the admin APIs , I noticed that the daemon's admin API paths are declared in `pkg/status`, and then declared again by every client that talks to it. I guess the reason for this is that `pkg/status` needs the generated eBPF objects to build, so no client can import it. so each one keeps its own copy instead. This kind of duplication can cause bugs in the repo I did raise #1854  fixing a similar kind of bug. 

This PR moves the paths to `pkg/constants` and has both sides use them. `pkg/constants` has no eBPF dependencies and is already imported by `pkg/status` and by `ctl/dump`, so no new dependency edges are introduced. It aslo updates the call sites and adds a test to check the value of these constants.

No behaviour change.

**What type of PR is this?**

Fixes #

```release-note
No
```

### DIFF
--- a/ctl/authz/authz.go
+++ b/ctl/authz/authz.go
@@ -28,12 +28,9 @@ import (
 	"github.com/spf13/cobra"
 
 	"kmesh.net/kmesh/ctl/utils"
+	"kmesh.net/kmesh/pkg/constants"
 	"kmesh.net/kmesh/pkg/kube"
 	"kmesh.net/kmesh/pkg/logger"
-)
-
-const (
-	patternAuthz = "/authz"
 )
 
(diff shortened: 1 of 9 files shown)


sathwikshetty33:
cc/ @yashisrani could you please take a look a this whenever you find time. Thanks in advance!

LiZhenCheng9527 (commented):
I don't think this change makes much sense.

sathwikshetty33:
Thanks for the review on this one. I guess this is a premature change which isn't fixing anything right now but the intention was to prevent these paths from drifting in future. So I will close it.

Issue #1800, opened 2026-07-23
[Feature]: Develop MCP Server for AI-Native Kmesh Service Mesh Management


yashisrani:
### **Description:**

  Build a Model Context Protocol (MCP) server that exposes Kmesh's capabilities as callable tools for AI agents (Claude, Cursor, GitHub Copilot). The MCP server acts as a middle layer between AI tools and Kmesh infrastructure, enabling natural language queries like "What services are running?" or "Why is service A not reaching service B?" instead of requiring users to learn complex `kmeshctl` commands, xDS APIs, and eBPF internals.

### **Architecture:**

  ```mermaid
  graph TB
      subgraph "AI Clients (End User's LLM Cost)"
          Claude["Claude Desktop"]
          Cursor["Cursor IDE"]
          Copilot["GitHub Copilot"]
      end

      subgraph "Kmesh MCP Server"
          ToolRegistry["Tool Registry"]
          AuthHandler["Auth Handler"]
          SSETransport["SSE Transport"]

          ServiceDiscovery["Service Discovery Tools"]
          TrafficMgmt["Traffic Management Tools"]
          Observability["Observability Tools"]
          Security["Security Tools"]
          Diagnostics["Diagnostics Tools"]
      end

      subgraph "Kmesh Infrastructure"
          StatusServer["Status Server :15200"]
          KubeAPI["Kubernetes API"]
          eBPFMaps["eBPF Maps"]
          XDSController["xDS Controller"]
      end

      Claude & Cursor & Copilot -->|"MCP Protocol (HTTP+SSE)"| ToolRegistry
      ToolRegistry --> AuthHandler & SSETransport
      ToolRegistry --> ServiceDiscovery & TrafficMgmt & Observability & Security & Diagnostics

      ServiceDiscovery --> KubeAPI
      TrafficMgmt --> StatusServer & XDSController
      Observability --> StatusServer
      Security --> StatusServer & KubeAPI
      Diagnostics --> StatusServer & eBPFMaps
  ```

  **Data Flow:**

  ```mermaid
  sequenceDiagram
      participant User as User
      participant AI as AI Client
      participant MCP as Kmesh MCP Server
      participant Kmesh as Kmesh Infrastructure

      User->>AI: "What services are running?"
      AI->>MCP: MCP Tool Call: list_services()
      MCP->>MCP: Validate auth token
      MCP->>Kmesh: Query services + endpoints
      Kmesh-->>MCP: Service data
[...]

PR #1930, opened 2026-08-15
kmeshctl: escape logger name in the log query URL

### LINKED ISSUE #1929: kmeshctl log <pod> <logger> fails for logger names that are not URL safe
**What happened**:

`kmeshctl log <pod> <logger>` puts the logger name straight into the query string without escaping. A name containing a space builds a malformed request line, which the daemon's HTTP server rejects before routing, so the handler never runs and the user gets a bare `400 Bad Request` with nothing to act on.

```
$ kmeshctl log <pod> "cni installer"
failed to get logger level: received status code 400, Response body: 400 Bad Request
```

With the name escaped, the request reaches the handler and the daemon explains itself:

```
failed to get logger level: received status code 400, Response body: 	logger cni installer does not exist
```

**Root cause**:

`ctl/log/log.go`:

```go
url += fmt.Sprintf("?name=%s", args[1])
```

`args[1]` is arbitrary user input from the command line and is never escaped.

**Why this is worth fixing**:

The daemon registers exactly three loggers: `default`, `fileOnly` and `bpf`. But Kmesh tags every log line with a `subsys` field, and those values look like logger names:

```
time="..." level=error msg="..." subsys="cni installer"
time="..." level=error msg="..." subsys=cache/v2
time="..." level=info  msg="..." subsys=manage_controller
```

Someone reading daemon logs and trying `kmeshctl log <pod> "cni installer"` is already asking for something that does not exist. The daemon has a message that says exactly that. The unescaped URL is what stops them from seeing it.

Names with slashes (`cache/v2`, `kmeshctl/authz`) reach the handler fine today, since a slash is legal in a query value, but they are equally unescaped.

Any name passed on the command line should reach the daemon intact, so the user sees the daemon's own error rather than a transport-level 400.

1. Deploy kmesh (a kind cluster is enough)
2. `kmeshctl log <kmesh-daemon-pod> default` works and prints the level
3. `kmeshctl log <kmesh-daemon-pod> "cni installer"` fails with the bare 400 above, rather than `logger cni installer does not exist`

I found this while building a prototype MCP server for #1800 that wraps `/debug/loggers` as a tool. My own client had the same bug.

An earlier version of this issue claimed `cni installer` was a registered logger that `kmeshctl log <pod>` lists. That was wrong, and I corrected it after testing against a real daemon: `NewLoggerScope` attaches a `subsys` field to the default logger, it does not register a new one, so the list really is just `default`, `fileOnly`, `bpf`. The escaping bug itself reproduces exactly as described above, verified both ways against kmesh-daemon in a kind cluster.

`--set` is unaffected, since that path sends the name in the POST body rather than the query string.

**Environment**:

- Kmesh version: main branch (`ae15e8a7`)
- Kmesh mode(kmesh has `Kernel-Native Mode` and `Dual-Engine Mode`): both, this is in kmeshctl rather than the datapath
- Istio version: 1.24.2
- Kernel version: 6.12 (kind on Docker Desktop)
- Others: verified against a real kmesh-daemon, not only by reading source

Anand-240:
**What type of PR is this?**

/kind bug

The logger name comes straight from the command line and went into the query string unescaped. A name containing a space builds a malformed request line, which the daemon's HTTP server rejects before routing, so the handler never runs and the user gets a bare `400 Bad Request`.

Before:

```
$ kmeshctl log <pod> "cni installer"
failed to get logger level: received status code 400, Response body: 400 Bad Request
```

After:

```
$ kmeshctl log <pod> "cni installer"
failed to get logger level: received status code 400, Response body: 	logger cni installer does not exist
```

Both verified against a kmesh-daemon running in a kind cluster, using kmeshctl built from `upstream/main` and from this branch.

This is worth fixing because the daemon registers only `default`, `fileOnly` and `bpf`, while Kmesh tags every log line with a `subsys` field whose values (`cni installer`, `cache/v2`, `manage_controller`) read like logger names. Someone who tries one of those is already asking for something that doesn't exist, and the daemon has a message saying so. The unescaped URL is what stops them from seeing it.

This moves the URL building into `loggerLevelURL` and escapes the name there. The local variable was renamed from `url` to `loggersURL` since it shadowed the `net/url` package.

`--set` is unaffected, since that path sends the name in the POST body.

Fixes #1929

**Testing**:

Added `ctl/log/log_test.go`, which this package did not have. `TestLoggerLevelURL` covers the escaping directly and `TestLoggerLevelURLReachesHandler` asserts the name arrives at the handler intact.

Without the fix the second one fails with the same error the CLI shows:

```
--- FAIL: TestLoggerLevelURLReachesHandler/cni_installer
    log_test.go:78: GetJson for logger "cni installer": received status code 400, Response body: 400 Bad Request
```

I found this while building a prototype MCP server for #1800 that wraps `/debug/loggers` as a tool. My own client had the same bug.

Worth flagging: the first version of #1929 claimed `cni installer` was a registered logger that `kmeshctl log <pod>` lists. That was wrong. `NewLoggerScope` attaches a `subsys` field to the default logger rather than registering a new one, so the real list is just the three above. I caught it by standing up a kind cluster and testing against a real daemon, and I have corrected the issue. The escaping bug itself reproduces exactly as shown.

I used an AI coding assistant (Claude) to help investigate this and write the test. I reviewed and understand every line here and can answer questions about any of it in review.

```release-note
fix: `kmeshctl log <pod> <logger>` now escapes the logger name, so a name containing a space returns the daemon's error instead of a bare 400.
```

### DIFF
--- a/ctl/log/log.go
+++ b/ctl/log/log.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 
@@ -61,6 +62,20 @@ kmeshctl log <kmesh-daemon-pod> default`,
 	}
 	cmd.Flags().String("set", "", "Set the logger level (e.g., default:debug)")
 	return cmd
+}
+
+// loggerLevelURL builds the URL that queries a single logger's level.
+//
+// The name comes straight from the command line, so it has to be escaped. A
+// name containing a space builds a malformed request line that the daemon's
+// HTTP server rejects before routing, and the caller sees a bare
+// "400 Bad Request" instead of the daemon's own "logger <name> does not exist".
+// That matters because Kmesh logs tag every line with a subsys field, and those
+// values ("cni installer", "cache/v2") read like logger names even though the
+// daemon only registers default, fileOnly and bpf. Someone reading logs will
+// reasonably try one, and should get told what is actually wrong.
+func loggerLevelURL(base, name string) string {
+	return base + "?name=" + url.QueryEscape(name)
 }
 
 func GetJson(url string, val any) error {
@@ -176,17 +191,16 @@ func RunGetOrSetLoggerLevel(cmd *cobra.Command, args []string) {
 	}
 	defer fw.Close()
 
-	url := fmt.Sprintf("http://%s%s", fw.Address(), patternLoggers)
+	loggersURL := fmt.Sprintf("http://%s%s", fw.Address(), patternLoggers)
 
 	setFlag, _ := cmd.Flags().GetString("set")
 	if setFlag == "" {
 		if len(args) >= 2 {
-			url += fmt.Sprintf("?name=%s", args[1])
-			GetLoggerLevel(url)
+			GetLoggerLevel(loggerLevelURL(loggersURL, args[1]))
 		} else {
-			GetLoggerNames(url)
+			GetLoggerNames(loggersURL)
 		}
 	} else {
-		SetLoggerLevel(url, setFlag)
+		SetLoggerLevel(loggersURL, setFlag)
 	}
 }
--- a/ctl/log/log_test.go
+++ b/ctl/log/log_test.go
@@ -1,0 +1,95 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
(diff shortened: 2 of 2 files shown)


LiZhenCheng9527 (approved):
/lgtm

PR #1935, opened 2026-08-17
fix(netns): actually insert into netnsObserved during proc scan

magic-peach:
/kind bug

Noticed this while looking at the netns scan for pod->namespace matching. processEntry takes a netnsObserved set and checks it before doing the cgroup read, but nothing ever inserts an inode into it. So the "already processed" skip never fires - two proc entries pointing at the same netns (e.g. two threads of the same process, which is the normal case) both get fully processed instead of the second one being skipped.

Not a correctness bug for FindNetnsForPod itself since it returns on first match anyway, but it does mean we redo the inode lookup + cgroup read for every entry that happens to share a netns with one we already looked at, for no reason.

Swapped it for InsertContains, which checks and inserts in one call instead of two.

Test hard-links two proc dirs' ns/net files to the same inode and confirms the second processEntry call gets skipped once this is fixed (it wasn't, before).

```release-note
NONE
```

### DIFF
--- a/pkg/controller/netns/netns.go
+++ b/pkg/controller/netns/netns.go
@@ -112,7 +112,7 @@ func processEntry(proc fs.FS, netnsObserved sets.Set[uint64], filter types.UID, 
 	if err != nil {
 		return "", err
 	}
-	if _, ok := netnsObserved[inode]; ok {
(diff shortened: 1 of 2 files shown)


LiZhenCheng9527 (approved):
/lgtm

PR #1937, opened 2026-08-18
fix: initialize port-forward namespace and address flags from provide…

### LINKED ISSUE #1879: pkg/kube: NewPortForwarder ignores provided namespace and localAddress

Samyra312007:
**What type of PR is this?**

/kind bug

Fixes `pkg/kube.NewPortForwarder()` so that `newPortForwarder()` initializes the Cobra `namespace` and `address` flags from the `ns` and `localAddress` arguments passed by the caller instead of the hardcoded `KmeshNamespace` (`kmesh-system`) and `DefaultLocalAddress` (`localhost`) defaults.

Previously the supplied namespace and bind address were stored in the `portForwarder` struct but never reflected in the Cobra flags, so the port-forward command bound to `localhost` regardless of the requested `localAddress` and exposed an inconsistent `namespace` flag value. Existing callers pass `""`/`kmesh-system`, so behavior is preserved for them.

Also adds unit tests in `pkg/kube/portforwarder_test.go` covering flag initialization, the localhost fallback, dynamic port selection, and struct field population.

Fixes #1879

NONE

```release-note
NONE

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -95,8 +95,8 @@ func newPortForwarder(cliClient *client, podName string, ns string, localAddress
 
 	cmd := &cobra.Command{}
 	cmd.Flags().Duration("pod-running-timeout", DefaultPodRunningTimeout, "Timeout for waiting for pod to be running")
-	cmd.Flags().String("namespace", KmeshNamespace, "Specify the namespace to use")
-	cmd.Flags().StringSlice("address", []string{DefaultLocalAddress}, "Specify the addresses to bind")
+	cmd.Flags().String("namespace", ns, "Specify the namespace to use")
+	cmd.Flags().StringSlice("address", []string{localAddress}, "Specify the addresses to bind")
 	return &portForwarder{
 		cmd:              cmd,
 		RESTClientGetter: cliClient.clientFactory,
--- a/pkg/kube/portforwarder_test.go
+++ b/pkg/kube/portforwarder_test.go
@@ -1,0 +1,86 @@
+/*
+ * Copyright The Kmesh Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/kube/client.go [hunk 1]
```diff
@@ -95,8 +95,8 @@ func newPortForwarder(cliClient *client, podName string, ns string, localAddress
 
 	cmd := &cobra.Command{}
 	cmd.Flags().Duration("pod-running-timeout", DefaultPodRunningTimeout, "Timeout for waiting for pod to be running")
-	cmd.Flags().String("namespace", KmeshNamespace, "Specify the namespace to use")
-	cmd.Flags().StringSlice("address", []string{DefaultLocalAddress}, "Specify the addresses to bind")
+	cmd.Flags().String("namespace", ns, "Specify the namespace to use")
```
LiZhenCheng9527, line 98: Changing this flag updates the command metadata, but does not change the namespace used by port-forward. PortForwardOptions.Complete() obtains its namespace from f.ToRawKubeConfigLoader().Namespace(), rather than this Cobra flag. The factory passed to it still uses cliClient.clientFactory, whose namespace is initialized to kmesh-system.

Consequently, passing ns = "my-ns" still looks up the pod in kmesh-system, potentially failing with NotFound or selecting a same-named pod in the wrong namespace.

Could we propagate ns through a per-forwarder RESTClientGetter/config loader before calling Complete(), without mutating the shared client's configuration? This would make the runtime behavior match the supplied namespace, not just the flag value.

PR #1938, opened 2026-08-20
fix(cache): assign real cluster ids in kernel-native mode

### LINKED ISSUE #1936: Kernel-native mode: every cluster is flushed with Id 0, collapsing all circuit-breaker state onto one counter
### What I noticed

In kernel-native mode every cluster is flushed to the BPF map with `Id = 0`.

`pkg/cache/v2/cluster.go:193`:

```go
if cluster.GetApiStatus() == core_v2.ApiStatus_UPDATE {
	cluster.Id = cache.hashName.StrToNum(name)
```

`StrToNum` is a plain map read (`pkg/utils/hash_name.go:127-129`):

```go
func (h *HashName) StrToNum(str string) uint32 {
	return h.strToNum[str]
}
```

The allocator is `Hash` (`hash_name.go:93`) — it is the only function that inserts into `strToNum`. Grepping the tree, `hashName.Hash(` is called exclusively from `pkg/controller/workload/workload_processor.go`, i.e. the workload path. Nothing on the ads path ever calls it.

`NewAdsCache` builds a fresh `utils.NewHashName()` (`pkg/controller/ads/cache.go:55`) and hands it only to `NewClusterCache`. `NewHashName` restores `strToNum` from the persist file, but that file is written only by `flushDelta` (from `Hash`) and `flush` (from `Delete`), so in kernel-native mode it stays empty too. `strToNum` is therefore always empty on this path, and the lookup returns the zero value for every cluster.

The delete branch calling `cache.hashName.Delete(name)` a few lines below (`cluster.go:213`) suggests ids were meant to be registered here.

### Reproduction

Flushing two clusters through `ClusterCache.Flush`:

```
--- FAIL: TestClusterFlushAssignsIds
    cluster_test.go:417: Should not be zero, but was 0
    cluster_test.go:418: Should not be zero, but was 0
    cluster_test.go:419: Should not be: 0x0
```

Both clusters come back with `Id = 0`, and the ids are not distinct.

### Impact

Circuit breaker stats are keyed by `{netns_cookie, cluster_id}` (`bpf/kmesh/ads/include/circuit_breaker.h:78-84`), so every cluster in a netns collapses onto one counter:

- A pod calling `reviews` (`maxConnections: 10`) and `ratings` (`maxConnections: 100`) shares a single `cluster_stats` entry. After 10 *combined* connections, `on_cluster_sock_bind` returns `-1` and further connections to `reviews` are rejected, while `ratings` is throttled at the wrong threshold.
- `clearClusterStats` (`cluster.go:163`) resolves `clusterId = 0` and `BatchDelete`s every key with `ClusterId == 0`, which is all of them. Removing one cluster wipes the live connection counters of every cluster in every netns. Subsequent closes then hit the `-delta > stats->active_connections` guard (`circuit_breaker.h:60-63`) and the counters stay skewed.

### Proposed fix

One line — use the allocator:

```go
cluster.Id = cache.hashName.Hash(name)
```

`clearClusterStats` should keep using `StrToNum`: it resolves an id that `Flush` has already registered, and it runs before `hashName.Delete(name)` on the delete path, so the entry is still present. Switching it to `Hash` would allocate an id for a cluster that is being removed.

I have this locally with a test asserting flushed clusters get non-zero, distinct ids. `TestClusterFlush`, `TestClusterFlushAssignsIds` and the rest of `pkg/cache/v2` that does not require a live BPF map pass. Happy to open a PR.

### Notes on my environment

Five tests in `pkg/cache/v2` (`TestClusterLookupAll`, `TestClearClusterStats`, `TestListenerLookupAll`, `TestListenerFlushAndLookup`, `TestListenerUpdateAndDeleteFlush`) fail identically on unmodified `main` here because I cannot attach real BPF maps, so I could not exercise those paths.

Separately, `bpf/deserialization_to_bpf_map` does not build with clang 21: `deserialization_to_bpf_map.c:1429` trips `-Werror,-Wsometimes-uninitialized` on `prev_key` in `map_restore`. I worked around it locally; happy to file that separately if it is not already known.

### Environment

- Commit: `ae15e8a7`
- Go: 1.26.6

opx0:
**What type of PR is this?**

/kind bug

In kernel-native mode every cluster is flushed to the BPF map with `Id = 0`.

`ClusterCache.Flush` took the id from `HashName.StrToNum`, which is a plain map read rather than the allocator:

```go
func (h *HashName) StrToNum(str string) uint32 {
	return h.strToNum[str]
}
```

`HashName.Hash` is the only function that inserts into `strToNum`, and it is called exclusively from `pkg/controller/workload/workload_processor.go` — the workload path. Nothing on the ads path ever calls it. `NewAdsCache` builds a fresh `utils.NewHashName()` and hands it only to `NewClusterCache`, and the persist file that `NewHashName` restores from is written only by `Hash` and `Delete`, so in kernel-native mode it stays empty too. The lookup therefore returned the zero value for every cluster.

That matters because circuit breaker stats are keyed by `{netns_cookie, cluster_id}` (`bpf/kmesh/ads/include/circuit_breaker.h`), and `on_cluster_sock_bind` reads the id straight off the cluster:

```c
struct cluster_stats_key {
    __u64 netns_cookie;
    __u32 cluster_id;
};
```

So with every id at 0:

- All clusters in a network namespace collapse onto a single `cluster_stats` entry. A pod calling `reviews` (`maxConnections: 10`) and `ratings` (`maxConnections: 100`) shares one counter: after 10 *combined* connections `on_cluster_sock_bind` returns `-1` and further connections to `reviews` are rejected, while `ratings` is throttled at the wrong threshold.
- `clearClusterStats` resolves `clusterId = 0` and `BatchDelete`s every key with `ClusterId == 0`, which is all of them. Removing one cluster wipes the live connection counters of every cluster in every netns, and subsequent closes then hit the `-delta > stats->active_connections` guard, so the counters stay skewed.

The fix is to use the allocator:

```go
cluster.Id = cache.hashName.Hash(name)
```

`Hash` returns the existing id when the name is already registered, so a cluster keeps its id across config pushes and its stats entries stay reachable.

The second commit is a test-only fix. `TestClearClusterStats` built its stats keys from `StrToNum` on a `HashName` nothing had registered into, so all five keys carried `ClusterId 0` and collapsed to three distinct entries; the first `clearClusterStats` call then deleted all of them and the per-cluster assertions never ran. It now registers the names through `Hash` and asserts the whole remaining map, which is the property the test is named for: clearing one cluster leaves the other clusters' counters untouched.

Fixes #1936

- **`clearClusterStats` deliberately stays on `StrToNum`.** It resolves an id that `Flush` has already registered, and it runs before `hashName.Delete(name)` on the delete path, so the entry is still present. Using the allocator there would mint an id for a cluster that is being removed.
- **New behaviour worth knowing:** because `Hash` persists new entries, the ads path now writes `/mnt/hash_name.yaml`, which it never did before. `/mnt` is already a hostPath mount in the daemonset (`deploy/yaml/kmesh.yaml`, `deploy/charts/kmesh-helm/templates/daemonset.yaml`), and this is what makes cluster ids survive a restart, so the `cluster_stats` entries written before the restart stay addressable afterwards.
- **Relation to #1904.** This makes the ads path depend on `HashName.Hash`, which has a separate edge case: `for num = h.hash.Sum32(); num < math.MaxUint32; num++` never enters the body when the hash is exactly `math.MaxUint32`, so it returns an id it never registered. On this path that would mean a live cluster whose name is missing from `strToNum`, and a later `clearClusterStats` resolving 0 for it. It is a 1-in-2^32 case and #1905 already fixes it, so I have not touched it here; flagging the interaction since it becomes reachable from the ads path with this change.
- **Verification.** `TestClusterFlushAssignsIds` fails on the unmodified line (`Should not be zero, but was 0` for both clusters, and the two ids equal) and passes with the fix, under the same `-race -gcflags=all=-l` flags CI uses. `pkg/controller/ads/...` and the rest of `pkg/cache/v2` show no new failures against `main`; five tests in `pkg/cache/v2` (`TestClusterLookupAll`, `TestClearClusterStats`, `TestListenerLookupAll`, `TestListenerFlushAndLookup`, `TestListenerUpdateAndDeleteFlush`) fail identically on unmodified `main` in my environment because I cannot attach real BPF maps there. That means the `TestClearClusterStats` change in the second commit is exercised by CI rather than by me locally.
- **Coverage I did not add.** @AnouarMohamed suggested asserting that two clusters with different circuit-breaker limits do not share `cluster_stats`. That property lives in the BPF map, not in the Go cache, so it needs an e2e test rather than a unit test; what the unit test can show is what it now shows, that the two clusters resolve to distinct non-zero ids, with the differing `maxConnections` in the fixture to keep the scenario recognisable. Happy to follow up with the e2e case if you would like it.
- **AI disclosure**, per CONTRIBUTING.md: I used Claude Code while preparing this PR — tracing the call paths, drafting the tests, and drafting this description. I have reviewed and run every change myself.

```release-note
Fix kernel-native mode assigning cluster id 0 to every cluster, which made all clusters in a network namespace share a single circuit breaker counter and let the removal of one cluster wipe every cluster's connection stats.
```

### DIFF
--- a/pkg/cache/v2/cluster.go
+++ b/pkg/cache/v2/cluster.go
@@ -190,7 +190,7 @@ func (cache *ClusterCache) Flush() {
 	defer cache.mutex.Unlock()
 	for name, cluster := range cache.apiClusterCache {
 		if cluster.GetApiStatus() == core_v2.ApiStatus_UPDATE {
-			cluster.Id = cache.hashName.StrToNum(name)
+			cluster.Id = cache.hashName.Hash(name)
 			err := maps_v2.ClusterUpdate(name, cluster)
 			if cluster.GetLbPolicy() == cluster_v2.Cluster_MAGLEV {
 				// create consistent lb here and update table to bpf map
--- a/pkg/cache/v2/cluster_test.go
+++ b/pkg/cache/v2/cluster_test.go
@@ -391,3 +391,58 @@ func TestClearClusterStats(t *testing.T) {
 		assert.Nil(t, iter.Err())
 	}
 }
+
+func TestClusterFlushAssignsIds(t *testing.T) {
+	patches := gomonkey.NewPatches()
+	patches.ApplyFunc(maps_v2.ClusterUpdate, func(key string, value *cluster_v2.Cluster) error {
+		return nil
+	})
+	defer patches.Reset()
+
+	// Circuit breaker stats are held in a map keyed by {netns_cookie,
+	// cluster_id}, so two clusters only keep independent connection counters
+	// and limits while their ids differ.
+	newCluster := func(name string, maxConnections uint32) *cluster_v2.Cluster {
+		return &cluster_v2.Cluster{
+			ApiStatus:       core_v2.ApiStatus_UPDATE,
+			Name:            name,
+			LbPolicy:        cluster_v2.Cluster_RANDOM,
+			CircuitBreakers: &cluster_v2.CircuitBreakers{MaxConnections: maxConnections},
+		}
+	}
+
+	t.Run("clusters get distinct non-zero ids", func(t *testing.T) {
+		hashName := utils.NewHashName()
+		defer hashName.Reset()
+		cache := NewClusterCache(nil, hashName)
+		reviews := newCluster("outbound|9080||reviews.default.svc.cluster.local", 10)
+		ratings := newCluster("outbound|9080||ratings.default.svc.cluster.local", 100)
+		cache.SetApiCluster(reviews.Name, reviews)
+		cache.SetApiCluster(ratings.Name, ratings)
+
+		cache.Flush()
+
+		assert.NotZero(t, reviews.Id)
+		assert.NotZero(t, ratings.Id)
+		assert.NotEqual(t, reviews.Id, ratings.Id)
+	})
+
+	t.Run("id is stable across flushes", func(t *testing.T) {
+		hashName := utils.NewHashName()
+		defer hashName.Reset()
+		cache := NewClusterCache(nil, hashName)
+		cluster := newCluster("outbound|9080||reviews.default.svc.cluster.local", 10)
+		cache.SetApiCluster(cluster.Name, cluster)
+
+		cache.Flush()
+		firstId := cluster.Id
+		assert.NotZero(t, firstId)
+
+		// A later config push re-flushes the same cluster; its stats entries
+		// are keyed by the id, so reallocating one would strand them.
+		cluster.ApiStatus = core_v2.ApiStatus_UPDATE
+		cache.Flush()
+
+		assert.Equal(t, firstId, cluster.Id)
+	})
+}
--- a/pkg/cache/v2/cluster_test.go
+++ b/pkg/cache/v2/cluster_test.go
@@ -358,14 +358,15 @@ func TestClearClusterStats(t *testing.T) {
 	adsObj := loader.GetBpfKmesh()
 	assert.NotNil(t, adsObj)
 	hashName := utils.NewHashName()
+	defer hashName.Reset()
 	clusterCache := NewClusterCache(adsObj, hashName)
 	testClusters := []string{"test_cluster1", "test_cluster2", "test_cluster3"}
 	testKeys := []ClusterStatsKey{
-		{NetnsCookie: 0, ClusterId: hashName.StrToNum(testClusters[0])},
-		{NetnsCookie: 1, ClusterId: hashName.StrToNum(testClusters[0])},
-		{NetnsCookie: 2, ClusterId: hashName.StrToNum(testClusters[0])},
-		{NetnsCookie: 0, ClusterId: hashName.StrToNum(testClusters[1])},
-		{NetnsCookie: 0, ClusterId: hashName.StrToNum(testClusters[2])},
+		{NetnsCookie: 0, ClusterId: hashName.Hash(testClusters[0])},
+		{NetnsCookie: 1, ClusterId: hashName.Hash(testClusters[0])},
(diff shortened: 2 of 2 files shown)


Aryan-B2107:
@opx0 

I’ve taken the e2e for point 2 (two clusters with different circuit-breaker limits must not share cluster_stats).

The test is TestCircuitBreakerClusterStatsIsolation: one client pod, DestinationRules with maxConnections: 10 and 100 (same reviews/ratings scenario), hold more than 10 connections on the high-limit cluster, then assert the low-limit cluster still accepts. That is the BPF-map property the unit test cannot cover.

I will keep this off #1938 so we don’t both land the Hash/Id fix. After this merges I’ll open a follow-up with that test and ping you for review.

If you’d rather have the e2e on this PR, say so and I can send the commits here instead.

opx0:
@Aryan-B2107 
This is great thanks for taking it on. The setup is exactly right: holding more than 10 connections on the high-limit cluster and then confirming the low-limit one still accepts is precisely the isolation that lives in the BPF map, which is the part a unit test just can't reach. Good to have it nailed down.

Let's do it as a follow-up, the way you suggested. It keeps things clean on both sides — #1938 stays a small, one-line fix with its unit tests, so it's easy to review and merge, and your e2e gets to run against main once the fix is actually there, which is what it needs anyway. Bundling them would just gate the little fix behind the full Istio CI for no real gain.

So I'll focus on getting #1938 in, then ping me the moment your follow-up is up and I'll review it right away. Appreciate you jumping on this one.

Aryan-B2107:
Sounds good, follow-up after this merges, e2e only.

Could you ping me when #1938 is in? I’ll open the PR against main right after and tag you for review.

PR #1946, opened 2026-09-12
Fix: use proper logger for telemetry access logs instead of fmt.Println

### LINKED ISSUE #1947: Replace fmt.Println with proper logger in telemetry access logs
Describe the bug
In the pkg/controller/telemetry package, the outputAccesslog function uses fmt.Println to print access logs to the console instead of using the standard logger provided by the package.

Using fmt.Println bypasses the configured logging levels (e.g., info, debug, error), formatting, and output destinations of the Kmesh logger, which can make debugging and log aggregation difficult in production environments.

Expected Behavior
The access log should be printed using the existing log instance (logger.NewLoggerScope("telemetry")) initialized for the telemetry package.

Location of the issue
pkg/controller/telemetry/accesslog.go
Line: ~86

Proposed Solution
Replace fmt.Println("accesslog:", logStr) with log.Infof("accesslog: %s", logStr) to ensure access logs adhere to the Kmesh logging standards.

Rucha0901:
Describe the bug
In the pkg/controller/telemetry package, the outputAccesslog function uses fmt.Println to print access logs to the console instead of using the standard logger provided by the package.

Using fmt.Println bypasses the configured logging levels (e.g., info, debug, error), formatting, and output destinations of the Kmesh logger, which can make debugging and log aggregation difficult in production environments.

Expected Behavior
The access log should be printed using the existing log instance (logger.NewLoggerScope("telemetry")) initialized for the telemetry package.

Proposed Solution
Replace fmt.Println("accesslog:", logStr) with log.Infof("accesslog: %s", logStr) to ensure access logs adhere to the Kmesh logging standards.

Resolves #1947


sivakumar-mahalingam:
Thanks for the fix. Few things, I observed:

1. Access logs already have their own toggle (`kmeshctl monitoring --accesslog`). Routing them through `log.Infof` means they also depend on the log level, so setting the level to warn/error would silently drop them.
2. Output format changes from plain `accesslog: ` line to logger's format (timestamp, level, etc.). This will break existing log parsers and docs. Can you share a before/after sample, if possible?
3. Patch coverage is 0%. Please add a small test for `outputAccesslog`.

use a dedicated `accesslog` logger scope with its level fixed at info, so it stays independent of the daemon log level.

PR #1948, opened 2026-09-12
fix(telemetry): update map entry with appended operation durations

Norway-02:
In `updateOperationMetricCache`, when an existing label key is found in `operationMetricCache`, `v.durations = append(...)` updates a local copy of the `operationDuration` struct. However, the modified entry was never written back to `m.operationMetricCache[labels]`. This caused subsequent operation duration measurements for existing labels to be discarded before flushing metrics to Prometheus.

This PR re-assigns the updated struct `v` back to `m.operationMetricCache[labels]` and adds unit test coverage for multiple cache updates under identical labels.

Fixes # NONE

- Added unit test `TestUpdateOperationMetricCache_MultipleUpdates` in `bpf_prog_metric_test.go` to verify sequential updates.

**Release note**:
```release-note
fix(telemetry): update map entry with appended operation durations in bpf_prog_metric
```


Norway-02:
CC @LiZhenCheng9527 @YaoZengzeng Please review when you have time. Thank you!

PR #1949, opened 2026-09-14
free value when element list node allocation fails

### LINKED ISSUE #826: Memory Leak in /bpf/deserialization_to_bpf_map/deserialization_to_bpf_map.c
**What happened**:
Some places called calloc or malloc but not free it.

**Environment**:
- Kmesh version:
- Others:

SATVIKsynopsis:
**What type of PR is this?**

/kind bug

fixes a memory leak in `create_struct_list()`. when `create_struct()` successfully creates value but the following calloc for `element_list_node fails` value is lost without being freed.  so, this adds `free(value)` before breaking out of the loop.

Fixes #826 

verified with an `asan` test that forces the `element_list_node` allocation to fail. before the fix the asan reported a 64 byte leak from `create_struct()`. after the fix, no leak is reported.

before asan:
Screenshot 2026-09-14 at 12 00 46 PM

after asan:
Screenshot 2026-09-14 at 12 01 36 PM

NONE
If no, just write "NONE" in the release-note block below.
If yes, a release note is required.
-->
```release-note

```


SATVIKsynopsis:
changed `free(value)` to `deserial_free_elem(value)` so the allocated things are properly cleaned up. also verified the fix with asan as no leaks are reported after the change.

PR #1950, opened 2026-09-15
fix(kolog): close /proc/stat file descriptor in getBootTime

Norway-02:
In `getBootTime()`, `os.Open("/proc/stat")` opens `/proc/stat` to read boot time (`btime`). However, `defer data.Close()` was missing, causing the open file descriptor to leak every time `getBootTime()` is invoked.

This PR adds `defer func() { _ = data.Close() }()` to ensure the `/proc/stat` file handle is closed immediately upon function exit.

Fixes # NONE

- 1 file changed (`pkg/kolog/kolog.go`).
- All package linters and tests passed cleanly.

**Release note**:
```release-note
fix(kolog): close /proc/stat file descriptor in getBootTime
```


Norway-02:
CC @LiZhenCheng9527 @YaoZengzeng Please review when you have time. Thank you!

PR #1953, opened 2026-09-20
fix(ctl): preserve waypoint labels with revision

### LINKED ISSUE #1952: kmeshctl waypoint drops --for label when --revision is specified
**What happened**:

When generating a waypoint with both the `--for` and `--revision` flags, `kmeshctl` silently removes the label created by the `--for` flag.

For example, I ran:

```bash
kmeshctl waypoint generate \
  --for workload \
  --revision canary \
  --namespace default
```

The generated Gateway contains only the revision label:

```yaml
metadata:
  labels:
    istio.io/rev: canary
```

The expected `istio.io/waypoint-for: workload` label is missing.

The generated Gateway should contain both labels because both flags were supplied:

```yaml
metadata:
  labels:
    istio.io/waypoint-for: workload
    istio.io/rev: canary
```

1. Build or install `kmeshctl`.
2. Run:

```bash
kmeshctl waypoint generate \
  --for workload \
  --revision canary \
  --namespace default
```

3. Check `metadata.labels` in the generated YAML.

Actual result:

```yaml
labels:
  istio.io/rev: canary
```

Expected result:

```yaml
labels:
  istio.io/waypoint-for: workload
  istio.io/rev: canary
```

The issue can also be demonstrated by running the command without `--revision`:

```bash
kmeshctl waypoint generate \
  --for workload \
  --namespace default
```

In that case, the traffic-type label is generated correctly:

```yaml
labels:
  istio.io/waypoint-for: workload
```

The problem appears to be in the `makeGateway` closure in `ctl/waypoint/waypoint.go`.

The traffic-type label is added first:

```go
gw.Labels[KmeshWaypointForTrafficTypeLabel] = trafficType
```

When a revision is specified, the code then replaces the entire labels map:

```go
gw.Labels = map[string]string{label.IoIstioRev.Name: revision}
```

As a result, the previously added `istio.io/waypoint-for` label is lost. The revision label should be added to the existing labels map instead of replacing it.

This issue is reproducible locally with `waypoint generate`; no Kubernetes cluster is required.

**Environment**:

- Kmesh version: Main branch at commit `60881fc0bcb1` (`v1.2.0-130-g60881fc0`)
- Kmesh mode(kmesh has `Kernel-Native Mode` and `Dual-Engine Mode`): Not applicable; reproduced using local YAML generation
- Istio version: Not applicable
- Kernel version: `7.0.0-31-generic`
- Others: `kmeshctl` built with Go `1.24.2`

sivakumar-mahalingam:
**What type of PR is this?**

/kind bug

Preserves existing Gateway labels when adding `istio.io/rev`, so using `--for` and `--revision` together retains both labels. Adds a focused regression test covering both flags, each flag separately, and neither flag.

Fixes #1952

```release-note
Fix `kmeshctl waypoint` generation to preserve the `istio.io/waypoint-for` label when `--revision` is also specified.
```

**Tests**:

- `go test -count=1 ./ctl/waypoint`
- `go vet ./ctl/waypoint`


sivakumar-mahalingam:
The two Istio 1.28 E2E jobs were cancelled at the workflow’s 40-minute timeout. DCO, builds, and the other E2E jobs passed. I don’t have permission to rerun the cancelled jobs. @nlgwcy @LiZhenCheng9527 Could you please rerun them?

PR #1955, opened 2026-09-20
fix(cache): remove stale service address indexes

### LINKED ISSUE #1954: Service updates leave stale address indexes in ServiceCache
**What happened**:

`serviceCache.AddOrUpdateService` replaces the service stored in `servicesByResourceName` and adds indexes for the service's current addresses.

However, when an existing service changes its addresses, indexes belonging to the previous version are not removed from `servicesByAddr` or `servicesByIP`.

For example:

1. Add service `default/example` with address `10.240.10.1`.
2. Update the same service to use address `10.240.10.2`.
3. Look up `10.240.10.1`.

The old address still resolves to the previous service object.

Deleting the updated service does not remove this stale entry because `DeleteService` only processes addresses present in the latest service object.

After a service update:

- Removed addresses should no longer resolve to the service.
- Current addresses should resolve to the updated service.
- Address entries that have been reassigned to another service must remain
  untouched.

Add a unit test in `pkg/controller/workload/cache/service_cache_test.go`:

1. Add a service with address `10.240.10.1`.
2. Add the same service resource again with address `10.240.10.2`.
3. Verify `GetServiceByAddr` and `GetServiceByIP` return `nil` for
   `10.240.10.1`.
4. Verify both lookups return the updated service for `10.240.10.2`.

The checks for the old address currently fail.

`serviceCache.deleteAddr` already contains ownership checks intended to
avoid deleting an address that has been reassigned to another service.
The update path could reuse this behavior when removing indexes from the
previous service version.

This should be reproducible and testable without a Kubernetes cluster.

**Environment**:

- Kmesh version: current `main`
- Kmesh mode: N/A
- Istio version: N/A
- Kernel version: N/A
- Others: N/A

sivakumar-mahalingam:
**What type of PR is this?**

/kind bug

Cleans up address indexes belonging to the previous version of a service before indexing an updated service. This prevents removed addresses from continuing to resolve through `servicesByAddr` or `servicesByIP` while preserving addresses reassigned to another service.

Adds regression tests covering stale index removal and address reassignment.

Fixes #1954

The update path reuses the existing `deleteAddr` ownership checks. Tests cover both network-address and IP-only lookups.

Yes.

```release-note
Fixed stale service address lookups after a service's addresses are updated.
```


sivakumar-mahalingam:
The Istio 1.27 IPv6 E2E check failed in TestMixNsAndServiceWaypoint because X-Request-Id was missing. The same test passed on the preceding run with identical code. Could a maintainer rerun the failed job?

